PR #25, opened 2015-09-01
Implement exponential backoff for reservation system REST API communication

mksh:
Backoff functionality itself & tests.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,5 @@
 /junit-reports/
 *.swp
 /target/
+.idea
+*.iml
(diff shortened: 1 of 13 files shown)


emcho:
Thanks for the patch! Could you please make sure that you respect the project code convention:

https://jitsi.org/codeconvention

(I was seeing problems with accolades and a bunch of fields/methods with no javadocs)

Also it would help if you describe exactly what the issue is and how this is addressing it.

mksh:
Hello again. I`ve updated the code to match the code conventions, hope that nothing was missed.

As for issue description. The successful delivery of HTTP calls that are done by Jicofo for room reservation is critical for the product experience. Failures of `Creating conference` and `Reading conference info` calls handling will lead to error messages that worsen the user exerience; failures of `Deleting conference` call handling will lead to unacceptable data inconsistency for the straightforward room reservation system. This patch aims to deal with these issues by resending the HTTP callbacks in the case of server-side or networking failures with exponential backoff.

PR #27, opened 2015-10-15
UTF8 Charset bugfix

misi:
On shibboleth redirect html page the user "display name" charset is not set to UTF8, 
and it causes a displaying issue when it contains international chars


misi:
I have just signed the ICLA.

PR #32, opened 2016-02-10
Make Jicofo max memory configurable

CPWeaver:
Just a simple change to optionally configure jicofo max memory (-Xmx) from an environment variable.


CPWeaver:
We have signed a corporate agreement, can you add me to the list with @dwilson6 and @PetroccoCo?

PetroccoCo:
Not sure if it's needed but I can verify @CPWeaver is covered under our corporate agreement.

alexanderankin:
Bumping this, can this be merged? I was gonna submit this same patch, I am trying to configure this like the equivalent `VIDEOBRIDGE_MAX_MEMORY=3072m` setting that exists here https://github.com/jitsi/jitsi-videobridge/blob/master/resources/jvb.sh#L27

it would make sense for the repositories to be consistent in this way.

bgrozev:
See #540

PR #42, opened 2016-02-24
Optionally returns JVB list in REST health response.

paweldomas:
If you run Jicofo locally open http://localhost:8888/about/health?list_jvb=true to see Jitsi videobridges known to your Jicofo instance.

### DIFF
--- a/src/main/java/org/jitsi/jicofo/rest/Health.java
+++ b/src/main/java/org/jitsi/jicofo/rest/Health.java
@@ -26,6 +26,7 @@ import org.eclipse.jetty.server.*;
 
 import org.jitsi.jicofo.*;
 import org.jitsi.util.*;
+import org.json.simple.*;
 
 /**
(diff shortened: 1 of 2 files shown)


src/main/java/org/jitsi/jicofo/rest/Health.java
```diff
@@ -138,6 +145,30 @@ static void getJSON(
...
         {
             check(focusManager);
+
+            // At this point the health check has passed - now eventually list
+            // JVBs known to this Jicofo instance
+            String listJvbParam = request.getParameter(LIST_JVB_PARAM_NAME);
+
+            if (!StringUtils.isNullOrEmpty(listJvbParam)
+                && "true".equals(listJvbParam.toLowerCase()))
+            {
```
damencho: Just a typo, maybe using Boolean.parseBoolean(listJvbParam) makes it more readable.

PR #43, opened 2016-02-26
Returns 500 to health check if JVB list was requested, but empty.

paweldomas:
This way we know that Jicofo is ready to start conferences when returns 200.


damencho:
test this please

PR #52, opened 2016-03-15
Jibri

paweldomas:
The whole thing is activated by specifying "org.jitsi.jicofo.jibri.BREWERY" config property which is the address of MUC room where Jibri instances publish their idle/busy status. JibriDetector class will join that room track Jibri status changes which will be published further through JibriListener interface. Currently only JibriRecorder implements that and it's instance is created for every JitsiMeetConference.

JibriRecorder will publish the info about current recording status in Jicofo's MUC presence:
- "undefined" when there are no Jibris available
- "off" when recording is stopped and there are some Jibri available to start recording
- "pending" recording is starting and we're waiting for confirmation from Jibri(this takes some time)
- "on" recording is in progress

Recording is controlled by sending Jibri IQ from conference moderator user to Jicofo which then does select Jibri instance and starts the recording.

### DIFF
--- a/src/main/java/org/jitsi/jicofo/recording/jibri/JibriRecorder.java
+++ b/src/main/java/org/jitsi/jicofo/recording/jibri/JibriRecorder.java
@@ -1,0 +1,484 @@
+/*
+ * Jicofo, the Jitsi Conference Focus.
+ *
+ * Copyright @ 2015 Atlassian Pty Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jitsi.jicofo.recording.jibri;
+
+import net.java.sip.communicator.impl.protocol.jabber.extensions.colibri.*;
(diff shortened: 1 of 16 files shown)


src/main/java/org/jitsi/jicofo/recording/jibri/JibriRecorder.java
```diff
@@ -0,0 +1,484 @@
...
+            startIq.setAction(JibriIq.Action.START);
+
+            startIq.setStreamId(iq.getStreamId());
+            startIq.setUrl(iq.getUrl());
+
+            logger.debug("Starting Jibri recording: " + startIq.toXML());
+
+            IQ startReply
+                = (IQ) xmpp.getXmppConnection()
+                        .sendPacketAndGetReply(startIq);
```
bgrozev: If I am reading this correctly, the Jibri instance for jibriJid will remain with status "IDLE" until we receive a response from the network. And if we receive another request for recording during this time, the same Jibri instance will be selected again since it's on top of the list.

src/main/java/org/jitsi/jicofo/recording/jibri/JibriRecorder.java
```diff
@@ -0,0 +1,484 @@
...
+            }
+        }
+        // stop ?
+        else if (JibriIq.Action.STOP.equals(action) &&
+            recorderComponentJid != null &&
+            (JibriIq.Status.ON.equals(jibriStatus) ||
+             JibriIq.Status.PENDING.equals(jibriStatus)))
+        {
+            if (!verifyModeratorRole(iq))
+                return;
```
bgrozev: Maybe we should send a response IQ here? And we could move this to the beginning of the method and remove the duplicate check.

src/main/java/org/jitsi/jicofo/recording/jibri/JibriRecorder.java
paweldomas: If Jibri would receive start request twice then it will return an error, but you're right that this should be handled in Jicofo as well.

PR #59, opened 2016-03-29
Jibri fixes2

paweldomas:
Just another round of Jibri improvements. Commit messages contain more details.

### DIFF
--- a/src/main/java/org/jitsi/jicofo/JitsiMeetGlobalConfig.java
+++ b/src/main/java/org/jitsi/jicofo/JitsiMeetGlobalConfig.java
@@ -30,6 +30,12 @@ import org.osgi.framework.*;
 public class JitsiMeetGlobalConfig
 {
     /**
+     * The logger instance used by this class.
+     */
+    private final static Logger logger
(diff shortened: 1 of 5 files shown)


src/main/java/org/jitsi/jicofo/JitsiMeetGlobalConfig.java
```diff
@@ -103,7 +128,22 @@ private void init(ConfigurationService configService)
...
+                    MAX_SSRC_PER_USER_CONFIG_PNAME, DEFAULT_MAX_SSRC_PER_USER);
+
+        jibriPendingTimeout
+            = configService.getInt(
+                    PENDING_TIMEOUT_PROP_NAME, DEFAULT_PENDING_TIMEOUT);
+
+        if (jibriPendingTimeout > 0)
+        {
+            logger.info(
+                    "Jibri requests in PENDING state will be timed out after: "
```
yanas: It may be a good idea to change the log text to something like "The jibri pending timeout has been set to: ", because the above message can be confusing and we can think that the countdown has already started when we see it.

PR #62, opened 2016-04-01
Restart the conference without reloading the page

paweldomas:
This PR makes Jicofo move the conference to another bridge when the current one dies. This is done by sending 'transport-replace' message to the client which contains transport info for the new bridge.

### DIFF
--- a/src/main/java/org/jitsi/impl/protocol/xmpp/colibri/ColibriConferenceImpl.java
+++ b/src/main/java/org/jitsi/impl/protocol/xmpp/colibri/ColibriConferenceImpl.java
@@ -574,10 +574,7 @@ public class ColibriConferenceImpl
     @Override
     public void dispose()
     {
-        synchronized (syncRoot)
-        {
-            this.disposed = true;
-        }
+        this.disposed = true;
     }
 
     /**
--- a/src/main/java/org/jitsi/impl/protocol/xmpp/colibri/ColibriConferenceImpl.java
+++ b/src/main/java/org/jitsi/impl/protocol/xmpp/colibri/ColibriConferenceImpl.java
@@ -44,13 +44,14 @@ import java.util.*;
 public class ColibriConferenceImpl
     implements ColibriConference
 {
-    private final static net.java.sip.communicator.util.Logger logger
-            = Logger.getLogger(ColibriConferenceImpl.class);
+    private final static Logger logger
+        = Logger.getLogger(ColibriConferenceImpl.class);
 
     /**
      * The instance of XMPP connection.
      */
     private final XmppConnection connection;
+
     /**
      * XMPP address of videobridge component.
      */
@@ -138,7 +139,7 @@ public class ColibriConferenceImpl
         if (!StringUtils.isNullOrEmpty(conferenceState.getID()))
         {
             throw new IllegalStateException(
-                "Can not change the bridge on active conference");
+                "Cannot change the bridge on active conference");
         }
         this.jitsiVideobridge = videobridgeJid;
     }
--- a/src/main/java/org/jitsi/impl/protocol/xmpp/colibri/ColibriConferenceImpl.java
+++ b/src/main/java/org/jitsi/impl/protocol/xmpp/colibri/ColibriConferenceImpl.java
@@ -92,7 +92,13 @@ public class ColibriConferenceImpl
      * last {@link #createColibriChannels(boolean, String, boolean, List)} call.
      */
     private boolean justAllocated = false;
-    
+
+    /**
+     * Flag indicates that this instance has been disposed and should not be
+     * used anymore.
+     */
+    private boolean disposed;
+
     /**
      * Creates new instance of <tt>ColibriConferenceImpl</tt>.
      * @param connection XMPP connection object that wil be used by new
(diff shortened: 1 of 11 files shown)


paweldomas:
jenkins: after merging https://github.com/jitsi/lib-jitsi-meet/pull/41it is ok to test again

src/main/java/org/jitsi/impl/protocol/xmpp/colibri/ColibriConferenceImpl.java
```diff
@@ -512,16 +562,43 @@ public void expireConference()
...
         }
     }
 
     /**
      * {@inheritDoc}
      */
     @Override
+    public void dispose()
+    {
+        synchronized (syncRoot)
```
lyubomir: The field disposed changes its value from false to true only (i.e. there's no change from true to false). Additionally, the method isDisposed() is not synchronized on syncRoot. Do you think it makes sense to remove the synchronized on syncRoot here?

src/main/java/org/jitsi/impl/protocol/xmpp/colibri/ColibriConferenceImpl.java
```diff
@@ -104,6 +110,26 @@ public ColibriConferenceImpl(XmppConnection connection)
...
+     * Checks if this instance has been disposed already and if so prints
+     * a warning message.
+     *
+     * @param operationName the name of the operation that will not happen and
+     * should be mentioned in the warning message.
+     *
+     * @return <tt>true</tt> if this instance has not been disposed yet or
+     * <tt>false</tt> otherwise.
+     */
+    private boolean assertNotDisposed(String operationName)
```
lyubomir: I don't like the method name assertNotDisposed because:

1. It sets the wrong expectation for me. I expect assert methods to throw an exception (because that's how I've seen them behave elsewhere like in JUnit). Apple's Dictionary.app says that assert means "state a fact or belief confidently and forcefully". I think "confidently and forcefully" aligns with "throw an exception".

2. Its uses read unnecessarily non-trivially. You're only ever using the method in the form "if (!assertNotDisposed(...)) return ..." which a double negative. I believe the source code will be easier to read without a double negative.

src/main/java/org/jitsi/protocol/xmpp/AbstractOperationSetJingle.java
```diff
@@ -103,12 +104,38 @@ public boolean initiateSession(boolean useBundle,
 
         sessions.put(sid, session);
 
+        JingleIQ inviteIQ = createInviteIQ(
+                sid, useBundle, address, contents, startMuted);
```
lyubomir: I'm not sure why the above formatting is preferred over:

JingleIQ inviteIQ
    = createInviteIQ(sid, useBundle, address, contents, startMuted);

I find my example to be easier to read.

src/main/java/org/jitsi/protocol/xmpp/AbstractOperationSetJingle.java
```diff
@@ -134,8 +163,24 @@ public boolean initiateSession(boolean useBundle,
...
+     * Checks based on supplied <tt>reply</tt> IQ value whether or not given
+     * <tt>JingleSession</tt> has been accepted by the client.
+     *
+     * @param session <tt>JingleSession</tt> instance for which we're evaluating
+     * the response value.
+     * @param reply <tt>IQ</tt> response to Jingle invite IQ or <tt>null</tt> in
+     * case of timeout.
+     *
+     * @return <tt>true</tt> if invite IQ for which we're evaluating the value
+     * of response packet is considered accepted or <tt>false</tt> otherwise.
```
lyubomir: I don't get it... Is "<tt>reply</tt> IQ" the same as "response packet"?

paweldomas: yes, response packet which is type of IQ(as in smack types), by <tt>IQ</tt> I meant the type of argument, does it sense ?

src/main/java/org/jitsi/jicofo/ChannelAllocator.java
```diff
@@ -0,0 +1,688 @@
...
+     * A participant that is to be invited by this instance to the conference.
+     */
+    private final Participant newParticipant;
+
+    /**
+     * First argument stands for "start audio muted" and the second one for
+     * "start video muted". The information is included as a custom extension in
+     * 'session-initiate' sent to the user.
+     */
+    private final boolean[] startMuted;
```
lyubomir: I'm starting to feel uncomfortable with the startMuted array in its multiple incarnations. I've seen it in at least JitsiMeetConference and ChannelAllocator. And there were multiple methods calls with it as an arguments.

1. It's cryptic because you have to know what's what at which index. (I've used the approach as well so I'm not stating I'm a saint.)

2. JitsiMeetConference.hasToStartMuted checks for null but  AbstractOperationSetJingle.createInviteIQ doesn't. In practice, you're unlikely to get to a null with our current usage. But it's still not pretty.

I'd suggest an enum, maybe flag-style or something.

paweldomas: Yes, I agree and have added a FIXME for that, but it's not the purpose of this PR.

https://github.com/jitsi/jicofo/blob/ice_restart/src/main/java/org/jitsi/protocol/xmpp/AbstractOperationSetJingle.java#L156

lyubomir: Cool, thank you! I edited my comment before I saw your response that that's not to be fixed in this PR.

PR #66, opened 2016-04-06
Adds RTX by default to offers

bgrozev:
Cleans up JingleOfferFactory. Removes RED, ULPFEC and some unused audio formats from the offer, and adds RTX. RTX is optionally disabled on a per-conference basis by setting disableRtx=true in the conference creation request (or adding #config.disableRtx=true to the URL in Jitsi-Meet).

Note: this needs jitsi-videobridge#209 (not directly, but stuff will be broken without it).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -123,7 +123,7 @@
     <dependency>
       <groupId>${project.groupId}</groupId>
       <artifactId>libjitsi</artifactId>
-      <version>1.0-20160308.153702-107</version>
+      <version>1.0-20160329.203638-125</version>
     </dependency>
     <dependency>
       <groupId>org.osgi</groupId>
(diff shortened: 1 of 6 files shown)


paweldomas:
@bgrozev not sure if this PR's intention was to add RTX to offer by default, from the code it looks like it is added only when peer supports it, am I missing something ? I'm asking because PR title says something different and if it is meant to be always included by default then it should also end up here https://github.com/jitsi/jicofo/blob/master/src/main/java/org/jitsi/jicofo/discovery/DiscoveryUtil.java#L193

bgrozev:
Good catch, Pawel, I hadn't thought of this.

There's two parts.  The per-conference configuration: if it is false, no one in the conference will be offered rtx. This is the part controlled through disableRtx, and it defaults to enabled. Then there's the per-participant part which will not include rtx in the offer unless it has advertised support. I think that keeping the default here to 'false' is safer, since adding rtx for a participant that doesn't support it will break retransmissions for it.

paweldomas:
Yes, it's better to keep it this way

paweldomas:
Hey @gpolitis why have you merged the PR when the tests have failed :P ? There is complication error on "test" part where method signatures have changed

paweldomas:
https://github.com/jitsi/jicofo/commit/178d0a3edb7924f37ed0d54ea7ce920a503e034a

gpolitis:
I'm sorry, I apologize for this @paweldomas, I merged this too quickly.

PR #69, opened 2016-04-10
Remove ant to remove confusion about build.

bickelj:
Because mvn is now used to do all compilation and packaging, we can remove the build.xml and update the instructions in README.md to have the mvn command instead of ant command.


bickelj:
Hi @jitsi-jenkins I already signed the CLA, thanks ; )

cheenwe:
@bickelj @jitsi-jenkins 　I want to execute　"./jicofo.sh --domain=jitsi.example.com --secret=focus_secret --user_domain=auth.jitsi.example.com --user_name=focus --user_password=focuspassword"，　but i can't find this script  " /jicofo.sh",  thanks ; )

damencho:
@bickelj  Well, currently ant is used for building the debian packages, if we remove as is we will stop any debian packages and new releases.
@cheenwe The source of the file is in resources/install/linux and resources/install/linux-64 ... The easiest way is to make assembly and extract it, where everything will be in place (mvn -U package -DskipTests -Dassembly.skipAssembly=false)

bickelj:
@damencho I did not know ant was being used for the purpose of Debian builds, please do keep doing those : )

Perhaps the readme should still be updated, but the build.xml kept, what do you think?

damencho:
Yeah README needs update, you are right.

PR #70, opened 2016-04-11
Allows FEC/RED to be enabled.

bgrozev:
Adds FEC/RED to created offers, if the conference configuration includes
"enableFec=true".


bgrozev:
Jenkins: ok to test

paweldomas:
I don't like how the argument list grows for the method, maybe we could split it into createVideoContent, audio content and data content methods ?

PR #77, opened 2016-04-27
Lipsync

paweldomas:
Adds a LipSync hack which will selectively merge audio and video streams to enable lipsync in WebRTC stack on the client. Currently this will be enabled for Chrome clients only, given that at the time of SSRC advertisement(joining the conference) remote participant has not his video muted.

### DIFF
--- a/src/main/java/org/jitsi/protocol/xmpp/util/MediaSSRCMap.java
+++ b/src/main/java/org/jitsi/protocol/xmpp/util/MediaSSRCMap.java
@@ -242,7 +242,8 @@ public class MediaSSRCMap
      */
     public boolean remove(String media, SourcePacketExtension ssrc)
     {
-        return getSSRCsForMedia(media).remove(ssrc);
+        SourcePacketExtension toBeRemoved = findSSRC(media, ssrc.getSSRC());
+        return toBeRemoved != null && getSSRCsForMedia(media).remove(ssrc);
     }
 
     /**
--- a/src/main/java/org/jitsi/protocol/xmpp/util/MediaSSRCMap.java
+++ b/src/main/java/org/jitsi/protocol/xmpp/util/MediaSSRCMap.java
@@ -123,6 +123,35 @@ public class MediaSSRCMap
     }
 
     /**
+     * Creates a deep copy of this <tt>MediaSSRCMap</tt>.
+     *
+     * @return a new instance of <tt>MediaSSRCMap</tt> which contains copies of
+     *         <tt>SourcePacketExtension</tt> stored in this map.
+     */
+    public MediaSSRCMap copyDeep()
+    {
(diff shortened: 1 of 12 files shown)


src/main/java/org/jitsi/protocol/xmpp/util/SSRCSignaling.java
```diff
@@ -88,6 +88,31 @@ public static void deleteSSRCParams(SourcePacketExtension ssrcPe)
...
+     * @return the first <tt>SourcePacketExtension</tt> with a valid media
+     *         stream id or <tt>null</tt> if there aren't any such streams
+     *         in the list.
+     */
+    public static SourcePacketExtension getFirstWithMSID(
+            List<SourcePacketExtension> ssrcs)
+    {
+        for (SourcePacketExtension ssrc : ssrcs)
+        {
+            String videoStreamId = getStreamId(ssrc);
```
bgrozev: Is there a reason for the "video" in the name? The code seems generic.

paweldomas: Yes, it is generic. 'video' is by mistake will fix that, thanks !

src/main/java/org/jitsi/protocol/xmpp/util/SSRCSignaling.java
```diff
@@ -171,41 +196,44 @@ public static String getVideoType(SourcePacketExtension ssrcPe)
      *
      * @return <tt>true</tt> if streams were merged.
      */
-    public static boolean mergeAudioIntoVideo(MediaSSRCMap peerSSRCs)
+    public static boolean mergeVideoIntoAudio(MediaSSRCMap peerSSRCs)
```
bgrozev: The doc needs an update here.

src/main/java/org/jitsi/protocol/xmpp/util/MediaSSRCMap.java
```diff
@@ -181,25 +231,18 @@ public MediaSSRCMap remove(MediaSSRCMap mapToRemove)
...
-        {
-            String media = e.getKey();
-            List<SourcePacketExtension> listCopy
-                = new ArrayList<>(e.getValue());
-
-            mapCopy.put(media, listCopy);
-        }
-
-        return new MediaSSRCMap(mapCopy);
+        return getSSRCsForMedia(media).remove(ssrc);
```
bgrozev: This may not always work as expected because SourcePacketExtension doesn't override equals(). Since there is code which makes copies of these extensions, it could be easy to misuse in the future. I think that we at least need a note that this depends on the specific SourcePacketExtension instance.

paweldomas: Oh yes, will change that to find the SSRC instance to be removed. There is a method for that already.

PR #78, opened 2016-04-27
Implements new stats methods for mock media stream stats.

damencho:
This change is needed if PR https://github.com/jitsi/libjitsi/pull/154 is merged and we update libjitsi in jicofo.


damencho:
Commit https://github.com/jitsi/jicofo/commit/f0748019bd4983d95797aa97cd71c5a4154c4685 duplicates this PR.

Issue #39, opened 2016-02-23
Reservation is not revoked when conference ends


paulvt:
The reservation system REST API documentation mentions the following about conference (reservation) deletion: "Jicofo deletes conferences in the reservation system in two cases. First when all users leave XMPP Multi User Chat room. Second when conference duration limit is exceeded."

I am only experiencing the latter.  With the duration set to 10 minutes, the conference is never deleted before the 10 minutes expire, even if everyone leaves within the minute.

(I see no call to `deleteConference` except from within the `ConferenceExpireTask`.)

paweldomas:
Hi ! Thanks for the report ! Not sure when we'll have the time to fix the issue, but PR fixing it would be welcome !

paweldomas:
Actually it should be called from here as well:
https://github.com/jitsi/jicofo/blob/master/src/main/java/org/jitsi/jicofo/FocusManager.java#L368

paulvt:
Are you sure? Because the `ConferenceExpireTask` class of `RESTReservations` already calls `focusManager.destroyConference()` which then in turn would call `RESTReservation.destroyConference()` again. Seems a bit kludgy to me? Isn't is possible to hook it up somewhere in the `FocusComponent`, which already does the `createConference()` call and has a reference to the reservation system.

paulvt:
I've been digging through the code, but I see that the reservation system does a `focusManager.setFocusAllocationListener(this)` when started.  The `onFocusDestroyed` method on the reservation system thus already should handle conference destruction and deregistration.  However, it seems that the `AbstractAuthAuthority`, and thus the `XMPPDomainAuthAuthority` that we are using, also does a `focusManager.setFocusAllocationListener(this)`! Aren't these calls overwriting each other's listening objects?

paweldomas:
Nice catch ! Thanks for looking into this. I was planning to get rid of FocusAllocationListener in favour of "focus created" and "focus destroyed" events:
https://github.com/jitsi/jicofo/blob/master/src/main/java/org/jitsi/jicofo/FocusManager.java#L325
https://github.com/jitsi/jicofo/blob/master/src/main/java/org/jitsi/jicofo/FocusManager.java#L396
Unfortunately we're not using that code, so it may take a lot of time before that happens, but we would appreciate a PR that does that !

dihedral:
i am also running into this issue.
[...]

PR #82, opened 2016-05-04
Adding focus listeners to a set

rdsmallwood928:
Protecting set on listener set

Removing unneccesary comment

Changing set to list as that is what is specified in the
code...

Some test code to make sure the FocusListeners are getting
destroyed as expected...

Removing debug code

Some debugging code for conference timeouts...

Get room is not a method for conference :/

Debugging conference map

Changing listener to listeners

Reverting 100-Continue stuff


rdsmallwood928:
Currently conference deletions are not being received by our reservation system.  This change would allow for multiple FocusListeners so that the listeners don't overwrite each other. I believe this would also fix this issue https://github.com/jitsi/jicofo/issues/39

FearTheBadger:
We have signed a corporate agreement, can you add me to the list with @rdsmallwood928.
@PetroccoCo, @CPWeaver or @dwilson6 can verify.

PetroccoCo:
I can confirm @Fearthebadger is covered under our corporate agreement.

dihedral:
what happend? is this no longer of interes?

FearTheBadger:
The company who was working on this we bought, and the product was shelved. So those who worked on this have no vested interest.

Not to say that this shouldn't still be merged after fixing the conflict, but we have lost our vestment.

bgrozev:
Thanks for your contribution. I don't think this valid anymore (we already have a list)

PR #87, opened 2016-05-19
Jibri busy and error

paweldomas:
Implements new Jibri states:
- "failed": when Jibri recorder fail with an erron it will send it to Jicofo which will then forward it to the client in the "failed" state
- "busy": when there are Jibris connected to the system, but all are currently in "busy" state then Jicofo will broadcast "busy" state instead of "undefined"

### DIFF
--- a/src/main/java/org/jitsi/jicofo/JitsiMeetGlobalConfig.java
+++ b/src/main/java/org/jitsi/jicofo/JitsiMeetGlobalConfig.java
@@ -56,13 +56,13 @@ public class JitsiMeetGlobalConfig
      * The name of the config property which specifies how long we're going to
      * wait for Jibri to start recording from the time it accepted START request
      */
-    private static final String PENDING_TIMEOUT_PROP_NAME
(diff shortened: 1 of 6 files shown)


src/main/java/org/jitsi/jicofo/JitsiMeetGlobalConfig.java
```diff
@@ -62,7 +62,7 @@
     /**
      * The default value for {@link #PENDING_TIMEOUT_PROP_NAME}.
      */
-    private static final int DEFAULT_PENDING_TIMEOUT = 30;
+    private static final int DEFAULT_PENDING_TIMEOUT = 90;
```
yanas: Minor observation here for the PENDING_TIMEOUT_PROP_NAME. If this is Jicofo's global config it may be good to have JIBRI or RECORDING as part of the name of the property, for example JIBRI_PENDING_TIMEOUT_PROP_NAME. The same for the default value constant.

paweldomas: Good point, will fix that !

PR #89, opened 2016-05-19
Ignore the pre-configured bridge that comes from the client.

gpolitis:
We use pubsub and discoinfo to configure the bridge.


gpolitis:
jenkins: it's ok to test

PR #90, opened 2016-05-20
Jvb down restart

paweldomas:
Bug fixes for ice-restart and implements conference restart, after we get bridge up event.


paweldomas:
It also fixes NPE in expireChannels by setting "colibriConference" field in JitsiMeetConference to null in disposeConference().

gpolitis:
Hi @paweldomas, I don't understand why we need to restart the conference on the bridge UP event. When a bridge goes down, aren't we allocating all of its conferences to the next available bridge? Does this commit handle the case where you have a single bridge?

gpolitis:
I'm merging this because it's urgent.

paweldomas:
We need to handle bridge up event, because there can be situation where we end up without any bridge  for a moment. Then we will want to restart "dead" conference when a bridge goes up.

PR #95, opened 2016-06-16
Refactor starting Jibris out into separate method

SamWhited:
My Java is pretty rusty, but I think this is more or less idiomatic.

Retries are not implemented yet, but it should be relatively easy now that this is a separate method. Code reviews appreciated before I finish adding retries in case my Java is terrible or I'm completely barking up the wrong tree.

Sorry for all the import changes; the `*`'s were bothering me because I couldn't tell what package any identifiers came from. I can remove that from the diff before merge if it's not desired.

### DIFF
--- a/src/main/java/org/jitsi/jicofo/recording/jibri/JibriRecorder.java
+++ b/src/main/java/org/jitsi/jicofo/recording/jibri/JibriRecorder.java
@@ -17,21 +17,30 @@
  */
 package org.jitsi.jicofo.recording.jibri;
 
-import net.java.sip.communicator.impl.protocol.jabber.extensions.colibri.*;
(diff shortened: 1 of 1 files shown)


bgrozev:
LGTM, except for the star imports and some minor formatting.

bgrozev:
jenkins: add to whitelist

paweldomas:
I've merged the PR and added some minor fixes on top of that. One more important was to remove synchronised statement from startJibri, as currently it's not necessary. The method is private and synchronisation is done on public methods called by other threads: processPacket, onJibriStatusChanged and onJibriOffline. We'll see how it will look like when retries are added and eventually add it later.

PR #99, opened 2016-06-22
HQL-1502: Retry recording with new Jibri on fail

SamWhited:
Still having a bit of trouble testing this because of a problem with Jibri (if I kill the XMPP process, FFMPEG doesn't get cleaned up and continues recording, so any New Jibris that are selected can't send to YouTube because it's already in use), but I think it's working now. It's definitely retrying with a different Jibri anyways.

*EDIT:* I have no idea why GitHub thinks there are conflicts; this appears to merge cleanly.


SamWhited:
Something weird is going on with GitHub; it's not even showing me the correct branch for this. I think it has an old branch by the same name cached.

PR #100, opened 2016-06-22
HQL-1502: Retry recording with new Jibri on fail

SamWhited:
Still having a bit of trouble testing this because of a problem with Jibri (if I kill the XMPP process, ffmpeg doesn't get cleaned up and continues recording, so any New Jibris that are selected can't send to YouTube because it's already in use), but I think it's working now. It's definitely retrying with a different Jibri anyways.

### DIFF
--- a/src/main/java/org/jitsi/jicofo/recording/jibri/JibriRecorder.java
+++ b/src/main/java/org/jitsi/jicofo/recording/jibri/JibriRecorder.java
@@ -94,6 +94,11 @@ public class JibriRecorder
     private ScheduledFuture<?> pendingTimeoutTask;
 
     /**
+     * The stream ID received in the first Start IQ.
+     */
+    private String streamID;
+
+    /**
      * Creates new instance of <tt>JibriRecorder</tt>.
      * @param conference <tt>JitsiMeetConference</tt> to be recorded by new
      *        instance.
@@ -197,25 +202,11 @@ public class JibriRecorder
     synchronized private XMPPError startJibri(
         final String jibriJid
     ) throws SmackException.NoResponseException {
-        return startJibri(jibriJid, "");
-    }
-
-    /**
-     * Sends an IQ to the given Jibri instance and asks it to start recording.
-     */
-    synchronized private XMPPError startJibri(
-        final String jibriJid,
-        final String streamId
-    ) throws SmackException.NoResponseException {
         JibriIq startIq = new JibriIq();
         startIq.setTo(jibriJid);
         startIq.setType(IQ.Type.SET);
         startIq.setAction(JibriIq.Action.START);
-
-        if ("".equals(streamId))
-        {
-            startIq.setStreamId(streamId);
-        }
+        startIq.setStreamId(this.streamID);
 
         // Insert name of the room into Jibri START IQ
         startIq.setRoom(conference.getRoomName());
@@ -308,6 +299,11 @@ public class JibriRecorder
 
     private void processJibriIqFromMeet(final JibriIq iq, final XmppChatMember sender)
     {
+        if (this.streamID == null)
+        {
+            this.streamID = iq.getStreamId();
+        }
+
         JibriIq.Action action = iq.getAction();
 
         if (JibriIq.Action.UNDEFINED.equals(action))
@@ -648,14 +644,7 @@ public class JibriRecorder
                 {
                     try
                     {
-                        if (iq == null)
-                        {
-                            err = startJibri(jibriJid);
-                        }
-                        else
-                        {
-                            err = startJibri(jibriJid, iq.getStreamId());
-                        }
+                        err = startJibri(jibriJid);
                     }
                     catch (final SmackException.NoResponseException e)
                     {
@@ -669,7 +658,10 @@ public class JibriRecorder
                         break;
                     }
                     // Mask the original error in the IQ response.
-                    logger.info("Masking Jibri error, would have sent " + err.getCondition());
+                    if (err != null)
+                    {
+                        logger.info("Masking Jibri error, would have sent " + err.getCondition());
+                    }
                     err = new XMPPError(XMPPError.Condition.interna_server_error);
                 }
                 if (err != null && iq != null)
--- a/src/main/java/org/jitsi/jicofo/recording/jibri/JibriRecorder.java
+++ b/src/main/java/org/jitsi/jicofo/recording/jibri/JibriRecorder.java
@@ -53,6 +53,11 @@ public class JibriRecorder
     static private final Logger logger = Logger.getLogger(JibriRecorder.class);
 
     /**
+     * The number of times to retry connecting to a Jibri.
+     */
+    static private final int NUM_RETRIES = 3;
+
+    /**
      * Recorded <tt>JitsiMeetConference</tt>.
      */
     private final JitsiMeetConference conference;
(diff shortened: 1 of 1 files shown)


src/main/java/org/jitsi/jicofo/recording/jibri/JibriRecorder.java
```diff
@@ -619,13 +639,44 @@ synchronized public void onJibriStatusChanged(String jibriJid, boolean idle)
...
+            // Check if we have Jibri available
+            final String newJibriJid = jibriDetector.selectJibri();
+            if (newJibriJid == null)
+            {
+                recordingStopped(null);
+                return;
+            }
+
+            XMPPError err = null;
+            for (int i = 0; i < NUM_RETRIES; i++)
```
paweldomas: The loops is duplicated. Would it be possible to merge that into a method ?

src/main/java/org/jitsi/jicofo/recording/jibri/JibriRecorder.java
```diff
@@ -619,13 +639,44 @@ synchronized public void onJibriStatusChanged(String jibriJid, boolean idle)
...
+                catch (final SmackException.NoResponseException e)
+                {
+                    err = new XMPPError(XMPPError.Condition.request_timeout);
+                }
+
+                if (err == null) {
+                    break;
+                }
+
+                // Mask the original error in the IQ response.
```
paweldomas: If we are masking the original error, can we at least log it for debugging purposes ?

src/main/java/org/jitsi/jicofo/recording/jibri/JibriRecorder.java
paweldomas: Or maybe move the loop to startJibri ?

src/main/java/org/jitsi/jicofo/recording/jibri/JibriRecorder.java
```diff
@@ -323,31 +340,34 @@ private void processJibriIqFromMeet(JibriIq iq, XmppChatMember sender)
...
-                    iq, XMPPError.Condition.service_unavailable, null);
+                        iq, XMPPError.Condition.service_unavailable, null);
                 return;
             }
 
-            final String roomName = MucUtil.extractName(senderMucJid);
-
-            try
+            XMPPError err = null;
+            for (int i = 0; i < NUM_RETRIES; i++)
```
paweldomas: Also I am a bit worried about blocking packet processor thread on this loop. It will be called either from onJibriOffline or processJibriIqFromMeet where both are triggered by Smack's PacketListener. Currently timeout for packets sent in smack is set to 15 seconds which means that the thread can be blocked for up to 45 seconds if we retry 3 times. This will block packet processing in all other parts of Jicofo.

SamWhited: Oh yah, that would probably be bad. Will have a rethink when I get back to this PR.

src/main/java/org/jitsi/jicofo/recording/jibri/JibriRecorder.java
SamWhited: Done.

src/main/java/org/jitsi/jicofo/recording/jibri/JibriRecorder.java
SamWhited: Done.

src/main/java/org/jitsi/jicofo/recording/jibri/JibriRecorder.java
SamWhited: Done.

src/main/java/org/jitsi/jicofo/recording/jibri/JibriRecorder.java
```diff
@@ -323,33 +340,11 @@ private void processJibriIqFromMeet(JibriIq iq, XmppChatMember sender)
...
-                            iq, XMPPError.Condition.interna_server_error, null);
-                }
-            }
-            catch (final SmackException.NoResponseException e)
-            {
-                sendErrorResponse(
-                        iq, XMPPError.Condition.request_timeout, null);
-                return;
-            }
+            scheduleRetryStartJibri(jibriJid, iq);
```
paweldomas: There is no return after this statement and line 368 is always triggered which results in bad-request

paweldomas: The whole else if should be changed to get rid of returns which are error prone(that is not related to the changes in the PR itself, but was like that from the start)

src/main/java/org/jitsi/jicofo/recording/jibri/JibriRecorder.java
```diff
@@ -189,20 +194,32 @@ public boolean accept(Packet packet)
...
+        final String jibriJid,
+        final String streamId
+    ) throws SmackException.NoResponseException {
         JibriIq startIq = new JibriIq();
         startIq.setTo(jibriJid);
         startIq.setType(IQ.Type.SET);
         startIq.setAction(JibriIq.Action.START);
 
-        startIq.setStreamId(streamId);
+        if ("".equals(streamId))
```
paweldomas: The if works: if streamId equals to empty string then set it in the iq, but otherwise it is left empty.

paweldomas:
The code did not worked at all. I've tried working on the PR, but it looks like the problem is even more complex. It will be better to get rid of separate thread and the "retry for" loop. Will be better to handle everything from packet callbacks and count retry attempts in class field.

src/main/java/org/jitsi/jicofo/recording/jibri/JibriRecorder.java
SamWhited: > There is no return after this statement and line 368 is always triggered which results in bad-request

:oops: fixed

> The whole else if should be changed to get rid of returns which are error prone(that is not related to the changes in the PR itself, but was like that from the start)

I've never really bought into the "one return at the end of the function" Java dogma in general, but as far as I know there's no way to ensure resources get cleaned up at the end of a particular scopes lifetime in Java, so maybe this is where the "returns are bad" thing comes from and it actually makes sense here? Or is there a way to insert a stack frame above the current one or something so that it will run when a function ends that I don't know about?

src/main/java/org/jitsi/jicofo/recording/jibri/JibriRecorder.java
SamWhited: I think this was deliberate, but I can't remember why… will investigate.

paweldomas: Btw. we have to store the streamId in a field when start IQ is received in order to be able to restart when the first Jibri fails after a while.

src/main/java/org/jitsi/jicofo/recording/jibri/JibriRecorder.java
paweldomas: In this case returns combined with multiple "else if" structure make it easy to forget about "return" in the right place.

I'm not aware of method that will "insert a stack frame above the current one". Maybe wrapping the thing in a parent method or using finally block, but not sure if I understand what you meant correctly.

SamWhited: I was just trying to think of a way to ensure resources get cleaned up at the end of a function (regardless of how or where the function returns) since that was the only reason I could think of for returns being bad. I suppose that doing it this way wouldn't work because of how Java handles exceptions though. I just discovered Java 7's "try with resource"; I suppose that's the closest Java thing.

paweldomas:
Some more about dealing with restart from the callbacks. When we start Jibri we can send the packet and release the current thread without waiting for response using https://github.com/jitsi/jicofo/blob/master/src/main/java/org/jitsi/protocol/xmpp/XmppConnection.java#L34.At the same time we can respond with RESULT immediately to the start IQ. Then when we get ERROR response from Jibri we'll just stop with the error received. In order to handle this ERROR response packet we'll have to extend current packet filter condition to accept IQs from "recorderComponentJid" in the acceptPacket method: https://github.com/jitsi/jicofo/blob/master/src/main/java/org/jitsi/jicofo/recording/jibri/JibriRecorder.java#L186
In the place where we process failures the retry logic can be added which will count the attempt in class field.
Does it make sense ?

src/main/java/org/jitsi/jicofo/recording/jibri/JibriRecorder.java
SamWhited: Odd, I thought I'd done that but it's not there… maybe I lost a commit when I changed branches a while back. Done.

PR #101, opened 2016-06-28
Jwt login service

paweldomas:
Integrates external login service which provides JWT token once the user authenticates.


newboy2004:
how can i get jwt token  from jicofo ,if prosody have been authticated for  jwt enable? tks

PR #103, opened 2016-07-19
Shared document using room name config

yanas:


yanas:
I've made sure that it won't return null and the null case is taken care of in the JitsiMeetConfig file.

paweldomas:
How do we want the room name to look like ? Currently the value passed there is full MUC address, do we want to have there just the name ? It is all lower case.

yanas:
It should be just the room name all lower case.

paweldomas:
jenkins test this please

PR #111, opened 2016-08-23
fix(ChannelAllocator): reInvite = JingleSession != null

paweldomas:
If the Participant has JingleSession instance set it means that the
'session-accept' was received and processed already. Initialize
'reInvite' field early to avoid inconsistencies and confusing log
messages about the 'transport-replace' when it is really
the 'session-accept' being sent.


bgrozev:
The flow around setting JingleSession changed. I set it early now, and added a state (ACTIVE, PENDING, ENDED). I don't think this PR applies anymore.

PR #117, opened 2016-08-31
Sync fixes

bgrozev:


src/main/java/org/jitsi/impl/protocol/xmpp/ChatRoomImpl.java
```diff
@@ -889,21 +889,25 @@ private ChatMemberImpl addMember(String participant)
     {
         ChatMemberImpl newMember;
 
-        if (members.containsKey(participant))
+        synchronized (members)
```
paweldomas: I'm not sure I agree that we really need that. addMember is called from one place it should remain like that forever.

src/main/java/org/jitsi/impl/protocol/xmpp/OpSetSubscriptionImpl.java
```diff
@@ -177,11 +176,21 @@ private boolean isSubscribed(String jid, Node node)
     public synchronized void subscribe(String               node,
```
paweldomas: Both subscribe and unsubscribe are synchronised on the instance already.

src/main/java/org/jitsi/impl/reservation/rest/RESTReservations.java
```diff
@@ -217,7 +216,7 @@ else if (result.statusCode == 409
     /**
      * Deletes conference for given <tt>mucRoomName</tt> through the API.
      */
-    Result deleteConference(String mucRoomName)
+    private synchronized Result deleteConference(String mucRoomName)
```
paweldomas: If methods are private and run within the synchronised I disagree we want to wrap it again.

bgrozev: My thinking is: The method requires the lock, so it should be annotated. The fact that it is currently only called from places which already hold the lock is incidental. Specifically, I have two problems with the missing lock:
1. It is harder to verify that the code is safe (I had to follow the usage). 
2. It is a lot easier to screw it up (in a subtle and hard to find manner) when adding new code.

paweldomas: It is not "incidental" if it was written this way in purpose to take care of access synchronisation in the public methods, but I agree it may not be readable enough.

src/main/java/org/jitsi/impl/protocol/xmpp/ChatRoomImpl.java
```diff
@@ -58,7 +58,8 @@
      * Caches early presence packets triggered by Smack, before there was member
      * joined event.
      */
-    private Map<String, Presence> presenceCache = new HashMap<>();
+    private final Map<String, Presence> presenceCache
+        = new ConcurrentHashMap<>();
```
lyubomir: I'm not sure that it's much better than before. It's true that the presenceCache will be consistently modified as far as the generic Map interface goes. But will it be consistently modified as far as ChatRoomImpl goes? I'm not sure. In 2 out of 3 uses presenceCache is protected by members, in the remaining 1 out of 3 uses it's not protected by members. Murky business.

src/main/java/org/jitsi/impl/reservation/rest/RESTReservations.java
```diff
@@ -283,7 +282,7 @@ private int deleteConference(Number id)
      * @return <tt>Conference</tt> for given <tt>id</tt> or <tt>null</tt>
      *         if not found.
      */
-    private Conference findConferenceForId(Number id)
+    private synchronized Conference findConferenceForId(Number id)
```
lyubomir: After adding private to deleteConference, the synchronized here becomes the same: a question of readability.

src/main/java/org/jitsi/jicofo/BridgeSelector.java
```diff
@@ -91,7 +92,8 @@
     /**
      * The map of bridge JID to <tt>BridgeState</tt>.
      */
-    private final Map<String, BridgeState> bridges = new HashMap<>();
+    private final Map<String, BridgeState> bridges
+        = new ConcurrentHashMap<>();
```
lyubomir: I'm not sure why this is being done. Baring the readability and possibly the maintenance issues of not having synchronized on private methods, it looks like bridges is not protected only in getKnownBridgesCount()? I think that having bridges as a ConcurrentHashMap given all other `synchronized` all over the place makes one question why 2 things are being done when 1 would have sufficed.

src/main/java/org/jitsi/jicofo/BridgeSelector.java
```diff
@@ -102,7 +104,7 @@
     /**
      * The map of Pub-Sub nodes to videobridge JIDs.
      */
-    private final Map<String, String> pubSubToBridge = new HashMap<>();
+    private final Map<String, String> pubSubToBridge = new ConcurrentHashMap<>();
```
lyubomir: 2 out of 3 uses are protected by a synchronized on this, 1 out of 3 is not protected & it's in the init method of the instance (which presumably makes the instance usable). I'm not sure what practical advantages ConcurrentHashMap brings.

src/main/java/org/jitsi/jicofo/FocusManager.java
```diff
@@ -117,13 +118,13 @@
...
-    private FocusExpireThread expireThread = new FocusExpireThread();
+    private final FocusExpireThread expireThread = new FocusExpireThread();
 
     /**
      * Jitsi Meet conferences mapped by MUC room names.
      */
-    private Map<String, JitsiMeetConference> conferences
-        = new HashMap<String, JitsiMeetConference>();
+    private final Map<String, JitsiMeetConference> conferences
+        = new ConcurrentHashMap<>();
```
lyubomir: Is that because of getConferenceCount()? There are explicit `synchronized (this)` (including a `synchronized (FocusManager.this)`) which seem to be introduced for the sole purpose of protecting `conferences` so adding ConcurrentHashMap results in an unnecessary confusion or at the very least a doubt.

src/main/java/org/jitsi/jicofo/Participant.java
```diff
@@ -112,7 +113,7 @@
      * returns <tt>false</tt>.
      */
     private Map<String, IceUdpTransportPacketExtension> transportMap
-        = new HashMap<>();
+        = new ConcurrentHashMap<>();
```
lyubomir: What is it that the ConcurrentHashMap solves exactly? I see a check on a get which decides whether to perform a put and my concern in mostly whether get and put will be protected as a whole block, not whether they are individually protected. Anyway, I need more clarity on the goal here.

src/main/java/org/jitsi/jicofo/auth/AbstractAuthAuthority.java
```diff
@@ -83,8 +83,8 @@
     /**
      * The map of user JIDs to {@link AuthenticationSession}.
      */
-    private Map<String, AuthenticationSession> authenticationSessions
-            = new HashMap<String, AuthenticationSession>();
+    private final Map<String, AuthenticationSession> authenticationSessions
+            = new ConcurrentHashMap<>();
```
lyubomir: Is this because of getSession(String)? Because authenticationSessions is explicitly synchronized on syncRoot otherwise.

src/main/java/org/jitsi/protocol/xmpp/util/MediaSSRCGroupMap.java
```diff
@@ -33,22 +34,22 @@
     /**
      * Map backend.
      */
-    private final Map<String, List<SSRCGroup>> groupMap;
+    private final ConcurrentHashMap<String, List<SSRCGroup>> groupMap;
```
lyubomir: What's the purpose of this? The whole MediaSSRCGroupMap is thread-unsafe so is this single change supposed to make it thread-safe? I don't think it does but then the purpose that you had in mind may be different.

src/main/java/org/jitsi/protocol/xmpp/util/MediaSSRCMap.java
```diff
@@ -41,7 +41,7 @@
      */
     public MediaSSRCMap()
     {
-        this(new HashMap<String, List<SourcePacketExtension>>());
+        this(new ConcurrentHashMap<String, List<SourcePacketExtension>>());
```
lyubomir: I have the same question(s) as on MediaSSRCGroupMap.

src/main/java/org/jitsi/protocol/xmpp/util/MediaSSRCGroupMap.java
paweldomas: Both MediaSSRCGroupMap and MediaSSRCMap classes are thread unsafe just like the HashMap is. So they should be synchronised on a higher level.

src/main/java/org/jitsi/impl/protocol/xmpp/ChatRoomImpl.java
bgrozev: OK, I'm replacing this with a comment...

src/main/java/org/jitsi/jicofo/BridgeSelector.java
bgrozev: Replacing this with making getKnownBridgesCount() synchronized (I now see this is only used in tests).

src/main/java/org/jitsi/jicofo/BridgeSelector.java
bgrozev: Removing the change.

src/main/java/org/jitsi/jicofo/FocusManager.java
bgrozev: Yes, it is because of getConferenceCount(). I was reluctant to synchronize on the FocusManager object. If all synchronization on the FocusManager object is solely for the purpose of protecting #conferences, then I would rather synchronize on #conferences instead everywhere. But I wasn't able to convince myself that this is the case (e.g. enableGracefulShutdownMode() seems to need it because of #shutdownInProgress).

src/main/java/org/jitsi/jicofo/Participant.java
bgrozev: My idea was to protect the individual calls to get() and put(), since the map is exposed through getTransportMap().

src/main/java/org/jitsi/jicofo/auth/AbstractAuthAuthority.java
bgrozev: Yes.

bgrozev: I'm adding a comment.

src/main/java/org/jitsi/protocol/xmpp/util/MediaSSRCGroupMap.java
bgrozev: Removing the commit (not including it in the new version of the PR).

src/main/java/org/jitsi/protocol/xmpp/util/MediaSSRCMap.java
bgrozev: Removing the commit (not including it in the new version of the PR).

bgrozev:
Moved to #122

PR #120, opened 2016-09-02
Timeout lonely participant's media session who remain in the MUC

paweldomas:
After the call ends the person who remains alone in the room will have it's media session expired after 20 seconds to prevent from wasting the bandwidth/resources.

### DIFF
--- a/src/main/java/org/jitsi/jicofo/JitsiMeetConference.java
+++ b/src/main/java/org/jitsi/jicofo/JitsiMeetConference.java
@@ -335,7 +335,7 @@ public class JitsiMeetConference
         }
         catch(Exception e)
         {
-            this.stop();
+            stop();
 
             throw e;
         }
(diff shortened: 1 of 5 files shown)


src/main/java/org/jitsi/jicofo/JitsiMeetConference.java
```diff
@@ -1682,4 +1721,81 @@ private String createSharedDocumentName()
 
         return sharedDocumentName;
     }
+
+    /**
+     * (Re)schedules {@link SinglePersonTimeout}.
+     */
+    private void rescheduleSinglePeerTimeout()
+    {
+        if (this.executor != null)
```
hristoterezov: In this method the properties are accessed with `this` and in the next method the same properties are accessed without `this`. It's better to keep the same code style.

paweldomas: done !

PR #135, opened 2016-10-26
Fix problems in JibriDetector on XMPP reconnect

paweldomas:
1. fixes some threads dying while trying to send packets on broken XMPP connection(may lead to weird state left)
2. fixes invalid Jibri addresses kept in the memory after XMPP reconnection

### DIFF
--- a/src/main/java/org/jitsi/jicofo/recording/jibri/JibriDetector.java
+++ b/src/main/java/org/jitsi/jicofo/recording/jibri/JibriDetector.java
@@ -222,14 +222,14 @@ public class JibriDetector
             chatRoom = null;
 
             logger.info("Left JIBRI room: " + jibriBrewery);
+        }
 
-            // Clean up the list of Jibris
-            List<Jibri> jibrisCopy = new ArrayList<>(jibris);
-            jibris.clear();
-            for (Jibri jibri : jibrisCopy)
-            {
-                notifyJibriOffline(jibri.mucJid);
-            }
+        // Clean up the list of Jibris
+        List<Jibri> jibrisCopy = new ArrayList<>(jibris);
+        jibris.clear();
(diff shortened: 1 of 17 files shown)


src/main/java/org/jitsi/jicofo/recording/jibri/JibriRecorder.java
```diff
@@ -197,10 +197,19 @@ public void init()
...
+            {
+                logger.error(
+                    "An error response to the stop request: " + error.toXML());
+            }
+        }
+        catch (OperationFailedException e)
         {
-            logger.error("Error when sending stop request: " + error.toXML());
+            logger.error("Failed to send stop IQ - XMPP disconnected", e);
+            recordingStopped(null, false /* do not send any status updates */);
```
bgrozev: Is it intentional that we call recordingStopped on exeption, but not when "error != null" above?

src/main/java/org/jitsi/jicofo/recording/jibri/JibriDetector.java
```diff
@@ -222,6 +222,14 @@ private void stop()
             chatRoom = null;
 
             logger.info("Left JIBRI room: " + jibriBrewery);
+
```
bgrozev: Is there any particular reason this only executes if "chatRoom != null"?

paweldomas: No, will fix that, thanks !

src/main/java/org/jitsi/jicofo/recording/jibri/JibriRecorder.java
paweldomas: It got kind of complex now. Will take a second look at this.

bgrozev:
Jenkins: ok to test

PR #137, opened 2016-11-01
Process XMPP packets in order

paweldomas:
Updates Jicoco in order to use the single thread packet processor. DO NOT MERGE YET(Jicoco is referenced by the SNAPSHOT version)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -73,7 +73,7 @@
     <dependency>
       <groupId>${project.groupId}</groupId>
       <artifactId>jicoco</artifactId>
-      <version>1.0-SNAPSHOT</version>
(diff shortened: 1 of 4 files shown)


pom.xml
```diff
@@ -73,7 +73,7 @@
     <dependency>
       <groupId>${project.groupId}</groupId>
       <artifactId>jicoco</artifactId>
-      <version>1.0-20160420.185019-32</version>
+      <version>1.0-SNAPSHOT</version>
```
bgrozev: Is this intentionally SNAPSHOT and not just the new version?

paweldomas: yes, waiting for Jicoco to be merged first and will update that

PR #138, opened 2016-11-02
Configure colibri channels according to rtpLevelRelayType property

matteocampana-zz:
When colibri channels are created, use the rtpLevelRelayType config property to configure rtp-level-relay-type channel element.

(PR recreated after rebase).


bgrozev:
Thanks for your contribution. I don't think the issue is valid anymore (colibri1 has been removed)

PR #142, opened 2016-11-06
Minor clean-up: don't use Assert.notNull

bgrozev:
Uses Objects.requireNonNull instead of Assert.notNull.

### DIFF
--- a/src/main/java/org/jitsi/impl/protocol/xmpp/OpSetSubscriptionImpl.java
+++ b/src/main/java/org/jitsi/impl/protocol/xmpp/OpSetSubscriptionImpl.java
@@ -311,10 +311,9 @@ public class OpSetSubscriptionImpl
          */
         public Subscription(String node, SubscriptionListener listener)
         {
-            Assert.notNull(node, "node");
-            Assert.notNull(listener, "listener");
+            Objects.requireNonNull(listener, "listener");
+            this.node = Objects.requireNonNull(node, "node");
 
-            this.node = node;
             this.listeners.add(listener);
             this.retryStrategy
                 = new RetryStrategy(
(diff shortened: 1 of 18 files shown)


src/main/java/org/jitsi/impl/protocol/xmpp/OpSetSubscriptionImpl.java
```diff
@@ -311,10 +311,9 @@ public void handlePublishedItems(ItemPublishEvent event)
          */
         public Subscription(String node, SubscriptionListener listener)
         {
-            Assert.notNull(node, "node");
-            Assert.notNull(listener, "listener");
+            Objects.requireNonNull(listener, "listener");
+            this.node = Objects.requireNonNull(node, "node");
 
-            this.node = node;
             this.listeners.add(listener);
```
paweldomas: Will this work here this.listeners.add(Objects.requireNotNull(listener, "listener") ?

src/main/java/org/jitsi/jicofo/rest/Health.java
```diff
@@ -253,10 +253,7 @@ private static int cacheStatus(int status)
         JitsiMeetServices services
             = focusManager.getJitsiMeetServices();
 
-        Assert.notNull(services, "services");
```
paweldomas: The reason I've put those here was to get a message in NPE indicating which variable is null(to know even without looking at the line number and checking the code). I don't insist that those are necessary just explaining my reasoning behind that. If you think it's not of much use I'm good with having those deleted.

bgrozev: OK, they looked superfluous to me, but that makes sense. Will add them back.

src/main/java/org/jitsi/impl/protocol/xmpp/OpSetSubscriptionImpl.java
bgrozev: Yes, but I find it harder to read.

PR #145, opened 2016-11-10
Fix few things after removing InfluxDb logging

paweldomas:


pom.xml
```diff
@@ -147,12 +147,17 @@
     <dependency>
       <groupId>junit</groupId>
       <artifactId>junit</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>${project.groupId}</groupId>
```
hristoterezov: Is this dependancy related to the changes? Do we need it?

bgrozev: Wild guess: could be that the ice4j version which ends up being used is the one libjitsi depends on. We should update libjitsi's pom.xml and see if that helps.

paweldomas: hmm for some reason it selects ice4j 1.0 instead of ice4j 1.1 even if I bump the version in libjitsi. There is also jitsi-netaddr which depends on ice4j, but it does not have any version specified.

PR #156, opened 2017-01-16
doc: fix broken link

saghul:


damencho:
This doesn't need testing. It fails cause it searches for a repo from another pr which uses same branch name patch-1 :)

PR #157, opened 2017-01-18
Fix simulcast lipsync hack

paweldomas:
All video SSRCs have to be merged by the lipsync hack if simulcast is enabled.

### DIFF
--- a/src/main/java/org/jitsi/protocol/xmpp/util/SSRCSignaling.java
+++ b/src/main/java/org/jitsi/protocol/xmpp/util/SSRCSignaling.java
@@ -53,7 +53,7 @@ public class SSRCSignaling
      *            copy parameter value.
      * @param name the name of the parameter to copy.
      */
-    public static void copyParamAttr( SourcePacketExtension    dst,
+    private static void copyParamAttr( SourcePacketExtension    dst,
                                       SourcePacketExtension    src,
                                       String                  name)
     {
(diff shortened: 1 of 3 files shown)


src/main/java/org/jitsi/protocol/xmpp/util/SSRCSignaling.java
```diff
@@ -245,35 +247,42 @@ public static boolean mergeVideoIntoAudio(MediaSSRCMap peerSSRCs)
...
-        copyParamAttr(videoSSRC, audioSSRC, "mslabel");
+        // Find first video SSRC with non-empty stream ID and different
+        // than 'default' which is sometimes used when unspecified
+        List<SourcePacketExtension> videoSSRCs
+            = peerSSRCs.getSSRCsForMedia(MediaType.VIDEO.toString());
 
-        videoMsid.setValue(audioStreamId + " " + videoTrackId);
+        boolean merged = false;
+        // There are multiple video SSRCs in simulcast
+        // FIXME this will not work with more than 1 video stream
```
bbaldino: does that mean this will break when rtx ssrcs are in the sdp as well?

paweldomas: do RTX SSRCs belong to the same stream (do they have the same msid) ? if the answer is *yes* then it should be fine as it's the same as in simulcast. If you were asking if that would break *before the PR* then yes that could break. So far we were lucky that the client was taking the msid value only from the first SSRC, but I saw it randomly breaking while working on P2P.

PR #159, opened 2017-01-25
Fix NPE in reservation system when there's no authAuthority

debugmaster:
I am using Jicofo's reservation system to generate a PIN to each conference to integrate with Jigasi.

However, I get a NPE in [this line](https://github.com/jitsi/jicofo/blob/8f165d52b2cf4d34925de09e120fb13632ebcd18/src/main/java/org/jitsi/impl/reservation/rest/RESTReservations.java#L201) for every new conference. 

Looking at the code, I found out that the reason for getting a NPE is that identity/creator variable is set to null by default in [this line](https://github.com/jitsi/jicofo/blob/master/src/main/java/org/jitsi/jicofo/xmpp/FocusComponent.java#L361) and it does not change its value if there is no authentication system, which is my case.  

This PR allows the conference to be created when there isn't an authentication system. 
Tell me if this is not a desirable behavior, so I can work on a fix.

### DIFF
--- a/src/main/java/org/jitsi/impl/reservation/rest/RESTReservations.java
+++ b/src/main/java/org/jitsi/impl/reservation/rest/RESTReservations.java
@@ -201,7 +201,7 @@ public class RESTReservations
         if (creator == null)
         {
             logger.warn(
-                "Room " + mucRoomName + " was created without a creator"
+                "Room " + mucRoomName + " was created without a creator");
             return new Result(RESULT_OK);
         }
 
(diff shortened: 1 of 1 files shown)


paweldomas:
It should be fine as long as the reservation system is fine with having null as the creator. Jenkins test this please.

debugmaster:
I've signed it.

debugmaster:
Ops. wrong button [again!]. 

I've signed it.

paweldomas:
Jenkins test this please.

paweldomas:
Compilation failure:
14:21:11 [ERROR] /home/jenkins/jicofo/src/main/java/org/jitsi/impl/reservation/rest/RESTReservations.java:[204,73] ')' expected

paweldomas:
Jenkins test this please.

debugmaster:
Sorry for the );

Wow, that's a lot of tests. I will check.

paweldomas:
No, these test are failing randomly recently. I've restarted that

paweldomas:
Hmm looks like the tests are broken, but it doesn't seem to be related with this PR... Will try again once we fix the problem.

PR #165, opened 2017-02-07
Offer h264 (after vp8) so interested clients can use it

bbaldino:


bbaldino:
thanks @paweldomas. the two branches are independent, so that's why i didn't name them the same (we don't need to block merging this one on the other...if this gets merged then h264 will be after vp8 so shouldn't change any behavior of any clients that are already doing vp8).

PR #166, opened 2017-02-08
Verify SSRCs and groups

paweldomas:
Checks the following, before accepting SSRCs to the conference description:
1. Allow SSRC value between 1 and 0xFFFFFFFF
2. Do not allow the same SSRC to appear more than once per media type
3. Skip SSRCs if the SSRC count limit is exceeded
4. Filter out SSRC parameters other than 'cname' and 'msid'
5. Drop empty SSRC groups
6. Skip duplicated groups (the same semantics and SSRCs)
7. Look for MSID conflicts between SSRCs which do not belong to the same
    group
8. Make sure that SSRCs described by groups exist in media description

Some of the above were checked even before the PR changes.

### DIFF
--- a/src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
+++ b/src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
@@ -18,6 +18,7 @@
 package org.jitsi.jicofo;
 
 import net.java.sip.communicator.impl.protocol.jabber.extensions.colibri.*;
+import net.java.sip.communicator.impl.protocol.jabber.extensions.jibri.*;
 import net.java.sip.communicator.impl.protocol.jabber.extensions.jingle.*;
 import net.java.sip.communicator.service.protocol.*;
 import net.java.sip.communicator.service.protocol.event.*;
@@ -33,6 +34,8 @@ import org.jitsi.eventadmin.*;
 
 import org.jitsi.util.*;
 import org.jitsi.util.Logger;
+import org.jivesoftware.smack.*;
+import org.jivesoftware.smack.packet.*;
 import org.osgi.framework.*;
 
 import java.text.*;
@@ -935,18 +938,19 @@ public class JitsiMeetConferenceImpl
      * {@inheritDoc}
      */
     @Override
-    public void onSessionAccept( JingleSession peerJingleSession,
+    public XMPPError onSessionAccept( JingleSession peerJingleSession,
                                  List<ContentPacketExtension> answer)
     {
         Participant participant
             = findParticipantForJingleSession(peerJingleSession);
-        String peerJingleSessionAddress = peerJingleSession.getAddress();
+        String peerAddress = peerJingleSession.getAddress();
 
         if (participant == null)
         {
-            logger.error(
-                    "No participant found for: " + peerJingleSessionAddress);
-            return;
+            String errorMsg
+                = "No participant found for: " + peerAddress;
+            logger.error(errorMsg);
+            return new XMPPError(XMPPError.Condition.item_not_found, errorMsg);
         }
 
         if (participant.getJingleSession() != null)
@@ -954,7 +958,7 @@ public class JitsiMeetConferenceImpl
             //FIXME: we should reject it ?
             logger.error(
                     "Reassigning jingle session for participant: "
-                        + peerJingleSessionAddress);
+                        + peerAddress);
         }
 
         // XXX We will be acting on the received session-accept bellow.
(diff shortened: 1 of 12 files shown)


bgrozev:
>2. Do not allow the same SSRC to appear more than once per media type

With bundle we use the SSRC to find the media type of an RTCP packet (on the bridge, that is), so I think we should not allow the same SSRC in the same conference regardless of media type.

gpolitis:
> 1. Allow SSRC value between 1 and 0xFFFFFFFF

Probably not that much big of a deal, but 0 is a valid SSRC. If we represent SSRCs with `long`s, then we can use -1 as an invalid value.

paweldomas:
> Probably not that much big of a deal, but 0 is a valid SSRC. If we represent SSRCs with longs, then we can use -1 as an invalid value.

Technically you're right, but Chrome seems to break when it sees 0. First it accept the remote description, but does not display video. Then any subsequent cals to setRemoteDescription fail.

src/test/java/org/jitsi/jicofo/ParticipantTest.java
```diff
@@ -44,88 +42,397 @@
...
+            createGroupSSRC(50L, cname, msid),
+            createGroupSSRC(60L, cname, msid)
+        };
+
+        this.videoGroups = new SourceGroupPacketExtension[] {
+            createGroup(
+                SourceGroupPacketExtension.SEMANTICS_SIMULCAST,
+                new long[] { 10L, 30L, 50L },
+                cname, msid),
+            createGroup("FIR", new long[] { 10L, 20L }, cname, msid),
```
bgrozev: Where does "FIR" come from? Just a random string?

src/test/java/org/jitsi/jicofo/ParticipantTest.java
```diff
@@ -44,88 +42,397 @@
...
+        this.addDefaultAudioSSRCs();
+
+        try
+        {
+            participant.addSSRCsAndGroupsFromContent(answerContents);
+            fail("Did not detect SSRC 0 as invalid");
+        }
+        catch (InvalidSSRCsException exc)
+        {
+            assertEquals("Illegal SSRC value: 0", exc.getMessage());
```
bgrozev: This seems error prone, maybe move the string to a constant?

src/test/java/org/jitsi/jicofo/ParticipantTest.java
paweldomas: No, it should be FID. Thanks !

src/test/java/org/jitsi/jicofo/ParticipantTest.java
paweldomas: I had the same thought initially, but then when working on the tests it turned out that it's as easy as doing copy paste in case the error message is changed. Another excuse was that no real app logic depends on that.

I would prefer to not do that, but only because of being lazy. I can spend some time on fixing it if you have strong objections against keeping it this way.

src/test/java/org/jitsi/jicofo/ParticipantTest.java
paweldomas: It would probably make sense to add constant to SourceGroupPacketExtension as well, but then I have to go and update protocol-jabber... do you think I should do that now ?

bgrozev: I thought we already have it there.

src/test/java/org/jitsi/jicofo/ParticipantTest.java
bgrozev: Hm. Given that a mismatch of the strings will only cause the tests to fail (and not make the succeed while there's a bug), I think that's fine.

bgrozev:
LGTM

src/test/java/org/jitsi/jicofo/ParticipantTest.java
paweldomas: Yes, actually it's there already. Will change the code

src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
```diff
@@ -970,15 +974,25 @@ public void onSessionAccept( JingleSession peerJingleSession,
         // Extract and store various session information in the Participant
         participant.setRTPDescription(answer);
         participant.addTransportFromJingle(answer);
-        participant.addSSRCsFromContent(answer);
-        participant.addSSRCGroupsFromContent(answer);
+
+        try
+        {
+            participant.addSSRCsAndGroupsFromContent(answer);
```
gpolitis: This is an informational comment and not a suggestion to change something; in the bridge we refer to SSRCs + SSRC groups as _RTP Encodings_, i.e. we reconstruct the _RTP encodings_ of a participant based on the SSRCs and the SSRC groups that it has signaled. I think that in the future, RTP encodings will be assigned an RIDs and may optionally have an SSRC. This is the direction that both ORTC and WebRTC seem to be heading towards.

src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
```diff
@@ -935,26 +938,27 @@ public ChatRoomMemberRole getRoleForMucJid(String mucJid)
      * {@inheritDoc}
      */
     @Override
-    public void onSessionAccept( JingleSession peerJingleSession,
+    public XMPPError onSessionAccept( JingleSession peerJingleSession,
```
gpolitis: Highly subjective and highly insignificant, but I think I would have liked better throwing an `XMPPException` here. That would also lead to less changes in the code.

paweldomas: Hmm yes it could be done this way as well

src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
paweldomas: Thanks ! I didn't know about that.

paweldomas: Where can I read some more about that ? Seems to be confusing to me. When I see RTP encodings I think of RTP payloads description.

gpolitis: You can read more about this here http://draft.ortc.org.

PR #168, opened 2017-02-16
Catches all errors in Runnables passed to ScheduledExecutorService.

damencho:
ScheduledExecutorService has the functionality to catch any Error coming out of the Runnable run method and swallow it without printing anything.

### DIFF
--- a/src/main/java/org/jitsi/jicofo/recording/jibri/JibriRecorder.java
+++ b/src/main/java/org/jitsi/jicofo/recording/jibri/JibriRecorder.java
@@ -357,11 +357,13 @@ public class JibriRecorder
                     }
                     //else the response will be handled in processPacket()
                 }
-                catch (OperationFailedException e)
+                catch (Throwable t)
                 {
                     logger.error(
-                            "Failed to start recording in " + roomName
-                                + " - XMPP connection is broken");
+                        "Failed to start recording in " + roomName
+                            + (t instanceof OperationFailedException ?
+                                " - XMPP connection is broken"
+                                : ""), t);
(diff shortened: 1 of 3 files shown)


src/main/java/org/jitsi/jicofo/recording/jibri/JibriRecorder.java
```diff
@@ -357,7 +357,7 @@ public void run()
                     }
                     //else the response will be handled in processPacket()
                 }
-                catch (OperationFailedException e)
+                catch (Throwable e)
```
paweldomas: OperatonFailedException was caught from sendPacketAndGetReply and the only case for that one there is when XMPP is disconnected. So the only thing logged there is a message that the connection is broken. For Throwable we will want to log the whole stack trace I guess.

damencho: Updated it, does it look ok now?

src/main/java/org/jitsi/jicofo/recording/jibri/JibriRecorder.java
```diff
@@ -357,11 +357,11 @@ public void run()
                     }
                     //else the response will be handled in processPacket()
                 }
-                catch (OperationFailedException e)
+                catch (Throwable t)
                 {
                     logger.error(
                             "Failed to start recording in " + roomName
-                                + " - XMPP connection is broken");
+                                + " - XMPP connection is broken", t);
```
paweldomas: kind of, the message should be adjusted as well

PR #172, opened 2017-03-02
Video stream estimation

paweldomas:
Jicofo will try to estimate what is the current video stream count for each JVB based on the amount of channels it allocates/expires.

### DIFF
--- a/src/main/java/org/jitsi/impl/protocol/xmpp/colibri/ColibriConferenceImpl.java
+++ b/src/main/java/org/jitsi/impl/protocol/xmpp/colibri/ColibriConferenceImpl.java
@@ -141,8 +141,7 @@ public class ColibriConferenceImpl
      * @param connection XMPP connection object that wil be used by the new
      *        instance to communicate.
      * @param eventAdmin {@link EventAdmin} instance which will be used to post
-     *        {@link BridgeEvent#VIDEOSTREAMS_ADDED} and
-     *        {@link BridgeEvent#VIDEOSTREAMS_REMOVED}.
+     *        {@link BridgeEvent#VIDEOSTREAMS_CHANGED}.
      */
     public ColibriConferenceImpl(XmppConnection    connection,
                                  EventAdmin        eventAdmin)
@@ -253,7 +252,7 @@ public class ColibriConferenceImpl
 
                 synchronized (stateEstimationSync)
                 {
-                    trackVideoChannelsAdded(newVideoChannelsCount);
+                    trackVideoChannelsAddedRemoved(newVideoChannelsCount);
                 }
 
                 acquireCreateConferenceSemaphore(endpointName);
@@ -325,7 +324,8 @@ public class ColibriConferenceImpl
                     {
                         synchronized (stateEstimationSync)
                         {
-                            trackVideoStreamsRemoved(newVideoChannelsCount);
+                            trackVideoChannelsAddedRemoved(
+                                -newVideoChannelsCount);
                         }
                     }
                 }
@@ -530,7 +530,7 @@ public class ColibriConferenceImpl
                     = ColibriConferenceIQUtil.getChannelCount(
                             channelInfo, "video");
 
-                trackVideoStreamsRemoved(expiredVideoChannels);
+                trackVideoChannelsAddedRemoved(-expiredVideoChannels);
             }
         }
     }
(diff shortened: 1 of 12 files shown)


src/main/java/org/jitsi/impl/protocol/xmpp/colibri/ColibriConferenceImpl.java
```diff
@@ -280,6 +314,30 @@ public ColibriConferenceIQ createColibriChannels(
                         (ColibriConferenceIQ) response, contents);
 
         }
+        catch (Exception e)
+        {
+            try
+            {
+                synchronized (syncRoot)
```
gpolitis: Do you think it would be possible to minimize the synchronization by moving the `synchronized (syncRoot)` inside `checkIfDisposed` and the `synchronized (stateEstimationSync)` in `trackVideoStreamsRemoved`? I don't see an obvious deadlock here, but we are posting a message to the event admin while holding two locks which may be a little risky?

paweldomas: It's using `postEvent` method which just puts that event on a queue (single threaded) without blocking the current thread:
https://github.com/jitsi/jicoco/blob/master/src/main/java/org/jitsi/eventadmin/EventAdmin.java#L33

src/main/java/org/jitsi/jicofo/BridgeSelector.java
```diff
@@ -729,11 +829,70 @@ public int compareTo(BridgeState o)
...
+            return this.getEstimatedVideoStreamCount()
+                - o.getEstimatedVideoStreamCount();
+        }
+
+        private int getEstimatedVideoStreamCount()
+        {
+            int videoStreamEstimate
+                = videoStreamCount != null ? videoStreamCount : 0;
+
+            if (videoStreamCountDiff != null)
```
gpolitis: Is it possible that the `videoStreamCountDiff` is set to `null` (via a to setVideoStreamCount call) from another thread?

paweldomas: yes.. I guess I should store a snapshot in a var, before evaluating. Will that work ? Probably I should revise some other calculations around BridgeState as well

gpolitis: That should do it; Alternatively, you could make the `videoStreamCountDiff` a primitive and sum it unconditionally (null is the same a 0). I don't have any preference tho :)

src/main/java/org/jitsi/jicofo/event/BridgeEvent.java
```diff
@@ -51,11 +51,36 @@
...
+
+    /**
+     * The event is emitted by
+     * {@link org.jitsi.impl.protocol.xmpp.colibri.ColibriConferenceImpl}
+     * when video channels are being expired which results in video stream
+     * count change. It is consumed by {@link BridgeSelector} to estimate
+     * the current video stream count for each JVB.
+     */
+    public static final String VIDEOSTREAMS_REMOVED
+        = "org/jitsi/jicofo/JVB/VIDEOSTREAMS_REMOVED";
```
gpolitis: Not a blocker, just a small suggestion to remove some switches and de-duplicate some code, so please feel to ignore: instead of having two events (VIDEOSTREAMS_REMOVED and VIDEOSTREAMS_ADDED) that carry a positive integer, we can have a single event VIDEOSTREAMS_CHANGED that carry a positive or a negative integer (the diff).

src/main/java/org/jitsi/jicofo/BridgeSelector.java
paweldomas: oh yes stupid me to make it so complex in the first place

src/main/java/org/jitsi/jicofo/event/BridgeEvent.java
paweldomas: yes, good idea !

src/main/java/org/jitsi/jicofo/BridgeSelector.java
gpolitis: I'm pretty sure you had a good reason to use a nullable type :)

src/main/java/org/jitsi/jicofo/BridgeSelector.java
```diff
@@ -593,7 +659,19 @@ synchronized public Version getBridgeVersion(String bridgeJid)
          * If not set we consider it highly occupied,
          * because no stats we have been fetched so far.
          */
-        private int videoStreamCount = Integer.MAX_VALUE;
+        private Integer videoStreamCount = null;
```
gpolitis: Have you considered making this a primitive and equal to 0 initially? (I'm not suggesting that we should make it a primitive, just asking)

src/main/java/org/jitsi/jicofo/event/BridgeEvent.java
```diff
@@ -139,6 +178,17 @@ public String getBridgeJid()
     }
 
     /**
+     * Obtains the value of video stream count associated with the current
+     * <tt>BridgeEvent</tt>.
+     * @return <tt>Integer</tt> or <tt>null</tt> if attribute not present for
```
gpolitis: The property is always set, so we don't have to worry about it, but wouldn't this throw a `ClassCastException`, if the property is not present?

paweldomas: I think it would return null, but I haven't tried that.

gpolitis: okay :)

src/main/java/org/jitsi/jicofo/BridgeSelector.java
paweldomas: No and it doesn't make sense anymore to have nullable type, so I'm changing it to int. Thanks !

PR #176, opened 2017-03-10
Adds VP9 support.

gpolitis:
Other than VP9 support, this also adds a payload type for H264 RTX.


saghul:
Can you split the H264 stuff out? It's not really related and we could land that immediately.

gpolitis:
@saghul we are landing this today !

PR #180, opened 2017-03-21
feat: Adds transport-cc to the offer.

bgrozev:
DO NOT MERGE

This is only to make the transport-cc tests more thorough. It should be merge only after we implement the rest of the transport-cc-based bandwidth estimation.


gpolitis:
Superseded by #216.

PR #181, opened 2017-03-22
feat: Allows updating the COLIBRI channel media direction.

gpolitis:
This SHOULD NOT be merged. If a participant joins video muted, then the m-line direction is set to recvonly and the colibri channel in the bridge gets configured as sendonly and it's not getting updated.

### DIFF
--- a/src/main/java/org/jitsi/jicofo/ChannelAllocator.java
+++ b/src/main/java/org/jitsi/jicofo/ChannelAllocator.java
@@ -258,7 +258,8 @@ public class ChannelAllocator implements Runnable
                     newParticipant.getRtpDescriptionMap(),
                     newParticipant.getSSRCsCopy(),
                     newParticipant.getSSRCGroupsCopy(),
-                    null, null, null);
+                    null, null,
+                    newParticipant.getDirectionMap());
         }
     }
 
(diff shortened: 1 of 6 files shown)


src/main/java/org/jitsi/jicofo/ChannelAllocator.java
```diff
@@ -258,7 +258,7 @@ else if (reInvite)
                     newParticipant.getRtpDescriptionMap(),
                     newParticipant.getSSRCsCopy(),
                     newParticipant.getSSRCGroupsCopy(),
-                    null, null);
+                    null, null, null);
```
paweldomas: why not get media direction map from Participant instance here ?

src/main/java/org/jitsi/jicofo/Participant.java
```diff
@@ -72,6 +73,13 @@
     private Map<String, RtpDescriptionPacketExtension> rtpDescriptionMap;
 
     /**
+     * The map of the most recently received media direction for each Colibri
+     * content.
+     */
+    private Map<String, MediaDirection> directionMap;
+
```
paweldomas: extra new line ?

gpolitis: I did remove the extra line in dc80740, even tho Github doesn't show it here.

gpolitis:
This breaks video unmute when the participant as joined video muted.

PR #186, opened 2017-04-11
Muc call control

damencho:


src/main/java/org/jitsi/jicofo/xmpp/FocusComponent.java
```diff
@@ -466,7 +467,10 @@ else if (smackIq instanceof LogoutIq)
                         String.valueOf(authAuthority.isExternal())));
         }
 
-        if (focusManager.getJitsiMeetServices().getSipGateway() != null)
+        JigasiDetector detector
+            = focusManager.getJitsiMeetServices().getJigasiDetector();
+        if (focusManager.getJitsiMeetServices().getSipGateway() != null
```
paweldomas: That property is advertised only once for the whole conference duration, so it should be true if there is possibility that Jigasi will be discovered ever. Anyway it may turn out some more work is needed for the availability handling on the clients side like it's done for Jibri. Or we can just send error if there are no Jigasis available at the time of request - eventually.

src/main/java/org/jitsi/jicofo/xmpp/BaseBrewery.java
```diff
@@ -0,0 +1,380 @@
...
+import org.jivesoftware.smack.packet.*;
+
+import java.util.*;
+import java.util.concurrent.*;
+
+/**
+ * <tt>BaseBrewery</tt> manages the pool of service instances which
+ * exist in the current session. Does that by joining "brewery" room where
+ * instances connect to and publish their's status in MUC presence.
+ * <tt>AbstractPacketExtension</tt> is the packet extension that will be used
```
paweldomas: What AbstractPacketExtension are you referring to here ? is it T extends PacketExtension ?

src/main/java/org/jitsi/jicofo/xmpp/BaseBrewery.java
```diff
@@ -0,0 +1,380 @@
...
+        Presence presence = member.getPresence();
+
+        if (presence == null)
+            return;
+
+        PacketExtension ext
+            = presence.getExtension(extensionElementName, extensionNamespace);
+
+        // if the extension is required and we do not have it skip processing
+        if (ext == null && requiredExtension)
```
paweldomas: but don't we need that extension for Jigasi as well ? it needs to report it's load anyway

src/main/java/org/jitsi/jicofo/jigasi/JigasiDetector.java
```diff
@@ -0,0 +1,107 @@
...
+     */
+    public JigasiDetector(
+        ProtocolProviderHandler protocolProvider,
+        String breweryName)
+    {
+        super(protocolProvider,
+            breweryName,
+            ColibriStatsExtension.ELEMENT_NAME,
+            ColibriStatsExtension.NAMESPACE,
+            false);
```
paweldomas: Presence extension name and namespace are passed as an arguments, but yet the required flag is set to false. I don't get why it's optional nor it's described anywhere in the docs.

src/main/java/org/jitsi/jicofo/jigasi/JigasiDetector.java
```diff
@@ -0,0 +1,107 @@
...
+package org.jitsi.jicofo.jigasi;
+
+import net.java.sip.communicator.impl.protocol.jabber.extensions.colibri.*;
+import net.java.sip.communicator.util.*;
+import org.jitsi.jicofo.*;
+import org.jitsi.jicofo.xmpp.*;
+import org.jitsi.service.configuration.*;
+
+/**
+ * @author Damian Minkov
```
paweldomas: Some high level description is missing. What does it do what config options activate it ?

src/main/java/org/jitsi/jicofo/xmpp/BaseBrewery.java
damencho: When jigasi joins and there is no load, initial join, the extension is missing. This simplifies logic and for jigasi we can consider it as, no load, empty conferences set.

src/main/java/org/jitsi/jicofo/jigasi/JigasiDetector.java
damencho: Yep, I need to update it. See my reply to your previous comment.

src/main/java/org/jitsi/jicofo/xmpp/BaseBrewery.java
damencho: Yep, it used to be AbstractPacketExtension and at one moment I chang it to PacketExtension and apparently I haven't updated the docs.

src/main/java/org/jitsi/jicofo/jigasi/JigasiDetector.java
```diff
@@ -0,0 +1,107 @@
...
+
+    /**
+     * Loads the name of Jigasi brewery MUC room from the configuration.
+     * @param config the instance of <tt>ConfigurationService</tt> which will be
+     * used to read the properties required.
+     * @return the name of Jigasi brewery or <tt>null</tt> if none configured.
+     */
+    static public String loadBreweryName(ConfigurationService config)
+    {
+        return config.getString(JIGASI_ROOM_PNAME);
```
paweldomas: this duplicates the same method in JibriDetector

damencho: I was in doubt here, whether to remove that method at all, but I was assuming that current logic is to keep those properties(..._ROOM_PNAME) as internal to that class.

src/main/java/org/jitsi/jicofo/xmpp/BaseBrewery.java
```diff
@@ -0,0 +1,380 @@
...
+    /**
+     * Notifies that a brewing instance is going offline.
+     * @param jid the instance muc address
+     */
+    abstract protected void notifyInstanceOffline(String jid);
+
+    /**
+     * Internal structure for storing information about brewing instances.
+     */
+    protected class BrewInstance
```
paweldomas: Not sure if this name make sense here. It's meaning is something around "to brew instance" or "instance of a brew".

paweldomas: or maybe I don't get the Jibri stuff very well... I thought that a Jibri is the one that brews :P

src/main/java/org/jitsi/jicofo/xmpp/BaseBrewery.java
```diff
@@ -0,0 +1,380 @@
...
+    {
+        /**
+         * Eg. "room@muc.server.net/nick"
+         */
+        public final String mucJid;
+
+        /**
+         * One of {@link AbstractPacketExtension}
```
paweldomas: currently it's T extends PacketExtension, so it's not AbstractPacketExtension

src/main/java/org/jitsi/jicofo/jigasi/JigasiDetector.java
```diff
@@ -0,0 +1,107 @@
...
+     * which all Jigasi instances.
+     */
+    private static final String JIGASI_ROOM_PNAME
+        = "org.jitsi.jicofo.jigasi.BREWERY";
+
+    /**
+     * Loads the name of Jigasi brewery MUC room from the configuration.
+     * @param config the instance of <tt>ConfigurationService</tt> which will be
+     * used to read the properties required.
+     * @return the name of Jigasi brewery or <tt>null</tt> if none configured.
```
paweldomas: it is helpful when it's explained whether it's full a MUC address (room123@muc.server.com) or just the name of the room (room123)

src/main/java/org/jitsi/jicofo/jigasi/JigasiDetector.java
```diff
@@ -0,0 +1,107 @@
...
+    static public String loadBreweryName(ConfigurationService config)
+    {
+        return config.getString(JIGASI_ROOM_PNAME);
+    }
+
+    /**
+     * Constructs new JigasiDetector.
+     *
+     * @param protocolProvider the xmpp protocol provider
+     * @param breweryName the room name
```
paweldomas: same here, is it full room address ? can it be either full room address or just the room part of MUC address ?

src/main/java/org/jitsi/jicofo/jigasi/JigasiDetector.java
```diff
@@ -0,0 +1,107 @@
...
+     * no Jigasis available currently.
+     */
+    public String selectJigasi()
+    {
+        BrewInstance lessLoadedInstance = null;
+        int numberOfParticipants = Integer.MAX_VALUE;
+        for (BrewInstance jigasi : instances)
+        {
+            int currentParticipants
+                    = jigasi.status != null ?
```
paweldomas: not sure if Github is showing it weird or if the indentation is messed up here ?

src/main/java/org/jitsi/jicofo/xmpp/BaseBrewery.java
```diff
@@ -0,0 +1,380 @@
...
+    /**
+     * The logger
+     */
+    private static final Logger logger = Logger.getLogger(BaseBrewery.class);
+
+    /**
+     * The name fo XMPP MUC room where all service instances gather to brew
+     * together.
+     */
+    private final String breweryName;
```
paweldomas: same here is it full MUC address ?

src/main/java/org/jitsi/jicofo/xmpp/BaseBrewery.java
```diff
@@ -0,0 +1,380 @@
...
+     * Whether the extension is required in order to add the reporting instance
+     * to the list of operating instances.
+     */
+    private final boolean requiredExtension;
+
+    /**
+     * Creates new instance of <tt>BaseBrewery</tt>
+     * @param protocolProvider the instance fo <tt>ProtocolProviderHandler</tt>
+     * for Jicofo's XMPP connection.
+     * @param breweryName the name of the brewery MUC room where all
```
paweldomas: is it full MUC address ?

src/main/java/org/jitsi/jicofo/xmpp/BaseBrewery.java
```diff
@@ -0,0 +1,380 @@
...
+    /**
+     * Starts the detector by joining it to the brewery room.
+     */
+    private void start()
+    {
+        try
+        {
+            OperationSetMultiUserChat muc
+                = protocolProvider.getOperationSet(
+                OperationSetMultiUserChat.class);
```
paweldomas: indentation ?

src/main/java/org/jitsi/jicofo/xmpp/BaseBrewery.java
damencho: Haha, yep. So I reused that terminology. This used to be just Instance ... but then I changed to BrewInstance to make it more descriptive, a BrewInstance goes into brewery room. We can change naming of course, anyway I was in a doubt for all of the names here :)

src/main/java/org/jitsi/jicofo/xmpp/BaseBrewery.java
paweldomas: why don't we make it add the extension from the start ?

paweldomas: I wouldn't say it simplifies the logic, but the opposite, because you have to deal with the no extension present case

paweldomas: If it's a problem to have it added on join it will appear once it's added (presence will be updated on extension add)

src/main/java/org/jitsi/jicofo/jigasi/JigasiDetector.java
paweldomas: The logic there was to have detector instance null when property is not configured. I wonder if this can be preserved through some method in base class (static factory method or something ?).

src/main/java/org/jitsi/jicofo/xmpp/BaseBrewery.java
damencho: Yep, you are correct.

PR #187, opened 2017-04-12
Moves processing incoming xmpp packets in new thread.

damencho:
Moves incoming xmpp packet processing in new thread to avoid blocking packet reader thread. This causes problems with certain features like jigasi. Sending a message to jigasi and witing for response, while in the meantime jigasi is sending packets can lead for waiting for a timeout till we process other incoming packets.

### DIFF
--- a/src/main/java/org/jitsi/jicofo/MeetExtensionsHandler.java
+++ b/src/main/java/org/jitsi/jicofo/MeetExtensionsHandler.java
@@ -27,6 +27,7 @@ import net.java.sip.communicator.util.Logger;
 
 import org.jitsi.impl.protocol.xmpp.extensions.*;
 import org.jitsi.jicofo.event.*;
+import org.jitsi.jicofo.util.*;
 import org.jitsi.protocol.xmpp.*;
 import org.jitsi.protocol.xmpp.colibri.*;
 import org.jitsi.protocol.xmpp.util.*;
@@ -67,6 +68,11 @@ public class MeetExtensionsHandler
     private OperationSetDirectSmackXmpp smackXmpp;
 
     /**
+     * Process packets in different thread, keeping packets receive order.
+     */
+    private QueuePacketProcessor packetProcessor = null;
+
+    /**
      * Creates new instance of {@link MeetExtensionsHandler}.
      * @param focusManager <tt>FocusManager</tt> that will be used by new
      *                     instance to access active conferences and focus
@@ -98,7 +104,10 @@ public class MeetExtensionsHandler
             = focusManager.getOperationSet(
                     OperationSetDirectSmackXmpp.class);
 
-        smackXmpp.addPacketHandler(this, this);
+        if (this.packetProcessor == null)
+            this.packetProcessor = new QueuePacketProcessor(this);
+
+        smackXmpp.addPacketHandler(this.packetProcessor, this);
     }
 
     /**
@@ -108,7 +117,12 @@ public class MeetExtensionsHandler
     {
         if (smackXmpp != null)
         {
-            smackXmpp.removePacketHandler(this);
+            if (this.packetProcessor != null)
+            {
+                smackXmpp.removePacketHandler(this.packetProcessor);
+                this.packetProcessor.stop();
+                this.packetProcessor = null;
+            }
             smackXmpp = null;
         }
     }
--- a/src/main/java/org/jitsi/jicofo/recording/Recorder.java
+++ b/src/main/java/org/jitsi/jicofo/recording/Recorder.java
@@ -17,6 +17,7 @@
  */
 package org.jitsi.jicofo.recording;
 
+import org.jitsi.jicofo.util.*;
 import org.jitsi.protocol.xmpp.*;
 
 import org.jivesoftware.smack.*;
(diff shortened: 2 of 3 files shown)


src/main/java/org/jitsi/jicofo/MeetExtensionsHandler.java
```diff
@@ -131,26 +131,13 @@ public void processPacket(Packet packet)
...
-        {
-            handleColibriIq((ColibriConferenceIQ) packet);
-        }
-        else if (packet instanceof MuteIq)
-        {
-            handleMuteIq((MuteIq) packet);
-        }
-        else if (packet instanceof RayoIqProvider.DialIq)
-        {
-            handleRayoIQ((RayoIqProvider.DialIq) packet);
```
paweldomas: From all this methods only handleRayoIQ is blocking, but I guess it will not cause any issues to move all of them. But it will only create many thread unnecessary.

src/main/java/org/jitsi/jicofo/recording/Recorder.java
```diff
@@ -92,4 +98,46 @@ public void dispose()
...
+     * Incoming packets processing logic. Process the packet in separate thread
+     * so we do not block readers thread.
+     * <p>
+     * {@inheritDoc}
+     */
+    @Override
+    public void processPacket(Packet packet)
+    {
+        // process packet in separate thread to avoid blocking reader's thread
+        FocusBundleActivator.getSharedThreadPool()
```
paweldomas: Pushing all packets on a thread pool may cause reordering issues. If those are to be pushed without thinking it throughly it should be a single threaded executor.

damencho: Well, we are already in a single thread situation and seeing a problem with that. The other approach is to drop and do not use sendPacketAndGetReply.

src/main/java/org/jitsi/jicofo/MeetExtensionsHandler.java
paweldomas: I would go with utility class instead of duplicating the code:
https://gist.github.com/paweldomas/dff5be0aa4e1024b3d51daeb94d52eda

paweldomas:
With this changes also JibriRecorder needs some cleanup around: https://github.com/jitsi/jicofo/blob/master/src/main/java/org/jitsi/jicofo/recording/jibri/JibriRecorder.java#L336

damencho:
Ups, I missed that utility.

Overall I was thinking, to modify it, as we have detected a problem only with rayo, so we change the threading there and leave the other changes, minimizing risks. WDYT?

paweldomas:
Actually muting is also blocking. I was thinking about this and in general user should not be sending any new packets for the particular subject until the previous one has been processed and response (RESULT/ERROR) was returned. So it should be fine to process them in parallel in this case.

src/main/java/org/jitsi/jicofo/recording/Recorder.java
```diff
@@ -79,6 +79,12 @@ public void dispose()
     public abstract boolean isRecording();
 
     /**
+     * Process incoming packet.
+     * @param packet the packet to process.
+     */
+    public abstract void processIncomingPacket(Packet packet);
```
paweldomas: Adding that in base class requires to review and update docs in al subclasses/check for potential issues. There must be assumption that packets may be processed not in the order they were received.

damencho:
Rework this PR according to some offline discussions, now the processing in non blocking xmpp reader thread, but queues packets and process them in the order they are received.

src/main/java/org/jitsi/jicofo/util/QueuePacketProcessor.java
```diff
@@ -0,0 +1,115 @@
...
+        PacketListener packetListener)
+    {
+        this.packetListener = packetListener;
+
+        // creates and start the executor thread
+        executorThread = new Thread(this,
+            "QueuePacketProcessor@"
+                + this.packetListener.getClass().getCanonicalName());
+        executorThread.setDaemon(true);
+        executorThread.start();
```
paweldomas: wouldn't it be simpler to use Executors.newSingleThreadExecutor() ?

damencho: How about now? :) Thanks for pointing it out.

src/main/java/org/jitsi/jicofo/recording/Recorder.java
```diff
@@ -68,7 +76,12 @@ public void init() { }
      */
     public void dispose()
     {
-        xmpp.removePacketHandler(this);
+        if (this.packetProcessor != null)
+        {
+            this.packetProcessor.stop();
```
paweldomas: I think first you should remove the listener then stop the processor - to avoid submitting new tasks on stopped executor. I don't remember if that would cause any Exceptions.

damencho: Force pushed it.

paweldomas: ok, waiting for the tests...

PR #194, opened 2017-05-03
Deprecate Recorder

paweldomas:
JibriRecorder will no longer extend deprecated Recorder. Also JibriRecorder instance lifetime will no longer be tied to the lifetime of a Colibri conference on the JVB. That's because it needs
to report Jibri availability even if the conference has expired. For example when there's just one participant left and the JVB conference is being expired.


bgrozev:
LGTM

PR #201, opened 2017-05-24
Refactors JitsiMeetConferenceImpl

bgrozev:
Uses a list of bridges and refactors the invitations process to take
this into account.

This is in preparation for support for multi-bridge support (currently
the bridge selection algorithm is unmodified).

### DIFF
--- a/src/main/java/org/jitsi/jicofo/BridgeSelector.java
+++ b/src/main/java/org/jitsi/jicofo/BridgeSelector.java
@@ -39,6 +39,7 @@ import java.util.*;
  * based on feedback from Jitsi Meet conference focus.
  *
  * @author Pawel Domas
+ * @author Boris Grozev
  */
 public class BridgeSelector
     implements SubscriptionListener,
@@ -239,7 +240,9 @@ public class BridgeSelector
         }
 
         if (bridge != null)
+        {
             notifyBridgeDown(bridge);
+        }
     }
 
     /**
@@ -251,12 +254,30 @@ public class BridgeSelector
      * @param participant the participant for which a bridge is to be selected.
      */
     synchronized public BridgeState selectVideobridge(
-            JitsiMeetConference conference, Participant participant)
+            JitsiMeetConferenceImpl conference, Participant participant)
     {
         List<BridgeState> bridges = getPrioritizedBridgesList();
         if (bridges.size() == 0)
         {
             return null;
+        }
+
+        // TODO: this is temporary hack, which implements the "single bridge"
+        // bridge selection strategy. It will be replaced with a proper
+        // implementation later.
+        if (conference != null)
+        {
+            String jid = conference.getJvbJid();
+            if (jid != null)
+            {
+                BridgeState bridgeState = getBridgeState(jid);
+                if (bridgeState == null)
+                {
+                    logger.error(
+                        "Could not find a conference's existing bridge...");
+                }
+                return bridgeState;
+            }
         }
 
         BridgeState bridge = bridges.get(0);
@@ -271,7 +292,7 @@ public class BridgeSelector
      * @return the selected bridge, represented by its {@link BridgeState}.
      */
     public BridgeState selectVideobridge(
-            JitsiMeetConference conference)
+            JitsiMeetConferenceImpl conference)
     {
         return selectVideobridge(conference, null);
     }
@@ -295,7 +316,9 @@ public class BridgeSelector
         {
             BridgeState bridge = bridgesIter.next();
             if (!bridge.isOperational())
+            {
                 bridgesIter.remove();
+            }
         }
 
         return bridgeList;
@@ -659,6 +682,19 @@ public class BridgeSelector
         {
             handlerRegistration.unregister();
             handlerRegistration = null;
+        }
+    }
+
+    /**
+     * @return the {@link BridgeState} for the bridge with a particular XMPP
+     * JID.
+     * @param jid the JID of the bridge.
+     */
+    public BridgeState getBridgeState(String jid)
+    {
+        synchronized (bridges)
+        {
+            return bridges.get(jid);
         }
     }
 }
--- a/src/main/java/org/jitsi/jicofo/BridgeSelector.java
+++ b/src/main/java/org/jitsi/jicofo/BridgeSelector.java
@@ -459,6 +459,14 @@ public class BridgeSelector
                     bridgeState.setVideoStreamCount(videoStreamCount);
                 }
             }
+            else if ("relay_id".equals(stat.getName()))
+            {
+                Object relayId = stat.getValue();
+                if (relayId != null)
+                {
+                    bridgeState.setRelayId(relayId.toString());
+                }
+            }
         }
     }
 
--- a/src/main/java/org/jitsi/jicofo/BridgeSelector.java
+++ b/src/main/java/org/jitsi/jicofo/BridgeSelector.java
@@ -160,30 +160,37 @@ public class BridgeSelector
 
     /**
      * Adds next Jitsi Videobridge XMPP address to be observed by this selected
-     * and taken into account in best bridge selection process.
+     * and taken into account in best bridge selection process. If a bridge
+     * with the given JID already exists, it is returned and a new instance is
+     * not created.
      *
      * @param bridgeJid the JID of videobridge to be added to this selector's
-     *                  set of videobridges.
+     * set of videobridges.
+     * @return the {@link BridgeState} for the bridge with the provided JID.
      */
-    public void addJvbAddress(String bridgeJid)
+    public BridgeState addJvbAddress(String bridgeJid)
     {
-        addJvbAddress(bridgeJid, null);
+        return addJvbAddress(bridgeJid, null);
     }
 
     /**
      * Adds next Jitsi Videobridge XMPP address to be observed by this selected
-     * and taken into account in best bridge selection process.
+     * and taken into account in best bridge selection process. If a bridge
+     * with the given JID already exists, it is returned and a new instance is
+     * not created.
      *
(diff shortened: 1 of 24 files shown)


src/main/java/org/jitsi/jicofo/ChannelAllocator.java
```diff
@@ -95,30 +102,36 @@
...
 
     /**
-     * Creates new instance of <tt>ChannelAllocator</tt> which is meant to
-     * invite given <tt>Participant</tt> into the given
-     * <tt>JitsiMeetConference</tt>.
+     * A flag which indicates whether channel allocation is cancelled. Raising
+     * this makes the allocation thread discontinue the allocation process and
+     * return.
+     */
+    private boolean canceled = false;
```
paweldomas: I'm not an expert here, but shouldn't it be 'cancelled' ?

src/main/java/org/jitsi/jicofo/BridgeSelector.java
```diff
@@ -251,14 +254,32 @@ synchronized public void removeJvbAddress(String bridgeJid)
      * @param participant the participant for which a bridge is to be selected.
      */
     synchronized public BridgeState selectVideobridge(
-            JitsiMeetConference conference, Participant participant)
+            JitsiMeetConferenceImpl conference, Participant participant)
```
bgrozev: Yes. I was planning to remove JitsiConferenceImpl#getJvbJid() (which is the reason Impl is needed here) and JitsiConferenceImpl#getJvbConferenceId(). When I do, I will also remove the usage of Impl.

src/main/java/org/jitsi/jicofo/ChannelAllocator.java
bgrozev: I think both are correct spellings (with the single "l" version being preferred in American English). I see in jicofo we use double "l", but in the other projects we use both spellings.

paweldomas: you even used "cancelled" in the docs :P

bgrozev: Oops. I'll at least make it consistent :)

paweldomas: Do you think it makes sense to make cancel volatile, since one thread is writing and another one is reading ?

src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
```diff
@@ -772,41 +884,37 @@ public void destroy(String reason)
...
-                else
-                {
-                    colibriConference.dispose();
-                }
-                colibriConference = null;
+                // No need to expire channels, just expire the whole colibri
+                // conference.
+                // bridgeDesc.terminateAll();
+                bridgeDesc.dispose();
+                bridges.remove(bridgeDesc);
```
paweldomas: will it trigger ConcurrentModificationException ?

src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
```diff
@@ -772,41 +884,37 @@ public void destroy(String reason)
...
             }
         }
+
+        // TODO: what about removing the participants and ending their jingle
```
paweldomas: Clients are "programmed" to leave when Jicofo leaves. Their channels are expired (or at least used to be in colibriConference.expireConference())

src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
```diff
@@ -1740,19 +1845,28 @@ void onChannelAllocationFailed(
...
-            // Dispose the conference. This way we'll know there is no
-            // conference active and we can restart on new bridge
-            disposeConference();
+        BridgeDesc bridgeDesc = channelAllocator.getBridgeDesc();
+        Participant participant = channelAllocator.getParticipant();
+        bridgeDesc.terminate(participant);
+
+        // Retry once.
+        // TODO: be smarter about re-trying: e.g. don't retry if we know there
+        // was a bridge bridge
```
paweldomas: "there was a bridge bridge" typo  ?

src/main/java/org/jitsi/jicofo/ChannelAllocator.java
```diff
@@ -301,20 +326,30 @@ else if (reInvite)
...
+                // responsibility to retry the invitation if necessary.
+                return null;
+            }
+            throw new OperationFailedException(
+                "Colibri channel allocation failed",
+                CHANNEL_ALLOCATION_FAILED_ERR_CODE);
+        }
 
-        newParticipant.setColibriChannelsInfo(peerChannels);
+        participant.setColibriChannelsInfo(colibriChannels);
```
paweldomas: there's no check for canceled, before calling 'setColibriChannelsInfo' ?

src/main/java/org/jitsi/jicofo/Participant.java
```diff
@@ -748,4 +759,27 @@ public String getMucJid()
...
+     */
+    public void setChannelAllocator(ChannelAllocator channelAllocator)
+    {
+        synchronized (channelAllocatorSyncRoot)
+        {
+            if (this.channelAllocator != null)
+            {
+                // There is an ongoing thread allocating channels and sending
+                // an invite for this participant. Tell it to stop.
+                this.channelAllocator.cancel();
```
paweldomas: We need to make sure that ChannelAllocator once canceled will not modify Participant, because it can overwrite what was written by the new ChannelAllocator (see my comment above about "why there is no check for canceled").

src/main/java/org/jitsi/jicofo/BridgeSelector.java
```diff
@@ -141,6 +141,12 @@ private static Integer getInt(Object obj)
     private final Map<String, String> pubSubToBridge = new HashMap<>();
 
     /**
+     * The bridge selection strategy.
+     */
+    private BridgeSelectionStrategy bridgeSelectionStrategy
```
paweldomas: ❤️ ❤️ ❤️

src/main/java/org/jitsi/jicofo/ChannelAllocator.java
```diff
@@ -414,7 +414,7 @@ private ColibriConferenceIQ allocateChannels(
                     return null;
                 }
 
-                bridgeSelector.updateBridgeOperationalStatus(jvb, true);
+                bridgeDesc.bridgeState.setIsOperational(true);
```
paweldomas: shouldn't all variables written in bridgeState.setIsOperational be volatile ?

paweldomas:
Other than the comments above LGTM. I wonder if it would make sense to merge BridgeDesc with ColibriConference. I have a feeling that it could make things simpler, but I don't know what else you will be doing there.

Also BridgeDesc name seems to be confusing to me with BridgeState. It feels more like a kind of bridge session. But we can change it later if we find something more suitable.

src/main/java/org/jitsi/jicofo/ChannelAllocator.java
bgrozev: Yes, I will add it.

src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
bgrozev: Yes, the colibri channels are still expired. But we are still left with Participant instances and their jingle sessions. Normally I would expect dispose() to take care of these, but it's not a prio so I only added a comment for now.

paweldomas: Jingle session are discarded in jingle.terminateHandlersSessions(this);

paweldomas: Nope, forget about it. It's a different thing

src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
bgrozev: Yes! Fixed

src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
bgrozev: fixed

src/main/java/org/jitsi/jicofo/ChannelAllocator.java
bgrozev: added

src/main/java/org/jitsi/jicofo/Participant.java
bgrozev: I added some more checks

src/main/java/org/jitsi/jicofo/ChannelAllocator.java
bgrozev: Made volatile

bgrozev:
I really like BridgeSession! I renamed it. As for merging with ColibriConferenceImpl, I don't know -- CCI lives in protocol.xmpp, but BridgeSession has a list of Participant.

paweldomas:
LGTM, feel free to merge once the tests pass. Btw. have you tested conference migration scenario ? You can do that by having two bridges on your instance, starting a new conference and stopping the bridge that was selected.

bgrozev:
I *think* I tested this, but I'm not sure. I will test it and then merge unless there are problems.

bgrozev:
I tested conference migration, and after the last fix it works. I'll merge after the jenkins tests.

PR #204, opened 2017-07-05
advertise rtp stream id header ext

bbaldino:


bbaldino:
hold off on this, need to update the libjitsi version

Issue #203, opened 2017-07-01
Focus fails to connect when other SASL Mechs are offered by Openfire


speedy01:
I'm using GSSAPI and PLAIN.  focus fails to connect with invalid username/password.  If GSSAPI is disabled, the focus user connects as expected.  It appears that focus is trying to connected using GSSAPI.  
To work around this, I have added SASLAuthentication.supportSASLMechanism("PLAIN");  to the connConfig within src/main/java/org/jitsi/impl/protocol/xmpp/XmppProtocolProvider.java 

I'm unsure if this is the best way to work around this issue.  It maybe better to use blacklist/unsupported, but I'll leave that up to the experts.

PR #205, opened 2017-07-07
Update XmppProtocolProvider.java

speedy01:
The focus user connects with SASL Mechanisms ANONYMOUS or PLAIN, but might try to authenticate  with a non supported mechanism if one is offered by the server.  This moves the preference of PLAIN to the top, and disables the unused mechanisms.  This should resolve issue #203
https://github.com/jitsi/jicofo/issues/203

### DIFF
--- a/src/main/java/org/jitsi/impl/protocol/xmpp/XmppProtocolProvider.java
+++ b/src/main/java/org/jitsi/impl/protocol/xmpp/XmppProtocolProvider.java
@@ -160,8 +160,8 @@ public class XmppProtocolProvider
         ConnectionConfiguration connConfig
             = new ConnectionConfiguration(
                     serverAddressUserSetting, serverPort, serviceName);
-                    // focus uses SASL Mechanisms ANONYMOUS and PLAIN, but trys to authenticate
-                    // with GSSAPI when it's offered by the server. Disable GSSAPI.
+        // focus uses SASL Mechanisms ANONYMOUS and PLAIN, but trys to authenticate
+        // with GSSAPI when it's offered by the server. Disable GSSAPI.
                     SASLAuthentication.unsupportSASLMechanism("GSSAPI");
                  
         connection = new XMPPConnection(connConfig);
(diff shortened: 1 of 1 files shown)


speedy01:
I have signed the CLA

speedy01:
after receiving feedback, it was suggested to only disable GSSAPI, and not all sasl mechs. I have updated the PR

src/main/java/org/jitsi/impl/protocol/xmpp/XmppProtocolProvider.java
```diff
@@ -160,7 +160,13 @@ public synchronized void register(SecurityAuthority securityAuthority)
         ConnectionConfiguration connConfig
             = new ConnectionConfiguration(
                     serverAddressUserSetting, serverPort, serviceName);
-
+                    
+                    // focus uses SASL Mechanisms ANONYMOUS and PLAIN, but trys to authenticate
```
linkmauve: Indentation is incorrect here.

speedy01:
@linkmauve I believe I have fixed the issue..

linkmauve:
I think you have not, you should align to the beginning of the scope, not to the previous line.  Alternatively check your editor’s configuration for using tabs instead of spaces, or the opposite.

speedy01:
@linkmauve sorry..I'm still learning.  Hopefully I got it correct this time.

src/main/java/org/jitsi/impl/protocol/xmpp/XmppProtocolProvider.java
```diff
@@ -160,7 +160,10 @@ public synchronized void register(SecurityAuthority securityAuthority)
         ConnectionConfiguration connConfig
             = new ConnectionConfiguration(
                     serverAddressUserSetting, serverPort, serviceName);
-
+        // focus uses SASL Mechanisms ANONYMOUS and PLAIN, but trys to authenticate
+        // with GSSAPI when it's offered by the server. Disable GSSAPI.
+                    SASLAuthentication.unsupportSASLMechanism("GSSAPI");
```
linkmauve: Almost. :)

This line should also be aligned like the previous two.

speedy01:
@linkmauve ..updated

linkmauve:
Perfect. :)

Now you can wait for someone from this project to review your patch, and merge it if they find it correct.

damencho:
Do you see the problem with latest jicofo, we updated it to use smack4?

speedy01:
@damencho 
yes..I'm still seeing this issue with the jicofo and smack4.

PR #212, opened 2017-07-26
Changes in preparation for relay support

bgrozev:


src/main/java/org/jitsi/impl/protocol/xmpp/colibri/ColibriConferenceImpl.java
```diff
@@ -377,50 +381,60 @@ public ColibriConferenceIQ createColibriChannels(
...
-                = OperationFailedException.NETWORK_FAILURE;
-            allocChannelsErrorMsg
-                = "Failed to allocate colibri channels: response is null."
-                + " Maybe the response timed out.";
-        }
-        else if (response.getError() != null)
+        // This code block must be protected, because in the last "if" a
+        // decision is made based on the value of allocChannelsErrorCode, which
+        // could have been written by another thread. We use syncRoot, because
+        // the existing code already runs with its lock acquired.
```
paweldomas: I don't understand this part of the comment:
> We use syncRoot, because the existing code already runs with its lock acquired.
It is being acquired here, isn't it ? It was released in top level, before calling this method.

paweldomas:
LGTM, only that one comment is not very clear to me

src/main/java/org/jitsi/impl/protocol/xmpp/colibri/ColibriConferenceImpl.java
bgrozev: You are right, I don't know why I wrote this comment. Removed the sentence that you quoted.

PR #213, opened 2017-07-28
fix: Does not hold the lock on conferencesSyncRoot

bgrozev:
while creating a conference. Creating the conference blocks until the
MUC is joined, and it has been observed to block for long periods of
time (the 15 seconds timeout used by smack), during which new
conferences can not be created (and thus health checks fail).

### DIFF
--- a/src/main/java/org/jitsi/jicofo/FocusManager.java
+++ b/src/main/java/org/jitsi/jicofo/FocusManager.java
@@ -322,17 +322,20 @@ public class FocusManager
         JitsiMeetConferenceImpl conference;
         synchronized (conferencesSyncRoot)
         {
-            if (!conferences.containsKey(room))
-            {
-                if (shutdownInProgress)
-                {
-                    return false;
-                }
+            conference = conferences.get(room);
+        }
 
-                createConference(room, properties, loggingLevel);
+        if (conference == null)
+        {
+            // A new conference request.
+            if (shutdownInProgress)
+            {
+                return false;
             }
 
-            conference = conferences.get(room);
+            // createConference blocks waiting for the MUC to be joined,
+            // and it should not hold any locks while doing this.
+            conference = createConference(room, properties, loggingLevel);
         }
 
         return conference.isInTheRoom();
@@ -350,7 +353,7 @@ public class FocusManager
      *
      * @throws Exception if any error occurs.
      */
-    private void createConference(
+    private JitsiMeetConferenceImpl createConference(
             String room, Map<String, String> properties, Level logLevel)
         throws Exception
     {
@@ -360,14 +363,18 @@ public class FocusManager
             = JitsiMeetGlobalConfig.getGlobalConfig(
                 FocusBundleActivator.bundleContext);
 
-        String id = generateConferenceId();
-        JitsiMeetConferenceImpl conference
-            = new JitsiMeetConferenceImpl(
+        JitsiMeetConferenceImpl conference;
+        String id;
+        synchronized (conferencesSyncRoot)
+        {
+            id = generateConferenceId();
+            conference
+                = new JitsiMeetConferenceImpl(
                     room, focusUserName, protocolProviderHandler,
                     this, config, globalConfig, logLevel, id);
 
-        conferences.put(room, conference);
-        conferenceIds.add(id);
+            conferences.put(room, conference);
+        }
 
         StringBuilder options = new StringBuilder();
         for (Map.Entry<String, String> option : properties.entrySet())
@@ -411,9 +418,16 @@ public class FocusManager
             // on FocusManager instance.
 
             //conference.stop();
+            synchronized (conferencesSyncRoot)
+            {
+                conferences.remove(room);
+                conferenceIds.remove(id);
+            }
 
             throw e;
         }
+
+        return conference;
     }
 
     /**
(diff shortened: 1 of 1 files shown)


src/main/java/org/jitsi/jicofo/FocusManager.java
```diff
@@ -320,19 +320,45 @@ public boolean conferenceRequest(
...
 
         JitsiMeetConferenceImpl conference;
+        String id;
         synchronized (conferencesSyncRoot)
         {
-            if (!conferences.containsKey(room))
+            conference = conferences.get(room);
+            if (conference == null)
+            {
+                id = generateConferenceId();
```
bbaldino: is there a reason to do these things here and not in the same block where we call `createConference`? then all the 'create new conference'-type logic would be in the same place

bbaldino: it seems like it would be even better for `createConference` to do all this work, actually, but we don't need to do that in this PR, perhaps note as a FIXME?

src/main/java/org/jitsi/jicofo/FocusManager.java
```diff
@@ -360,14 +387,15 @@ private void createConference(
             = JitsiMeetGlobalConfig.getGlobalConfig(
                 FocusBundleActivator.bundleContext);
 
-        String id = generateConferenceId();
```
bbaldino: oh, why did this have to be moved to outside the method?

src/main/java/org/jitsi/jicofo/FocusManager.java
```diff
@@ -360,14 +387,15 @@ private void createConference(
...
         JitsiMeetConferenceImpl conference
             = new JitsiMeetConferenceImpl(
```
bbaldino: where is the muc join actually happening...is it part of this?

src/main/java/org/jitsi/jicofo/FocusManager.java
bgrozev: generateConferenceId() needs the lock, but we specifically want createConference to execute without holding the lock.

bbaldino: but `createConference` still acquires the lock inside for a small block?

src/main/java/org/jitsi/jicofo/FocusManager.java
bgrozev: No, it's in conference.start() below

src/main/java/org/jitsi/jicofo/FocusManager.java
bgrozev: Hm. Now that I look at it again I don't see any reason.

src/main/java/org/jitsi/jicofo/FocusManager.java
bgrozev: Yes, that's fine. We want to avoid the lockbeing held while conference.start() executes.

bgrozev:
I forced pushed a new version which leaves generateId() where it used to be.

src/main/java/org/jitsi/jicofo/FocusManager.java
```diff
@@ -360,14 +363,18 @@ private void createConference(
...
                 FocusBundleActivator.bundleContext);
 
-        String id = generateConferenceId();
-        JitsiMeetConferenceImpl conference
-            = new JitsiMeetConferenceImpl(
+        JitsiMeetConferenceImpl conference;
+        String id;
+        synchronized (conferencesSyncRoot)
+        {
+            id = generateConferenceId();
```
bbaldino: this can technically be outside, no? since it grabs it on its own?  probably doesn't make a difference i guess.

bgrozev: I'd rather obtain the lock once than multiple times

bbaldino: what i meant was that `generateConferenceId` already grabs the lock inside, so calling `id = generateConferenceId()` outside this synchronized block will still generate the conference id safely.

bbaldino: it could be done as `String id = generateConferenceId();` above the synchronized instead of splitting across the two lines, i mean.

bgrozev: Yeah, I agree it could be done this way, and I can change it if you think that's a better way. I prefer the current code, because it obtains the lock once (when the second synchronized block is reached the thread already has the lock so it doesn't block) as opposed to obtaining it, then releasing it and then obtaining again. I don't know if that line of reasoning is correct.

src/main/java/org/jitsi/jicofo/FocusManager.java
```diff
@@ -322,17 +322,20 @@ public boolean conferenceRequest(
...
+            // A new conference request.
+            if (shutdownInProgress)
+            {
+                return false;
             }
 
-            conference = conferences.get(room);
+            // createConference blocks waiting for the MUC to be joined,
+            // and it should not hold any locks while doing this.
+            conference = createConference(room, properties, loggingLevel);
```
paweldomas: isn't it now possible 2 threads to enter createConference ?

paweldomas:
If Jicofo is stuck there for 15 seconds there is clearly another problem and Jicofo is not healthy. It runs on locahost connection, right ?

src/main/java/org/jitsi/jicofo/FocusManager.java
bgrozev: Yes!

PR #216, opened 2017-08-28
feat: Implements send-side bandwidth estimations.

gpolitis:
Provides https://github.com/jitsi/libjitsi/pull/361
Depends https://github.com/jitsi/jitsi/pull/391/files

### DIFF
--- a/src/main/java/org/jitsi/jicofo/util/JingleOfferFactory.java
+++ b/src/main/java/org/jitsi/jicofo/util/JingleOfferFactory.java
@@ -89,6 +89,18 @@ public class JingleOfferFactory
     public static final String ENABLE_TCC_PNAME = "org.jitsi.jicofo.ENABLE_TCC";
 
     /**
+     * The name of the property which enables the inclusion of the AST RTP
+     * header extension in the offer.
+     */
+    public static final String ENABLE_AST_PNAME = "org.jitsi.jicofo.ENABLE_AST";
+
+    /**
+     * The name of the property which enables the inclusion of the TOF RTP
+     * header extension in the offer.
+     */
+    public static final String ENABLE_TOF_PNAME = "org.jitsi.jicofo.ENABLE_TOF";
+
+    /**
      * The VP8 payload type to include in the Jingle session-invite.
      */
     private final int VP8_PT;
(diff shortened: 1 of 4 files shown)


gpolitis:
Provides https://github.com/jitsi/libjitsi/pull/361

gpolitis:
Thanks for approving @bbaldino.

gpolitis:
duh, I pressed "Close and comment"...

src/main/java/org/jitsi/jicofo/util/JingleOfferFactory.java
```diff
@@ -322,6 +344,13 @@ private void addVideoToContent(ContentPacketExtension content,
         // a=rtcp-fb:100 goog-remb
         vp8.addRtcpFeedbackType(createRtcpFbPacketExtension("goog-remb", null));
 
+        if (ENABLE_TCC)
+        {
+            // a=rtcp-fb:100 transport-cc
+            vp8.addRtcpFeedbackType(
```
bbaldino: i know we have these all organized as per-codec, but for things like this i wonder if it would be best to have a single `if (ENABLE_TCC)` block which adds it for all codecs (perhaps the codecs should even be added to an array or something so it could just iterate over it rather than having to hard-code each one which was created above?)

src/main/java/org/jitsi/jicofo/util/JingleOfferFactory.java
```diff
@@ -539,6 +582,13 @@ private static void addAudioToContent(ContentPacketExtension content,
         // fmtp:111 useinbandfec=1
         addParameterExtension(opus, "useinbandfec", "1");
 
+        if (ENABLE_TCC)
```
bbaldino: could we just pass this in (like `stereo`) and keep this method static?

gpolitis: ENABLE_TCC looks like a static final but is a final instance field. I can rename the field to lower-case (enableTcc), keep the method static with an extra param, and overload it with an instance method that will pass the enableTcc field. Would that work?

src/main/java/org/jitsi/jicofo/util/JingleOfferFactory.java
gpolitis: Will do, make more sense.

src/main/java/org/jitsi/jicofo/util/JingleOfferFactory.java
bbaldino: sounds good!

PR #221, opened 2017-09-04
Don't start an unused SOCKS5 proxy.

guusdk:
This commit prevents the default SOCKS5 proxy (from Smack) from being started.


guusdk:
Isn't jitsi-jenkins supposed to keep track of people that already tracked this? It continues to ask me.

damencho:
Nope, it is per project.

damencho:
Jenkins add to whitelist

guusdk:
I must be contributing to to many projects then :)

saghul:
> I must be contributing to to many projects then :)

And we salute you for that! 🖖

PR #223, opened 2017-09-15
Update README.md

neilzone:
Added a note to Secure Domain configuration on how to create a Prosody user.

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,6 +97,11 @@ from anonymous domain. Here's what has to be configured:
  VirtualHost guest.jitsi-meet.example.com
      authentication = "anonymous"
  ```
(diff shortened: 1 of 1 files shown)


saghul:
Looks good, thanks @neilzone! Have you signed the CLA mentioned above? Cheers!

neilzone:
No - when I try to do so gives an error:
_Unable to submit form data, Error: An error has occurred._

(And that's leaving aside the point about requiring a CLA for such a small patch!)

saghul:
@neilzone I'll pass the message along.

kpfleming:
I'm just now following the Secure Domain setup instructions: does this command *replace* the `register` command? It seems to duplicate the information.

saghul:
@kpfleming Oops, you're right.

PR #226, opened 2017-09-19
feat(SSRCValidator): check for MSID conflicts

paweldomas:
SSRCValidator will now check for MSID conflicts across SIM and FID groups.

### DIFF
--- a/src/main/java/org/jitsi/jicofo/SSRCValidator.java
+++ b/src/main/java/org/jitsi/jicofo/SSRCValidator.java
@@ -158,6 +158,18 @@ public class SSRCValidator
         for (SimulcastGrouping simGrouping : simGroupings)
         {
             String simulcastMsid = simGrouping.getSimulcastMsid();
+
+            if (simGrouping.isUsingRidSignaling())
+            {
+                // Skip RID simulcast group
+                continue;
+            }
+            else if (StringUtils.isNullOrEmpty(simulcastMsid))
+            {
+                throw new InvalidSSRCsException(
+                        "No MSID in simulcast group: " + simGrouping);
+            }
+
             List<SourcePacketExtension> sourcesWithTheMsid
                 = groupedSources.findSourcesWithMsid(
                         mediaType, simulcastMsid);
--- a/src/main/java/org/jitsi/jicofo/SSRCValidator.java
+++ b/src/main/java/org/jitsi/jicofo/SSRCValidator.java
@@ -71,6 +71,111 @@ public class SSRCValidator
     private final int maxSourceCount;
 
     /**
+     * Filters out FID groups that do belong to any simulcast grouping.
+     *
+     * @param simGroupings the list of all {@link SimulcastGrouping}s.
+     * @param fidGroups the list of all FID groups even these which are part of
+     *        the <tt>simGroupings</tt>.
+     *
+     * @return a list of FID groups that are not part of any SIM grouping.
+     */
+    static private List<SourceGroup> getIndependentFidGroups(
+            List<SimulcastGrouping>    simGroupings,
+            List<SourceGroup>          fidGroups)
+    {
+        if (simGroupings.isEmpty())
+        {
+            // Nothing to be done here...
+            return new ArrayList<>(fidGroups);
+        }
+
+        List<SourceGroup> independentFidGroups = new LinkedList<>();
+
+        for (SourceGroup fidGroup : fidGroups)
+        {
+            for (SimulcastGrouping simGrouping : simGroupings)
+            {
+                if (!simGrouping.belongsToSimulcastGrouping(fidGroup))
+                {
+                    independentFidGroups.add(fidGroup);
+                }
+            }
+        }
+
+        return independentFidGroups;
+    }
+
+    /**
+     * Checks if there are no MSID conflicts across all independent FID groups.
+     *
+     * @param independentFidGroups the list of all independent FID groups (that
+     *        do not belong to any other higher level grouping).
+     *
+     * @throws InvalidSSRCsException in case of MSID conflict
+     */
+    static private void verifyNoMsidConflictsAcrossFidGroups(
+            List<SourceGroup> independentFidGroups)
+        throws InvalidSSRCsException
+    {
+        for (SourceGroup fidGroup : independentFidGroups)
+        {
+            // NOTE at this point we're sure that every source has MSID
+            String fidGroupMsid = fidGroup.getGroupMsid();
+            List<SourceGroup> withTheMsid
+                = SSRCSignaling.selectWithMsid(
+                        independentFidGroups, fidGroupMsid);
+
+            for (SourceGroup conflictingGroup : withTheMsid)
+            {
+                if (conflictingGroup != fidGroup)
+                {
+                    throw new InvalidSSRCsException(
+                            "MSID conflict across FID groups: "
+                                + fidGroupMsid + ", " + conflictingGroup
+                                + " conflicts with group " + fidGroup);
+                }
+            }
+        }
+    }
+
+    /**
+     * Checks if there are no MSID conflicts across simulcast groups.
+     *
+     * @param mediaType the media type to be checked in this call
+     * @param groupedSources the map holding all sources which belong to any
+     *        group for all media types.
+     * @param simGroupings the list of all {@link SimulcastGrouping}s for
+     *        the <tt>mediaType</tt>.
+     *
+     * @throws InvalidSSRCsException in case of MSID conflict
+     */
+    static private void verifyNoMsidConflictsAcrossSimGroupings(
+            String                     mediaType,
+            MediaSourceMap             groupedSources,
+            List<SimulcastGrouping>    simGroupings)
+        throws InvalidSSRCsException
+    {
+        for (SimulcastGrouping simGrouping : simGroupings)
+        {
+            String simulcastMsid = simGrouping.getSimulcastMsid();
+            List<SourcePacketExtension> sourcesWithTheMsid
+                = groupedSources.findSourcesWithMsid(
+                        mediaType, simulcastMsid);
+
+            for (SourcePacketExtension src : sourcesWithTheMsid)
+            {
+                if (!simGrouping.belongsToSimulcastGrouping(src))
+                {
+                    throw new InvalidSSRCsException(
+                        "MSID conflict across SIM groups: "
+                            + simulcastMsid + ", " + src
+                            + " conflicts with group " + simGrouping);
+                }
+            }
+        }
+    }
+
+    /**
      * Creates new <tt>SSRCValidator</tt>
      * @param endpointId participant's endpoint ID
      * @param sources participant's source map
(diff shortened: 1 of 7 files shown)


gpolitis:
@paweldomas is this ready to be merged? @bbaldino any observations and/or suggestions?

src/main/java/org/jitsi/protocol/xmpp/util/SourceGroup.java
```diff
@@ -103,6 +117,61 @@ public void addSources(List<SourcePacketExtension> video)
...
     /**
+     * Checks if given {@link SourcePacketExtension} is part of this group.
+     *
+     * @param source {@link SourcePacketExtension} to be checked.
+     *
+     * @return <tt>true</tt> or <tt>false</tt>
+     */
+    public boolean belongsToGroup(SourcePacketExtension source)
+    {
+        if (source.hasSSRC())
```
bbaldino: there is `SourcePacketExtension#sourceEquals` that can be used here:

```
for (SourcePacketExtension groupSrc : this.getSources())
{
    if (groupSrc.sourceEquals(source))
    {
        return true;
    }
}
```

src/main/java/org/jitsi/protocol/xmpp/util/SourceGroup.java
```diff
@@ -81,6 +82,19 @@ public SourceGroup(SourceGroupPacketExtension group)
     }
 
     /**
+     * Creates new instance of <tt>SourceGroup</tt>.
+     * @param semantics the group's semantics
+     * @param sources a {@link List} of the group's
+     *        {@link SourcePacketExtension}s
+     */
+    public SourceGroup(String semantics, List<SourcePacketExtension> sources)
```
bbaldino: this seems to overlap a lot with the existing constructor (this takes in the fields of a `SourceGroupPacketExtension` whereas the other one just takes in a `SourceGroupPacketExtension` directly, is there a need for both?

src/main/java/org/jitsi/protocol/xmpp/util/SimulcastGrouping.java
```diff
@@ -0,0 +1,199 @@
...
+
+import net.java.sip.communicator.impl.protocol.jabber.extensions.colibri.*;
+import net.java.sip.communicator.impl.protocol.jabber.extensions.jingle.*;
+
+import java.util.*;
+
+/**
+ * Utility class for the simulcast {@link SourcePacketExtension} grouping.
+ */
+public class SimulcastGrouping
```
bbaldino: why not have this extend `SourceGroupPacketExtension` ? Or `SourceGroup`?

src/main/java/org/jitsi/protocol/xmpp/util/SimulcastGrouping.java
```diff
@@ -0,0 +1,199 @@
...
+     * @param mediaType the type of the media of the SIM group.
+     * @param groupsMap the map of all groups of all media types.
+     * @param simGroup the root SIM group for which {@link SimulcastGrouping}
+     *        will be extracted.
+     *
+     * @return {@link SimulcastGrouping} which contains SIM and all FID
+     * subgroups.
+     */
+    public static SimulcastGrouping extractSimGrouping(
+            String                 mediaType,
```
bbaldino: is `mediaType` necessary?  i'd think a non-video simulcast group would be an error.

bbaldino: and we already pass in `simGroup` which has the simulcast group?  maybe i'm just missing what this class is trying to accomplish.  is it just to include all the FID groups which relate to the simulcast-grouped ssrcs?

src/main/java/org/jitsi/protocol/xmpp/util/SimulcastGrouping.java
```diff
@@ -0,0 +1,199 @@
...
+     * Checks if given {@link SourcePacketExtension} belongs to this
+     * {@link SimulcastGrouping}.
+     *
+     * @param src {@link SourcePacketExtension} to be checked.
+     *
+     * @return <tt>true</tt> or <tt>false</tt>.
+     */
+    public boolean belongsToSimulcastGrouping(SourcePacketExtension src)
+    {
+        if (fidGroups.size() > 0)
```
bbaldino: i think `if (!fidGroups.empty())` is a bit clearer for what's being tested here

src/main/java/org/jitsi/protocol/xmpp/util/SSRCSignaling.java
```diff
@@ -343,6 +343,38 @@ public static boolean mergeVideoIntoAudio(MediaSourceMap peerSSRCs)
...
+    public static List<SourceGroup> selectWithMsid(
+        List<SourceGroup>    groups,
+        String               groupMsid)
+    {
+        if (StringUtils.isNullOrEmpty(groupMsid))
+        {
+            throw new IllegalArgumentException("Null or empty 'groupMsid'");
+        }
+
+        List<SourceGroup> result = new LinkedList<>();
```
bbaldino: this could be:

```
return groups.stream()
       .filter(group -> group.getGroupMsid().equalsIgnoreCase(groupMsid)
       .collect(Collectors.toList());

```
which is a bit more concise and still readable i think, but i don't feel too strongly about changing it.

src/main/java/org/jitsi/protocol/xmpp/util/MediaSourceMap.java
```diff
@@ -192,6 +193,39 @@ public SourcePacketExtension findSourceViaSsrc(String media, long ssrc)
...
+            String    groupMsid)
+    {
+        if (StringUtils.isNullOrEmpty(groupMsid))
+        {
+            throw new IllegalArgumentException("Null or empty 'groupMsid'");
+        }
+
+        List<SourcePacketExtension> mediaSources
+            = getSourcesForMedia(mediaType);
+        List<SourcePacketExtension> result = new LinkedList<>();
```
bbaldino: same comment as below

src/main/java/org/jitsi/protocol/xmpp/util/MediaSourceGroupMap.java
```diff
@@ -70,6 +124,32 @@ private MediaSourceGroupMap(Map<String, List<SourceGroup>> map)
...
+     * @return a {@link List} of {@link SourceGroup}
+     */
+    private List<SourceGroup> findSourceGroups(String media, String semantics)
+    {
+        Objects.requireNonNull(semantics, "semantics");
+
+        List<SourceGroup> mediaGroups = groupMap.get(media);
+        List<SourceGroup> result = new LinkedList<>();
+
+        for (SourceGroup group : mediaGroups)
```
bbaldino: same comment as below

src/main/java/org/jitsi/protocol/xmpp/util/MediaSourceGroupMap.java
```diff
@@ -54,6 +54,60 @@ private MediaSourceGroupMap(Map<String, List<SourceGroup>> map)
     }
 
     /**
+     * Finds and extracts {@link SimulcastGrouping}s for given media type.
+     *
+     * @param mediaType eg. 'audio', 'video', etc.
+     *
+     * @return a {@link List} of {@link SimulcastGrouping}s.
+     */
+    public List<SimulcastGrouping> findSimulcastGroupings(String mediaType)
```
bbaldino: is `mediaType` needed here?  simulcast will only be for video

src/main/java/org/jitsi/jicofo/SSRCValidator.java
```diff
@@ -94,6 +211,59 @@ public SSRCValidator(String               endpointId,
...
+     * MSID, etc.) those have to be copied from the
+     * {@link SourcePacketExtension}s signalled in the media section in order
+     * to simplify the stream validation process by not having to look up
+     * another collection.
+     *
+     * @throws InvalidSSRCsException if a corresponding
+     * {@link SourcePacketExtension} signalled in {@link #sourceGroups} does not
+     * exist in {@link #sources}.
+     */
+    private void copySourceParamsToGroups()
```
bbaldino: i think it's better to have this be a standalone function which takes in only the data it needs to read/modify, that way it keeps it more isolated. (i.e. any members it's currently accessing, make it a non-member method and just pass those fields in as arguments and modify them in place ore return the modified version which the caller can assign).

src/main/java/org/jitsi/protocol/xmpp/util/SimulcastGrouping.java
paweldomas: SourceGroupPacketExtension is a packet extension where this one is a logical wrapper around simulcast grouping which consists of SIM + multiple FID groups

src/main/java/org/jitsi/protocol/xmpp/util/SourceGroup.java
paweldomas: Yes, we need because it wraps the logic of creating new SourcePacketExtension. Which first needs to create new instance, set semantics and then set extensions. If you would need to use that in any other place this will be duplicating this logic.

src/main/java/org/jitsi/protocol/xmpp/util/SimulcastGrouping.java
paweldomas: Packet extension is a Smack's thing and we used to only store there fields and attributes of the XML elements.

src/main/java/org/jitsi/protocol/xmpp/util/SimulcastGrouping.java
paweldomas: I will rework that, get rid of media type for simulcast in few other places as well

src/main/java/org/jitsi/protocol/xmpp/util/SSRCSignaling.java
paweldomas: I don't think we are at Java 8 already, but will add a FIXME, thanks !

src/main/java/org/jitsi/protocol/xmpp/util/MediaSourceGroupMap.java
paweldomas: I will keep this one as is, because it's generic and doesn't mention it's about video kind of groups (just an example in arg description). We don't know if there will be no groups for audio.

src/main/java/org/jitsi/jicofo/SSRCValidator.java
paweldomas: Hmm I don't know. It modifies it's member fields and that's the intention, to modify this specific fields knowing what they hold specifically.

paweldomas: I will extract this to some static utility method.

paweldomas:
@bbaldino I've addressed your comments here:
https://github.com/jitsi/jicofo/pull/227

src/main/java/org/jitsi/protocol/xmpp/util/SimulcastGrouping.java
bbaldino: what about `SourceGroup`?

src/main/java/org/jitsi/protocol/xmpp/util/SSRCSignaling.java
bbaldino: ah oops forgot about that. ok no worries.

PR #227, opened 2017-09-20
ref(SSRCValidator): improvements

paweldomas:
Gets rid of media type where video specific groups are in use. Other minor refactoring.


src/main/java/org/jitsi/protocol/xmpp/util/SSRCSignaling.java
```diff
@@ -362,6 +419,11 @@ public static boolean mergeVideoIntoAudio(MediaSourceMap peerSSRCs)
             throw new IllegalArgumentException("Null or empty 'groupMsid'");
         }
 
+        // FIXME change once migrated to Java 8
+        // return groups.stream()
+        //     .filter(group -> group.getGroupMsid().equalsIgnoreCase(groupMsid)
+        //     .collect(Collectors.toList());
+
```
bbaldino: i think we can just leave this out

paweldomas: no, lets keep it

bgrozev:
Should we merge this?

bbaldino:
jenkins, test this please

PR #229, opened 2017-10-11
Introduces StatsId and colibri changes around it

damencho:
Changes colibri to use <endpoint and stops using <channel-bundle. Uses statsId if available.

### DIFF
--- a/src/test/java/org/jitsi/jicofo/ColibriTest.java
+++ b/src/test/java/org/jitsi/jicofo/ColibriTest.java
@@ -113,13 +113,13 @@ public class ColibriTest
 
         ColibriConferenceIQ peer1Channels
             = colibriConf.createColibriChannels(
-                peer1UseBundle, peer1, true, contents);
+                peer1UseBundle, peer1, null, true, contents);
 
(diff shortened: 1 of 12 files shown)


bgrozev:
LGTM! Does it need an update of pom.xml?

damencho:
Yep, it needs update for the jitsi.desktop maven artifact.

src/test/java/org/jitsi/jicofo/ColibriTest.java
```diff
@@ -113,13 +113,13 @@ public void testChannelAllocation()
 
         ColibriConferenceIQ peer1Channels
             = colibriConf.createColibriChannels(
-                peer1UseBundle, peer1, true, contents);
-
+                peer1UseBundle, peer1, null, true, contents);
+        System.err.println("bbbbbbb  " + peer1Channels.toXML());
```
bgrozev: woops

damencho: yep, woops. will update it with the pom

PR #234, opened 2017-10-23
Connection closed error

damencho:
Fixes a problem where if there is a parse error the connection to the xmpp server is dropped and no more conferences are possible.

### DIFF
--- a/src/main/java/org/jitsi/impl/protocol/xmpp/extensions/ConferenceIqProvider.java
+++ b/src/main/java/org/jitsi/impl/protocol/xmpp/extensions/ConferenceIqProvider.java
@@ -24,6 +24,9 @@ import org.jivesoftware.smack.provider.*;
 
 import org.jxmpp.jid.*;
 import org.jxmpp.jid.impl.*;
+import org.jxmpp.jid.parts.*;
+import org.jxmpp.stringprep.*;
+
 import org.xmlpull.v1.*;
 
 /**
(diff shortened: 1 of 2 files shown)


src/main/java/org/jitsi/impl/protocol/xmpp/extensions/ConferenceIqProvider.java
```diff
@@ -172,4 +174,32 @@ else if (ConferenceIq.Property.ELEMENT_NAME.equals(name))
...
+    /**
+     * Constructs the jid for the room by taking the last '@' part as domain
+     * and everything before it as the node part. Doing validation on the node
+     * part for allowed chars.
+     *
+     * @param unescapedValue the unescaped jid as received in the iq
+     * @return a bare JID constructed from the given parts.
+     * @throws XmppStringprepException if an error occurs.
+     */
+    private EntityBareJid getRoomJid(String unescapedValue)
```
paweldomas: hmm what do we put here currently ? is it just the room name like "conference123" or is it full address like "conference123@muc.server.com" ?

paweldomas: technically full room address should be allowed I think, even if we don't do that in jitsi meet currently

paweldomas: is it that with smack 4 we're not able to pass just the name of the room to join ? without muc address ? then I guess it is fine this way

damencho: So this is conference123@muc.server.com. The problem I'm trying to fix is that if that room is malformed we try processing it, joining a room, a lot more stuff, talking with xmpp server using incorrect jid.
An example of incorrect jid, that I was using for testing it "damencho@g mail.com@muc.server.com", this seems to be a common mistake of mobile users, if you enter your email as room name and jicofo dies.
The method Localpart.from( do that validation of the node part of the jid, throws an error and a result iq with internal server error is sent back to the client.

damencho: The jicofo dies part is fixed with ExceptionLoggingCallback, this is just improvement to immediately reply with an error in such situations.

PR #238, opened 2017-10-27
Uses an error message to prevent NPE in smack.

damencho:
Smack currently do not support errors without text and NPE is thrown
when operating over these stanzas. There is a PR to fix that https://github.com/igniterealtime/Smack/pull/178


bgrozev:
If smack borks when the message is null we should require non-null in ErrorFactory

damencho:
@bgrozev I'm submitting a PR there and it will be fixed to not throw an error. It is a fix till the 4.2.2 is released.

damencho:
New PR created against 4.2 branch https://github.com/igniterealtime/Smack/pull/179

PR #239, opened 2017-10-31
Fixes 'Operation not permitted' on startup.

damencho:
java.nio.file.FileSystemException: /etc/jitsi/jicofo: Operation not permitted

### DIFF
--- a/resources/install/debian/postinst
+++ b/resources/install/debian/postinst
@@ -113,10 +113,7 @@ case "$1" in
             usermod -g jitsi jicofo
         fi
 
-        # we create home folder only if it doesn't exist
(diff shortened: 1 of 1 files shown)


bgrozev:
If "dpkg-statoverride --list /usr/share/jicofo/" returns true, then OWNER and GROUP are set to the owner and groups of "/use/share/jicofo". Then if "dpkg-statoverride --list /etc/jitsi/jicofo/" returns false, we chown /etc/jitsi/jicofo to $OWNER:$GROUP. Shouldn't we chowned them to jicofo:jitsi instead (since there are no overrides for /etc/jitsi/jicofo)?

PR #240, opened 2017-11-07
Refactor ChannelAllocator

bgrozev:
Refactors ChannelAllocator to make it more easily reusable for allocating bridge-to-bridge channels.

### DIFF
--- a/src/main/java/org/jitsi/impl/protocol/xmpp/colibri/ColibriConferenceImpl.java
+++ b/src/main/java/org/jitsi/impl/protocol/xmpp/colibri/ColibriConferenceImpl.java
@@ -457,7 +457,7 @@ public class ColibriConferenceImpl
      *
      * Methods exposed for unit test purpose.
      *
-     * @param endpointId the name of Colibri endpoint (conference participant)
+     * @param endpointId the ID of the Colibri endpoint.
      *
      * @return <tt>true</tt> if current thread is conference creator.
      *
@@ -475,7 +475,7 @@ public class ColibriConferenceImpl
      * Releases "create conference semaphore". Must be called to release the
      * semaphore possibly in "finally" block.
      *
-     * @param endpointId the name of colibri conference endpoint(participant)
+     * @param endpointId the ID of the Colibri endpoint.
      */
     protected void releaseCreateConferenceSemaphore(String endpointId)
     {
@@ -489,7 +489,7 @@ public class ColibriConferenceImpl
      *
      * Exposed for unit tests purpose.
      *
-     * @param endpointId Colibri endpoint name (participant)
+     * @param endpointId the ID of the Colibri endpoint.
      * @param request Colibri IQ to be send towards the bridge.
      *
      * @return <tt>Packet</tt> which is JVB response or <tt>null</tt> if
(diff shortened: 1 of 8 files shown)


src/main/java/org/jitsi/impl/protocol/xmpp/colibri/ColibriConferenceImpl.java
```diff
@@ -457,15 +457,15 @@ else if (!(response instanceof ColibriConferenceIQ))
      *
      * Methods exposed for unit test purpose.
      *
-     * @param endpointName the name of Colibri endpoint (conference participant)
+     * @param endpointId the name of Colibri endpoint (conference participant)
```
bbaldino: so is this the id or the name? (rest of comment still says 'name')

src/main/java/org/jitsi/impl/protocol/xmpp/colibri/ColibriConferenceImpl.java
```diff
@@ -475,9 +475,9 @@ protected boolean acquireCreateConferenceSemaphore(String endpointName)
      * Releases "create conference semaphore". Must be called to release the
      * semaphore possibly in "finally" block.
      *
-     * @param endpointName the name of colibri conference endpoint(participant)
+     * @param endpointId the name of colibri conference endpoint(participant)
```
bbaldino: same

src/main/java/org/jitsi/impl/protocol/xmpp/colibri/ColibriConferenceImpl.java
```diff
@@ -489,7 +489,7 @@ protected void releaseCreateConferenceSemaphore(String endpointName)
      *
      * Exposed for unit tests purpose.
      *
-     * @param endpointName Colibri endpoint name (participant)
+     * @param endpointId Colibri endpoint name (participant)
```
bbaldino: same

src/main/java/org/jitsi/jicofo/AbstractChannelAllocator.java
```diff
@@ -0,0 +1,394 @@
...
+import org.jitsi.jicofo.util.*;
+import org.jitsi.protocol.xmpp.*;
+import org.jitsi.protocol.xmpp.util.*;
+import org.jitsi.util.*;
+import org.jxmpp.jid.*;
+
+import java.util.*;
+
+/**
+ * A thread which allocates Colibri channels and (optionally) initiates a
```
bbaldino: i think it would be better to separate the logic for allocating the channels and initiating a session...keep classes to a single responsibility.  was there a reason to have this do both?

src/main/java/org/jitsi/protocol/xmpp/colibri/ColibriConference.java
```diff
@@ -94,7 +94,7 @@
      *
      * @param useBundle <tt>true</tt> if channel transport bundle should be used
      *                  for this allocation.
-     * @param endpointName the name that will identify channels endpoint.
+     * @param endpointId the name that will identify channels endpoint.
```
bbaldino: comment

bgrozev:
Will rebase and re-open

src/main/java/org/jitsi/jicofo/AbstractChannelAllocator.java
paweldomas: It was just that the allocation and invite happens on separate thread which is basically managed by this class. I don't really see the point in having another class just to send the invite. Jingle stuff is handled by separate thing anyway.

PR #241, opened 2017-11-07
Fixes memory leak in MultiUserChat and adds health check monitor.

damencho:


src/main/java/org/jitsi/jicofo/rest/Health.java
```diff
@@ -217,9 +225,21 @@ static synchronized void getJSON(
...
+                    monitor.start();
+                }
+
                 check(focusManager);
 
                 status = cacheStatus(HttpServletResponse.SC_OK);
+
+                if (monitor != null)
+                {
+                    monitor.stop();
```
paweldomas: this needs to by in try{ check(...); } finally { monitor.stop()} otherwise those will be leaking when check throws (see the outer try catch bloc).

src/main/java/org/jitsi/jicofo/FocusManager.java
```diff
@@ -116,6 +116,18 @@
         = "org.jitsi.jicofo.STATS_PUBSUB_NODE";
 
     /**
+     * A property to enable health check debug. Enabling this will result an
+     * extra thread which will be monitoring the health-checks execution times.
```
paweldomas: The description could mention that it will log full thread dump

src/main/java/org/jitsi/jicofo/FocusManager.java
```diff
@@ -657,6 +672,15 @@ public void registrationStateChanged(RegistrationStateChangeEvent evt)
     }
 
     /**
+     * Returns whether health check debug is enabled.
```
paweldomas: If you don't want to copy full description add @see and refer to the property which has the full description instead of copy pasting not very useful text. Also on @return you can just write boolean or "true if enabled" instead of copy paste.

src/main/java/org/jitsi/jicofo/rest/Health.java
```diff
@@ -271,4 +291,126 @@ private static int cacheStatus(int status)
 
         return bridgeSelector.listActiveJVBs();
     }
+
+    /**
+     * Health check monitor schedules execution with a delay
+     * {@link Health#BAD_HEALTH_CHECK_INTERVAL} if monitor is not stopped
+     * by the time it executes we consider a health check was takeing too much
```
paweldomas: typo: "takeing"

src/main/java/org/jitsi/jicofo/rest/Health.java
```diff
@@ -271,4 +291,126 @@ private static int cacheStatus(int status)
...
+     * time executing and we dumb the stack trace in the logs.
```
paweldomas: typo "dumb"

src/main/java/org/jitsi/jicofo/rest/Health.java
```diff
@@ -271,4 +291,126 @@ private static int cacheStatus(int status)
...
+
+        /**
+         * Starts the monitor. Schedules execution in separate thread
+         * after some interval, if monitor is not stopped and it executes
+         * we consider the health check took too much time.
+         */
+        public void start()
+        {
+            this.startedAt = System.currentTimeMillis();
+            this.monitorTask = FocusBundleActivator.getSharedThreadPool()
```
paweldomas: I was thinking that maybe it would be better to use it's own Timer for that task. It's because in case the shared scheduled executor is blocked with many scheduled tasks then it may not execute the monitor. We should not create more than 1 health task at a time, so too many threads should not be a concern here.

src/main/java/org/jitsi/jicofo/rest/Health.java
```diff
@@ -271,4 +291,126 @@ private static int cacheStatus(int status)
...
+                        }
+                    }
+                    dbg.append("\n\n");
+                }
+                logger.error("Health check took "
+                    + (System.currentTimeMillis() - this.startedAt)
+                    + " ms. \n"
+                    + dbg.toString());
+            }
+            catch (Throwable t)
```
paweldomas: What is the reason for catching Throwable here ?

damencho: In case something went wrong obtaining threads info and printing. But yeah it doesn't make sense as it is in a thread we don't care about, we care about the result of the print.

PR #244, opened 2017-11-13
fixes #225: Ensure H264 is advertised in upper case.

### LINKED ISSUE #225: H264 doesn't work in Chrome
Recently I noticed that h264 conferences don't work in Google Chrome. It seems this issue has been [known for a while](https://github.com/jitsi/jitsi-meet/issues/877#issuecomment-269054479) and is caused by a case sensitivity issue with Chrome.

Steps to reproduce:
1. Set up a Jitsi Meet environment using the [Quick Install](https://github.com/jitsi/jitsi-meet/blob/master/doc/quick-install.md) guide
2. In /etc/jitsi/meet/jitsi.example.com-config.js, add `preferH264: true`
3. Using two browsers, at least one of them Google Chrome (tested in 60.0.3112.113), navigate to the Jitsi Meet domain and observe that the video call never starts.

I'm happy to contribute a fix for this myself, but I'm not sure where Constants.H264 is defined. I'm very new to Java.

EDIT: I found it in jitsi/libjitsi /src/org/jitsi/service/neomedia/codec/Constants.java. Maybe this issue should be against that project instead?

gpolitis:


gpolitis:
@paweldomas Could you please take a look?

src/main/java/org/jitsi/jicofo/util/JingleOfferFactory.java
```diff
@@ -388,8 +388,10 @@ private void addVideoToContent(ContentPacketExtension content,
         }
 
         // a=rtpmap:107 H264/90000
-        PayloadTypePacketExtension h264
-            = addPayloadTypeExtension(rtpDesc, H264_PT, Constants.H264, 90000);
+        PayloadTypePacketExtension h264 = addPayloadTypeExtension(
+                // XXX(gp): older Chrome versions (users have reported 53/55/61)
+                // fail to enable h264, if the encoding name is in lower case.
+                rtpDesc, H264_PT, Constants.H264.toUpperCase(), 90000);
```
bbaldino: why not just change the constant?

gpolitis: Yeah, I thought about that but.. what if we break something in the Jitsi client? I'd be a little hesitant to change it there.

PR #246, opened 2017-11-30
Fix reconnection of client connection

damencho:


src/main/java/org/jitsi/impl/protocol/xmpp/XmppProtocolProvider.java
```diff
@@ -288,6 +295,12 @@ synchronized private boolean doConnect()
             // rely on Smack's built-in retries, as it will be handled by
             // the RetryStrategy
             connection.removeConnectionListener(connListener);
+
+            ReconnectionManager reconnectionManager =
```
paweldomas: shouldn't the equals sign be moved to the next line ?

damencho: ups

src/main/java/org/jitsi/impl/protocol/xmpp/XmppProtocolProvider.java
```diff
@@ -248,6 +255,12 @@ synchronized private boolean doConnect()
             connection.connect();
 
             connection.addConnectionListener(connListener);
+            ReconnectionManager.getInstanceFor(connection)
+                .addReconnectionListener(reConnListener);
+
+            discoInfoManager = new ScServiceDiscoveryManager(
```
paweldomas:
```
discoInfoManager
    = new ScServiceDiscoveryManager(
            XmppProtocolProvider.this,
            connection,
            new String[]{},
            new String[]{},
            false);
```

src/main/java/org/jitsi/impl/protocol/xmpp/XmppProtocolProvider.java
```diff
@@ -248,6 +255,12 @@ synchronized private boolean doConnect()
             connection.connect();
 
             connection.addConnectionListener(connListener);
+            ReconnectionManager.getInstanceFor(connection)
```
paweldomas:
```
ReconnectionManager
        .getInstanceFor(connection)
        .addReconnectionListener(reConnListener);
```

damencho:
Just fixed them.

PR #247, opened 2017-12-04
Add support for offering flexfec-03

bbaldino:


src/main/java/org/jitsi/jicofo/util/JingleOfferFactory.java
```diff
@@ -101,6 +108,9 @@
     public static final String ENABLE_REMB_PNAME
         = "org.jitsi.jicofo.ENABLE_REMB";
 
+    public static final String ENABLE_FLEXFEC_03_PNAME
```
gpolitis: JavaDoc missing

PR #255, opened 2018-01-18
Fix SSRC validation

paweldomas:
SSRCValidator used to check SSRC correctness only on per each participant basis, but it will not catch all of the problems. All sources have to be checked as a whole, from all participants, as this gets merged into one remote SDP on the client.

### DIFF
--- a/src/main/java/org/jitsi/jicofo/SSRCValidator.java
+++ b/src/main/java/org/jitsi/jicofo/SSRCValidator.java
@@ -74,7 +74,7 @@ public class SSRCValidator
      * The limit sources count per media type allowed to be stored by
      * each {@link Participant} at a time.
      */
-    private int maxSourceCount;
+    private final int maxSourceCount;
 
     /**
      * Filters out FID groups that do belong to any simulcast grouping.
@@ -212,16 +212,6 @@ public class SSRCValidator
     }
 
     /**
-     * Returns how many sources per media type per participant is allowed in the
-     * conference.
-     * @return an integer
-     */
-    public int getMaxSourceCount()
-    {
-        return this.maxSourceCount;
-    }
-
-    /**
      * Checks how many sources are current in the conference for given
      * participant.
      *
@@ -240,19 +230,6 @@ public class SSRCValidator
                     SSRCSignaling.getSSRCOwner(source), owner))
             .count();
 
-    }
-
-    /**
-     * Sets how many sources can be advertised by each participant of each media
-     * type.
-     * @param maxSourceCount an integer value. If lesser or equal to zero no
-     * sources will be allowed which may or may not make any sense, but this
-     * value is configurable with the value default of
-     * {@link JitsiMeetGlobalConfig#DEFAULT_MAX_SSRC_PER_USER}.
-     */
-    public void setMaxSourceCount(int maxSourceCount)
-    {
-        this.maxSourceCount = maxSourceCount;
     }
 
     /**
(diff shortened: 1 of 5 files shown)


src/main/java/org/jitsi/jicofo/SSRCValidator.java
```diff
@@ -211,6 +211,50 @@ public SSRCValidator(String              endpointId,
...
+     * @param mediaType The type of the media for which sources will be counted.
+     * @return how many sources are currently in the conference source map for
+     * given media type and owner's JID.
+     */
+    private long getSourceCountForOwner(Jid owner, String mediaType)
+    {
+        return this.sources.getSourcesForMedia(mediaType)
+            .stream()
+            .filter(
+                source -> Objects.equals(
```
bbaldino: is this doing a content comparison? or a reference comparison?

src/main/java/org/jitsi/jicofo/SSRCValidator.java
```diff
@@ -67,9 +72,9 @@
 
     /**
      * The limit sources count per media type allowed to be stored by
-     * the {@link Participant} at a time.
+     * each {@link Participant} at a time.
      */
-    private final int maxSourceCount;
+    private int maxSourceCount;
```
bbaldino: why can't we set this earlier anymore? (and keep it final) i thought this was a preconfigured/hard-coded thing

paweldomas: I added it when working on one of the tests, but when I'm looking at it now we can keep it final, will change that.

src/main/java/org/jitsi/jicofo/SSRCValidator.java
paweldomas: It will call .equals which will do content comparison if the first argument is not null. It will also return true if both are null.

paweldomas: https://docs.oracle.com/javase/7/docs/api/java/util/Objects.html#equals(java.lang.Object,%20java.lang.Object)

src/main/java/org/jitsi/jicofo/SSRCValidator.java
paweldomas: I've just pushed a fix - will rebase and squash it after all reviews are done.

bgrozev:
The code looks good to me. A general question (I'm probably missing some of the background). Is this the problem which we observed and which is fixed by the PR:
1. EndpointA re-connects and gets a new JID/endpoint id (EndpointB)
2. EndpointB uses the same SSRCs as A, and adds them in either session-accept or source-add
3. The updated offer to EndpointC contains the same SSRCs twice, so C fails sRD

If this is correct, then I see how the PR fixes the problem, but wouldn't it result in EndpointB's jingle-session to brake? Do we solve this problem somewhere else (e.g. the client automatically reconnecting again)? I'm probably mis-understanding something here

paweldomas:
Yes, you're right about all that. EndpointB connection will break indeed I am thinking how this can be fixed, because we can hit that from mobile. I wonder if user's of the external API are able to hit that too.

paweldomas:
But this PR will make it that only the EndpointB will have it's Jingle session terminated/broken and it will prevent from breaking the whole conference.

bgrozev:
Yeah, that makes sense

gpolitis:
> 2. EndpointB uses the same SSRCs as A, and adds them in either session-accept or source-add

Isn't that called an SSRC conflict? Or have I misunderstood your description?

paweldomas:
@gpolitis Yes, it's SSRC conflict. The changes in this PR should protect from them. From both SSRC and MSID conflicts as well as few other cases which could lead to setRemoteDescription failure on the client.

paweldomas:
The tests have passed - please merge.

PR #260, opened 2018-02-21
Webhook for jicofo

pekubu:
Send event when user join, left and room ended to api.

### DIFF
--- a/src/main/java/org/jitsi/jicofo/FocusManager.java
+++ b/src/main/java/org/jitsi/jicofo/FocusManager.java
@@ -26,7 +26,7 @@ import org.jitsi.jicofo.event.*;
 import org.jitsi.service.configuration.*;
 import org.jitsi.eventadmin.*;
 import org.jitsi.util.Logger;
-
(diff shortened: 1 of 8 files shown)


pekubu:
So, now I have signed the CLA. What next?
@jitsi-jenkins

damencho:
Hi, sorry to say but I don't think this functionality belongs here. 
We needed something like this, and we had and internal discussion about that and we end up creating a prosody module. It is a very simple module to write:

-- left
module:hook("muc-occupant-left", handleOccupantLeft, 10);
-- joined
module:hook("muc-occupant-joined", handleOccupantJoined, 10);
-- update in the presence
module:hook("muc-broadcast-presence", handleBroadcastPresence, 10);

One other option is to add an event to the existing events in org.jitsi.jicofo.event, and there is a pending PR https://github.com/jitsi/jitsi-android-osgi/pull/1 and you will be able to sideload your code to use the EventAdmin from jicofo. But this depends on that PR and on jicofo been updated ... so not sure when this will happen, while the prosody module is so simple to implement and deploy, so I think this is the best option.

PR #261, opened 2018-03-01
Add option to always trust XMPP domain certificate.

guusdk:
When jicofo connects to the XMPP domain, it uses a default, Smack-provided
implementation that controls verification and validation of the server-provided
TLS certificate. In certain setups, it is desirable to override this behavior,
and allow any certificate that's provided.

This commit adds an option that, once enabled, causes Jicofo to trust any and
all certificate that is provided by the XMPP server that it conntects to.

### DIFF
--- a/src/main/java/org/jitsi/impl/protocol/xmpp/XmppProtocolProvider.java
+++ b/src/main/java/org/jitsi/impl/protocol/xmpp/XmppProtocolProvider.java
@@ -47,8 +47,7 @@ import org.jxmpp.stringprep.*;
 
 import javax.net.ssl.*;
 import java.io.*;
-import java.security.cert.CertificateException;
-import java.security.cert.X509Certificate;
+import java.security.cert.*;
 import java.util.*;
 import java.util.concurrent.*;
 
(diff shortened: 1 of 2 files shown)


bgrozev:
I've to admit that while I'm not excited about making it easier to misconfigure your deployment, it's quite useful in some situations (I'm running locally with a similar patch because it's easier than getting the certs configured).

src/main/java/org/jitsi/impl/protocol/xmpp/XmppProtocolProvider.java
```diff
@@ -42,7 +45,10 @@
 import org.jxmpp.jid.parts.*;
 import org.jxmpp.stringprep.*;
 
+import javax.net.ssl.*;
 import java.io.*;
+import java.security.cert.CertificateException;
+import java.security.cert.X509Certificate;
 import java.util.*;
 import java.util.concurrent.*;
```
damencho: Hi, @guusdk can you just fix to use star imports, thanks.

src/main/java/org/jitsi/impl/protocol/xmpp/XmppProtocolProvider.java
```diff
@@ -25,10 +25,13 @@
...
 import org.jitsi.eventadmin.*;
 import org.jitsi.impl.protocol.xmpp.colibri.*;
+import org.jitsi.jicofo.FocusBundleActivator;
+import org.jitsi.jicofo.FocusManager;
 import org.jitsi.jicofo.recording.jibri.*;
 import org.jitsi.protocol.xmpp.*;
 import org.jitsi.protocol.xmpp.colibri.*;
 import org.jitsi.retry.*;
+import org.jitsi.service.configuration.ConfigurationService;
 import org.jitsi.util.Logger;
```
damencho: @guusdk and what about FocusBundleActivator, FocusManager and ConfigurationService?

PR #262, opened 2018-03-01
Another iteration of the Octo support

bgrozev:
Part of this was reviewed in https://github.com/jitsi/jicofo/pull/240

### DIFF
--- a/src/main/java/org/jitsi/jicofo/AbstractChannelAllocator.java
+++ b/src/main/java/org/jitsi/jicofo/AbstractChannelAllocator.java
@@ -76,7 +76,7 @@ public abstract class AbstractChannelAllocator implements Runnable
     /**
      * The participant that is to be invited by this instance to the conference.
      */
-    protected final Participant participant;
+    protected final AbstractParticipant participant;
 
     /**
      * A flag which indicates whether channel allocation is canceled. Raising
@@ -123,7 +123,7 @@ public abstract class AbstractChannelAllocator implements Runnable
     protected AbstractChannelAllocator(
             JitsiMeetConferenceImpl meetConference,
             JitsiMeetConferenceImpl.BridgeSession bridgeSession,
-            Participant participant,
+            AbstractParticipant participant,
             boolean[] startMuted,
             boolean reInvite)
     {
@@ -162,7 +162,6 @@ public abstract class AbstractChannelAllocator implements Runnable
         throws OperationFailedException
     {
         List<ContentPacketExtension> offer;
-        Jid address = participant == null ? null : participant.getMucJid();
 
         if (canceled)
         {
@@ -183,7 +182,7 @@ public abstract class AbstractChannelAllocator implements Runnable
 
         if (colibriChannels == null)
         {
-            logger.error("Channel allocator failed: " + address);
+            logger.error("Channel allocator failed: " + participant);
 
             // Notify conference about failure
             meetConference.onChannelAllocationFailed(this);
@@ -211,7 +210,7 @@ public abstract class AbstractChannelAllocator implements Runnable
 
         try
         {
-            invite(participant, offer);
+            invite(offer);
         }
         catch (OperationFailedException e)
         {
@@ -224,8 +223,7 @@ public abstract class AbstractChannelAllocator implements Runnable
      * {@link AbstractChannelAllocator}, if there is one, in order to invite
      * (or re-invite) him to the conference.
      */
-    protected void invite(
-        Participant participant, List<ContentPacketExtension> offer)
+    protected void invite(List<ContentPacketExtension> offer)
         throws OperationFailedException
     {
     }
@@ -267,7 +265,8 @@ public abstract class AbstractChannelAllocator implements Runnable
             {
                 logger.info(
                         "Using " + jvb + " to allocate channels for: "
-                                 + participant.getMucJid());
+                             + (participant == null
+                                    ? "null" : participant.toString()));
 
                 ColibriConferenceIQ colibriChannels
                     = doAllocateChannels(contents);
@@ -371,7 +370,7 @@ public abstract class AbstractChannelAllocator implements Runnable
     /**
      * @return the {@link Participant} of this {@link AbstractChannelAllocator}.
      */
-    public Participant getParticipant()
+    public AbstractParticipant getParticipant()
     {
         return participant;
     }
--- a/src/main/java/org/jitsi/jicofo/AbstractChannelAllocator.java
+++ b/src/main/java/org/jitsi/jicofo/AbstractChannelAllocator.java
@@ -26,7 +26,7 @@ import org.jxmpp.jid.*;
 import java.util.*;
 
 /**
- * A thread which allocates Colibri channels and (optionally) initiates a
+ * A task which allocates Colibri channels and (optionally) initiates a
  * Jingle session with a participant.
  *
  * @author Pawel Domas
@@ -87,8 +87,8 @@ public abstract class AbstractChannelAllocator implements Runnable
     protected final boolean[] startMuted;
 
     /**
-     * Indicates whether or not this thread will be doing a "re-invite". It
-     * means that we're going to replace previous conference which has failed.
+     * Indicates whether or not this task will be doing a "re-invite". It
+     * means that we're going to replace a previous conference which has failed.
      * Channels are allocated on new JVB and peer is re-invited with
      * 'transport-replace' Jingle action as opposed to 'session-initiate' in
      * regular invite.
@@ -157,11 +157,6 @@ public abstract class AbstractChannelAllocator implements Runnable
     {
         List<ContentPacketExtension> offer;
 
-        if (canceled)
-        {
-            return;
-        }
-
         offer = createOffer();
         if (canceled)
         {
@@ -181,7 +176,7 @@ public abstract class AbstractChannelAllocator implements Runnable
             // Notify conference about failure
             meetConference.onChannelAllocationFailed(this);
 
-            // Cancel this thread - nothing to be done after failure
+            // Cancel this task - nothing to be done after failure
             cancel();
             return;
         }
@@ -192,12 +187,7 @@ public abstract class AbstractChannelAllocator implements Runnable
         }
 
         offer = updateOffer(offer, colibriChannels);
-        if (canceled)
-        {
-            return;
-        }
-
-        if (offer == null)
+        if (offer == null || canceled)
         {
             return;
         }
--- a/src/main/java/org/jitsi/jicofo/AbstractChannelAllocator.java
+++ b/src/main/java/org/jitsi/jicofo/AbstractChannelAllocator.java
@@ -1,0 +1,394 @@
+/*
+ * Jicofo, the Jitsi Conference Focus.
+ *
+ * Copyright @ 2015 Atlassian Pty Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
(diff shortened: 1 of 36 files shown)


src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
```diff
@@ -536,23 +535,25 @@ protected void onMemberJoined(final XmppChatMember chatRoomMember)
      * Creates a new {@link ColibriConference} instance for use by this
      * {@link JitsiMeetConferenceImpl}.
      */
-    private ColibriConference createNewColibriConference(Jid bridgeJid)
+    private ColibriConferenceImpl createNewColibriConference(Jid bridgeJid)
```
bbaldino: why does this need to return the sub-type?

src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
```diff
@@ -590,60 +591,93 @@ private void inviteChatMember(XmppChatMember chatRoomMember,
         }
     }
 
-    private BridgeSession inviteParticipant(
-            Participant participant,
-            boolean reInvite,
-            boolean[] startMuted)
+    /**
+     * Selects a {@link BridgeState} to use for a specific {@link Participant}.
```
bbaldino: i'm confused by this, selecting a 'bridge state' for a participant seems weird.  isn't it a bridge we're selecting?

bbaldino: and what is the relationship between this and the 'doSelect' for bridges above?

src/main/java/org/jitsi/protocol/xmpp/util/MediaSourceMap.java
```diff
@@ -236,33 +217,33 @@ public SourcePacketExtension findSourceViaSsrc(String media, long ssrc)
...
+        List<SourcePacketExtension> mediaSSRCs = getSourcesForMedia(mediaType);
+        return mediaSSRCs.stream()
+            .filter(
+                ssrc ->
+                {
+                    Jid ssrcOwner = SSRCSignaling.getSSRCOwner(ssrc);
+                    return ssrcOwner != null && ssrcOwner.equals(owner);
+                })
+            .findFirst()
+            .orElse(null);
```
bbaldino: could be:
```
return mediaSSRCs.stream()
    .map(ssrc -> SSRCSignaling.getSSRCOwner(ssrc))
    .filter(Objects::nonNull)
    .filter(ssrcOwner -> ssrcOwner.equals(owner))
    .findFirst()
    .orElse(null)
```

(i don't care if you change it, i just had to take a break going through the PR :)

src/main/java/org/jitsi/protocol/xmpp/util/MediaSourceGroupMap.java
```diff
@@ -215,62 +204,35 @@ public void addSourceGroup(String media, SourceGroup sourceGroup)
...
-                    theSame = false;
-                    break;
-                }
-            }
-            if (theSame)
-                return true;
-        }
-        return false;
+        return getSourceGroupsForMedia(mediaType).stream()
+            .anyMatch(group -> group.equals(toFind));
```
bbaldino: well that's a nice improvement over the old code :)

src/main/java/org/jitsi/protocol/xmpp/colibri/ColibriConference.java
```diff
@@ -104,19 +104,63 @@
      * @throws OperationFailedException if channel allocation failed due to
      *                                  network or bridge failure.
      */
+    default ColibriConferenceIQ createColibriChannels(
+        boolean                         useBundle,
```
bbaldino: i notice some places we're using this spacing style and others we're not/it's being removed?

src/main/java/org/jitsi/jicofo/AbstractChannelAllocator.java
```diff
@@ -0,0 +1,393 @@
...
+    {
+        List<ContentPacketExtension> offer;
+
+        if (canceled)
+        {
+            return;
+        }
+
+        offer = createOffer();
+        if (canceled)
```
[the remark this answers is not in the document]
bgrozev: The repeated checks for canceled are because some of the calls wait for network operations (e.g. allocateCandidates). I'll use "task" instead of "thread"

src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
bgrozev: It doesn't, nice catch

src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
bgrozev: The naming is not ideal (suggestions for improvement are welcome). BridgeState is the global state of a jitsi-videobridge instance that jicofo knows about (one BridgeState object per jitsi-videobridge instance). So this is what we're selecting. I've improved the documentation in BridgeState.java which explains what it is.
The selectBridge() method in JitsiMeetConferenceImpl contains some conference-specific logic for the selection (the "enforced" part), and failure handling.

src/main/java/org/jitsi/protocol/xmpp/colibri/ColibriConference.java
bgrozev: I only see this spacing style in a few places in jicofo (and not in other java projects), so I remove it in favor of the common one. Will fix this one.

src/main/java/org/jitsi/protocol/xmpp/util/MediaSourceMap.java
bgrozev: I tried, but the call to map() loses track of the object we want to return

bbaldino: oh oops, didn't notice that

src/main/java/org/jitsi/jicofo/AbstractParticipant.java
```diff
@@ -393,4 +396,15 @@ public void addSourcesAndGroups(MediaSourceMap         addedSources,
...
+
+    /**
+     * @return {@code true} if the session with this participant has already
+     * been established. Before the session is established, we are unable to
+     * update the remote state (e.g. the list of sources (SSRCs) of this
+     * participant).
+     */
+    public boolean isSessionEstablished()
+    {
+        return true;
```
bbaldino: should this method just be abstract?

src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
bbaldino: could it just be named `Bridge`?

src/main/java/org/jitsi/jicofo/AbstractChannelAllocator.java
```diff
@@ -0,0 +1,394 @@
...
+    protected final boolean[] startMuted;
+
+    /**
+     * Indicates whether or not this thread will be doing a "re-invite". It
+     * means that we're going to replace previous conference which has failed.
+     * Channels are allocated on new JVB and peer is re-invited with
+     * 'transport-replace' Jingle action as opposed to 'session-initiate' in
+     * regular invite.
+     */
+    protected final boolean reInvite;
```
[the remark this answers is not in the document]
paweldomas: It occurs when bridge dies (health check fails) and the conference is moved to another bridge. Then a different Jingle message is used: transport-replace vs session-initiate used on the initial offer.

src/main/java/org/jitsi/jicofo/AbstractChannelAllocator.java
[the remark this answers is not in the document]
paweldomas: There's no really an easy way to kill/shutdown a thread in java and as far as I know it was recommended to use a flag. Although maybe it could be done better or in a cleaner way than I did. If I remember correctly the intention there was to check for cancel just before/after some long running operation like sending XMPP message.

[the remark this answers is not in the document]
bgrozev: Yes, I think I added more checks for 'canceled' just to be on the safe side, and they might not make sense.

src/main/java/org/jitsi/jicofo/AbstractChannelAllocator.java
```diff
@@ -0,0 +1,394 @@
...
+ * @author Boris Grozev
+ */
+public abstract class AbstractChannelAllocator implements Runnable
+{
+    /**
+     * Error code used in {@link OperationFailedException} when Colibri channel
+     * allocation fails.
+     * FIXME: consider moving to OperationFailedException ?
+     */
+    final static int CHANNEL_ALLOCATION_FAILED_ERR_CODE = 21;
```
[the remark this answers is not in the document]
bgrozev: I think there are some constants in Jitsi that go up to 20 :)

src/main/java/org/jitsi/jicofo/AbstractChannelAllocator.java
```diff
@@ -0,0 +1,394 @@
...
+    protected final JitsiMeetConferenceImpl meetConference;
+
+    /**
+     * The {@link JitsiMeetConferenceImpl.BridgeSession} on which
+     * to allocate channels for the participant.
+     */
+    protected final JitsiMeetConferenceImpl.BridgeSession bridgeSession;
+
+    /**
+     * The participant that is to be invited by this instance to the conference.
```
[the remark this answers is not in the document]
bgrozev: The AbstractChannelAllocator

src/main/java/org/jitsi/jicofo/AbstractChannelAllocator.java
```diff
@@ -0,0 +1,394 @@
...
+     * return.
+     */
+    protected volatile boolean canceled = false;
+
+    /**
+     * First argument stands for "start audio muted" and the second one for
+     * "start video muted". The information is included as a custom extension in
+     * 'session-initiate' sent to the user.
+     */
+    protected final boolean[] startMuted;
```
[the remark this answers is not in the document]
bgrozev: This is code being moved from ChannelAllocator to AbstractChannelAllocator. Refactoring it would be nice, but I'd rather leave it for another time

src/main/java/org/jitsi/jicofo/AbstractChannelAllocator.java
```diff
@@ -0,0 +1,394 @@
...
+     * the offer.
+     * @param colibriChannels the {@link ColibriConferenceIQ} which represents
+     * the channels allocated on a jitsi-videobridge instance for the participant
+     * for which the Jingle offer is being prepared.
+     */
+    protected List<ContentPacketExtension> updateOffer(
+            List<ContentPacketExtension> offer,
+            ColibriConferenceIQ colibriChannels)
+    {
+        return offer;
```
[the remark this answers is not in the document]
bgrozev: It's a method which modifies the offer. The default implementation (which is overriden for real participants, and not overriden in octo) just doesn't do any modifications.

src/main/java/org/jitsi/jicofo/AbstractParticipant.java
bgrozev: Yes, will fix

src/main/java/org/jitsi/jicofo/AbstractChannelAllocator.java
```diff
@@ -0,0 +1,394 @@
...
+     * {@link org.jitsi.jicofo.JitsiMeetConferenceImpl.BridgeSession} which
+     * identifies the jitsi-videobridge instance on which to allocate channels.
+     * @param participant the participant to be invited.
+     * @param startMuted an array which must have the size of 2 where the first
+     * value stands for "start audio muted" and the second one for "video
+     * muted". This is to be included in client's offer.
+     * @param reInvite whether to send an initial offer (session-initiate) or
+     * a an updated offer (transport-replace).
+     */
+    protected AbstractChannelAllocator(
```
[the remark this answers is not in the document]
bgrozev: They could be separated. I would like to avoid doing it because it's a complex piece and it's tricky to modify without breaking it. And I actually like that the whole task (which we run in a separate thread) is all contained in this class (and its subclasses).
If you feel strongly about it I'll try to separate it somehow.

bgrozev:
jenkins, test this please

bgrozev:
jenkins, test this please

PR #263, opened 2018-03-07
ref: Simplifies mvn-exec.sh.

gpolitis:


gpolitis:
@damencho @paweldomas could you PTAL?

gpolitis:
Included in https://github.com/jitsi/jicofo/pull/272

PR #264, opened 2018-03-07
Fix jibri iq routing

bgrozev:


bgrozev:
> (that's many changes)

The first commit _should_ have no functional changes. We were already using EntityFullJid objects, and the commit only makes it explicit.

PR #266, opened 2018-03-20
Fix #203 - Disables GSSAPI

### LINKED ISSUE #203: Focus fails to connect when other SASL Mechs are offered by Openfire

speedy01:
Focus fails to connect when other SASLMechs are offered by Openfire. This should address #203


speedy01:
CLA has been signed

PR #271, opened 2018-03-29
Jvb full jid

bgrozev:


src/main/java/org/jitsi/jicofo/event/BridgeEvent.java
```diff
@@ -174,9 +174,9 @@ private BridgeEvent(String topic, Jid bridgeJid)
      * @return <tt>String</tt> which is a JID of the JVB for which this event
```
paweldomas: can we update the JSDoc here ? It's not a String anymore (it should have been updated on the previous change to DomainBareJid, but ...)

PR #272, opened 2018-03-30
feat: enableTcc, enableRemb, minParticipants.

gpolitis:
Depends on https://github.com/jitsi/lib-jitsi-meet/pull/730

### DIFF
--- a/src/main/java/org/jitsi/jicofo/JitsiMeetConfig.java
+++ b/src/main/java/org/jitsi/jicofo/JitsiMeetConfig.java
@@ -49,36 +49,35 @@ public class JitsiMeetConfig
      * *NOTE* It is meant to be used for automated testing of
      * the {@link LipSyncHack} only !
      */
-    public static final String AUDIO_PACKET_DELAY = "audioPacketDelay";
+    public static final String PNAME_AUDIO_PACKET_DELAY = "audioPacketDelay";
 
     /**
      * The name of channel last N configuration property. Should be non-negative
      * number. Pass <tt>-1</tt> to disable last N functionality.
      */
-    public static final String CHANNEL_LAST_N_PNAME = "channelLastN";
+    public static final String PNAME_CHANNEL_LAST_N = "channelLastN";
+
+    /**
+     * The name of the "disableRtx" property.
+     */
+    public static final String PNAME_DISABLE_RTX = "disableRtx";
 
     /**
      * The name of the property that enables the {@link LipSyncHack}.
      */
-    public static final String ENABLE_LIPSYNC = "enableLipSync";
-
-    /**
-     * The name of the property that determines the min partcipants to start
-     * the call.
-     */
-    public static final String MIN_PARTICIPANTS_PNAME = "minParticipants";
+    public static final String PNAME_ENABLE_LIPSYNC = "enableLipSync";
 
     /**
      * The name of the property which enables the inclusion of the REMB RTCP
      * in the offer.
      */
-    public static final String ENABLE_REMB_PNAME = "enableRemb";
+    public static final String PNAME_ENABLE_REMB = "enableRemb";
 
     /**
      * The name of the property which enables the inclusion of the TCC RTP
      * header extension in the offer.
      */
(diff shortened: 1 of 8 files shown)


gpolitis:
@paweldomas could you PTAL?

src/main/java/org/jitsi/jicofo/ParticipantChannelAllocator.java
```diff
@@ -103,14 +105,16 @@ public ParticipantChannelAllocator(
         {
             contents.add(
                 jingleOfferFactory.createAudioContent(
-                    disableIce, useDtls, config.stereoEnabled()));
+                    disableIce, useDtls, config.stereoEnabled(),
+                    config.isRembEnabled(), config.isTccEnabled()));
```
bgrozev: Why not use the local variables you defined above?

src/main/java/org/jitsi/jicofo/OctoChannelAllocator.java
```diff
@@ -89,11 +91,12 @@ public OctoChannelAllocator(
         List<ContentPacketExtension> contents = new ArrayList<>();
         contents.add(
             jingleOfferFactory.createAudioContent(
-                    !useIce, useDtls, config.stereoEnabled()));
+                    !useIce, useDtls, config.stereoEnabled(),
```
bgrozev: The content for audio uses the config, but the config for video is always false?

src/main/java/org/jitsi/jicofo/JitsiMeetConfig.java
```diff
@@ -124,6 +130,12 @@
      */
     public static final int START_BITRATE_DEFAULT = 800;
 
+    /**
+     * The default value for the "minParticipants" property.
+     */
+    private static final int MIN_PARTICIPANTS_DEFAULT = 2;
```
paweldomas: It is starting to get messy with more constants being added to this file. Some of them have PNAME, or DEFAULT suffix, some of them don't. Do you mind cleaning this up before things get out of hands ? What do you think about renaming all properties to PNAME_{property name} and DEFAULT_{property name} ? Then if we sort things alphabetically properties will be grouped with properties and defaults with defaults. If that's too much work or you don't like it, could you please at least sort all constants alphabetically ? This way defaults will be close to the properties.

gpolitis: 👍

src/main/java/org/jitsi/jicofo/OctoChannelAllocator.java
paweldomas: @gpolitis is this now fixed ? I don't see such issue in the code anymore, am I missing something ?

src/main/java/org/jitsi/jicofo/JitsiMeetConfig.java
paweldomas: thanks !

src/main/java/org/jitsi/jicofo/OctoChannelAllocator.java
gpolitis: yes, now it's fixed

paweldomas:
I know @bgrozev does not like force push, but the first commit is kind of unrelated, so I don't want to squash everything when merging. Is it ok to force push in this case Boris ?

gpolitis:
I can squash the commits with the same title, with @bgrozev approval.

bgrozev:
I don't mind a force push

gpolitis:
I re-arranged the commits, please either approve (and I'll merge later) or approve & merge.

bgrozev:
COMPILATION ERROR : 
22:11:27 [INFO] -------------------------------------------------------------
22:11:27 [ERROR] /home/jenkins/jicofo/src/test/java/org/jitsi/jicofo/ColibriThreadingTest.java:[323,40] method createAudioContent in class org.jitsi.jicofo.util.JingleOfferFactory cannot be applied to given types;

gpolitis:
@paweldomas or @bgrozev could you please re-approve the final version? I re-based and fixed the compilation error.

PR #275, opened 2018-04-18
add support for youtube broadcast id

bbaldino:


bbaldino:
confirmed that tests passed here but @paweldomas says there's an issue with the ios script that's causing the build to show up as failed, so merging anyway.

PR #285, opened 2018-05-10
Recording/live streaming/sip gateway signaling changes

bbaldino:
The changes here are pretty much just adapting to the ones described in [this PR](https://github.com/jitsi/jitsi/pull/475).  Will need to update the pom here after the jitsi PR is merged so don't merge this yet.

One note I'll make is that the 'retry' logic for Jibri was removed.  We were simplifying the flows a bit and decided to yank that out for now as it was based on issues with the old Jibri and I think we should verify there's still a need for it with the new Jibri now.  If we do need to add it back, it shouldn't be too hard...I just thought I'd try and simplify as much as possible for now.

### DIFF
--- a/src/main/java/org/jitsi/impl/protocol/xmpp/XmppProtocolProvider.java
+++ b/src/main/java/org/jitsi/impl/protocol/xmpp/XmppProtocolProvider.java
@@ -817,12 +817,9 @@ public class XmppProtocolProvider
         {
             try
             {
-                System.out.println("XmppProtocolProvider sending iq with response callback");
-                System.out.println("Connections reply timeout is: " + connection.getReplyTimeout());
                 connection.sendIqWithResponseCallback(iq, stanzaListener);
             } catch (NotConnectedException | InterruptedException e)
             {
-                System.out.println("Sending iq with response callback failed: " + e.toString());
                 e.printStackTrace();
             }
         }
--- a/src/main/java/org/jitsi/impl/protocol/xmpp/XmppProtocolProvider.java
+++ b/src/main/java/org/jitsi/impl/protocol/xmpp/XmppProtocolProvider.java
@@ -814,14 +814,9 @@ public class XmppProtocolProvider
 
         @Override
         public void sendIqWithResponseCallback(IQ iq, StanzaListener stanzaListener)
+            throws NotConnectedException, InterruptedException
         {
-            try
-            {
-                connection.sendIqWithResponseCallback(iq, stanzaListener);
-            } catch (NotConnectedException | InterruptedException e)
-            {
-                e.printStackTrace();
-            }
+            connection.sendIqWithResponseCallback(iq, stanzaListener);
         }
 
         @Override
--- a/src/main/java/org/jitsi/impl/protocol/xmpp/XmppProtocolProvider.java
+++ b/src/main/java/org/jitsi/impl/protocol/xmpp/XmppProtocolProvider.java
@@ -252,8 +252,6 @@ public class XmppProtocolProvider
         }
 
         connection = new XMPPTCPConnection(connConfig.build());
-        // Jibri can take a while to reply, so give a decent-sized timeout
-        connection.setReplyTimeout(60000);
 
         if (logger.isTraceEnabled())
         {
@@ -813,10 +811,14 @@ public class XmppProtocolProvider
         }
 
         @Override
-        public void sendIqWithResponseCallback(IQ iq, StanzaListener stanzaListener)
+        public void sendIqWithResponseCallback(
+                IQ iq,
+                StanzaListener stanzaListener,
+                ExceptionCallback exceptionCallback,
+                long timeout)
             throws NotConnectedException, InterruptedException
         {
-            connection.sendIqWithResponseCallback(iq, stanzaListener);
+            connection.sendIqWithResponseCallback(iq, stanzaListener, exceptionCallback, timeout);
         }
 
         @Override
--- a/src/main/java/org/jitsi/impl/protocol/xmpp/XmppProtocolProvider.java
+++ b/src/main/java/org/jitsi/impl/protocol/xmpp/XmppProtocolProvider.java
@@ -252,6 +252,8 @@ public class XmppProtocolProvider
         }
 
         connection = new XMPPTCPConnection(connConfig.build());
+        // Jibri can take a while to reply, so give a decent-sized timeout
+        connection.setReplyTimeout(60000);
 
         if (logger.isTraceEnabled())
         {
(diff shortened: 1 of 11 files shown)


src/main/java/org/jitsi/impl/protocol/xmpp/XmppProtocolProvider.java
```diff
@@ -252,6 +252,8 @@ public boolean verify(String s, SSLSession sslSession)
         }
 
         connection = new XMPPTCPConnection(connConfig.build());
+        // Jibri can take a while to reply, so give a decent-sized timeout
+        connection.setReplyTimeout(60000);
```
paweldomas: This affects the whole connection, so I'm kind of worried if it would break other things...

paweldomas: Can we add a timeout argument to sendIqWithResponseCallback instead ?

src/main/java/org/jitsi/jicofo/recording/jibri/JibriSession.java
```diff
@@ -233,59 +212,33 @@ public JibriSession(
      * Starts this session. A new Jibri instance will be selected and start
      * request will be sent (in non blocking mode).
      */
-    synchronized public void start()
+    synchronized public Boolean start()
```
paweldomas: Can you describe the return argument in the docs since it's not void anymore ? Also it should be fine to go with the boolean primitive instead of Boolean, as we don't need the 'null' case here.

src/main/java/org/jitsi/jicofo/recording/jibri/JibriSession.java
```diff
@@ -300,9 +253,34 @@ private void sendStopIQ(XMPPError error)
 
         logger.info("Trying to stop: " + stopRequest.toXML());
 
-        xmpp.sendStanza(stopRequest);
+        // When we send stop, we won't get an OFF presence back (just
+        // a response to this message) so clean up the session
+        // in the processing of the response.
+        try
+        {
+            xmpp.sendIqWithResponseCallback(stopRequest, stanza -> {
```
paweldomas: what happens when there is a timeout ? does the sendIqWithResponseCallback method throw a timeout exception ? If yes, then is it handled ?

src/main/java/org/jitsi/jicofo/recording/jibri/JibriSession.java
```diff
@@ -335,127 +313,41 @@ public IQ processJibriIqFromJibri(JibriIq iq)
...
         }
 
-        currentJibriJid = null;
-        retryAttempt = 0;
+        return IQ.createResultIQ(iq);
+    }
 
-        // First we'll send an error and then follow with availability status
-        if (error != null)
+    public JibriIq.RecordingMode getRecordingMode()
```
paweldomas: Since the method is public we need it to be documented.

src/main/java/org/jitsi/jicofo/recording/jibri/JibriSession.java
```diff
@@ -543,24 +436,23 @@ private void setJibriStatus(JibriIq.Status newStatus, XMPPError error)
...
      * @return SIP address received from Jitsi Meet, which is used for SIP
      * gateway session (makes sense only for SIP sessions).
      */
-    public String getSipAddress()
+    String getSipAddress()
     {
         return sipAddress;
     }
 
+    public String getSessionId()
```
paweldomas: missing docs here as well

src/main/java/org/jitsi/jicofo/recording/jibri/Utils.java
```diff
@@ -0,0 +1,37 @@
...
+package org.jitsi.jicofo.recording.jibri;
+
+import java.util.*;
+
+/**
+ * @author bbaldino
+ */
+public class Utils
+{
+    public static String generateSessionId(int length) {
```
paweldomas: in java we always put { on new line plus the docs are missing

src/main/java/org/jitsi/protocol/xmpp/XmppConnection.java
```diff
@@ -61,4 +61,7 @@ IQ sendPacketAndGetReply(IQ packet)
     IQRequestHandler registerIQRequestHandler(IQRequestHandler handler);
 
     IQRequestHandler unregisterIQRequestHandler(IQRequestHandler handler);
+
+    void sendIqWithResponseCallback(IQ iq, StanzaListener stanzaListener)
```
paweldomas: missing docs

src/test/java/mock/xmpp/MockXmppConnection.java
```diff
@@ -235,6 +235,16 @@ public IQ sendPacketAndGetReply(IQ packet)
         }
     }
 
+    @Override
```
paweldomas: you can put empty docs tag here with {@inheritDoc} once they are filled in the interface

src/main/java/org/jitsi/impl/protocol/xmpp/XmppProtocolProvider.java
bbaldino: Ah! Didn't notice it had an overload with a timeout argument there.  Yeah I'll try that.

src/main/java/org/jitsi/jicofo/recording/jibri/JibriSession.java
bbaldino: fixed

src/main/java/org/jitsi/jicofo/recording/jibri/JibriSession.java
bbaldino: fixed

src/main/java/org/jitsi/jicofo/recording/jibri/JibriSession.java
bbaldino: fixed

src/main/java/org/jitsi/jicofo/recording/jibri/Utils.java
bbaldino: fixed

src/test/java/mock/xmpp/MockXmppConnection.java
bbaldino: i'll add this but no methods in this file have any javadocs it seems.

src/main/java/org/jitsi/protocol/xmpp/XmppConnection.java
bbaldino: fixed

src/test/java/mock/xmpp/MockXmppConnection.java
bbaldino: turns out i didn't need this anymore

PR #286, opened 2018-05-10
Revert jingle retry

bgrozev:


src/main/java/org/jitsi/jicofo/ParticipantChannelAllocator.java
```diff
@@ -217,10 +217,12 @@ else if (!canceled)
...
             // Whether another thread intentionally canceled us, or there was
-            // a failure to invite the participant in the jingle level, we will
-            // not trigger a retry here.
-            // In any case, try and expire the channels on the bridge.
-            bridgeSession.terminate(participant);
+            // a failure to invite the participant on the jingle level, we will
+            // not trigger a retry here. To assure that the conference doesn't
+            // attempt a retry, we pretend that this channel allocator is for
+            // a reinvite.
+            reInvite = true;
```
paweldomas: It used to be that the reInvite flag's meaning is to distinguish between the Jingle message that needs to go to the participant ('session-initiate' vs 'transport-replace'). Wouldn't it be less confusing to come up with a separate flag which meaning would be clear that it's about "do not retry to allocate channels" instead of mixing up the two concepts together ?

paweldomas:
Btw. what was the reasoning behind the logic to retry channel allocation only if it was not a reinvite ? Was it solving some specific issue ?

paweldomas:
Btw. the tests are fine - only iOS failed because of a known issue

src/main/java/org/jitsi/jicofo/ParticipantChannelAllocator.java
bgrozev: Sure ;)

bgrozev:
> Btw. what was the reasoning behind the logic to retry channel allocation only if it was not a reinvite ? Was it solving some specific issue ?

AFAIK the reason is to retry only once.

src/main/java/org/jitsi/jicofo/ParticipantChannelAllocator.java
```diff
@@ -218,11 +218,8 @@ else if (!canceled)
         {
             // Whether another thread intentionally canceled us, or there was
             // a failure to invite the participant on the jingle level, we will
-            // not trigger a retry here. To assure that the conference doesn't
-            // attempt a retry, we pretend that this channel allocator is for
-            // a reinvite.
-            reInvite = true;
-            meetConference.onChannelAllocationFailed(this);
+            // not trigger a retry here.
+            meetConference.onChannelAllocationFailed(this, false);
```
paweldomas: I find it very useful if arguments are commented sometimes, but not a big deal:
`meetConference.onChannelAllocationFailed(this, /* retry */ false);`

PR #287, opened 2018-05-11
Fix bridge down on jingle failure

bgrozev:


paweldomas:
iOS failed with this stupid dialog again

PR #291, opened 2018-06-12
fix: prevent hiding original exception when MUC start fails.

guusdk:
When starting/joining a MUC fails, 'leave' is called as part of the
exception handling. In case 'leave' throws an exception, this exception
should not be thrown instead of the original exception that was thrown
when start failed.

### DIFF
--- a/src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
+++ b/src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
@@ -364,7 +364,18 @@ public class JitsiMeetConferenceImpl
         }
         catch (Exception e)
         {
-            stop();
+            try
+            {
+                stop();
+            }
+            catch (Exception x)
+            {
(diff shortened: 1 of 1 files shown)


guusdk:
Jenkins reports a failing check, but all unit tests succeeded? What's going on here?

> All checks have failed
> 1 failing check
> 
> default — 771 tests run, 635 skipped, 0 failed.

paweldomas:
The iOS build failed:
https://travis-ci.org/jitsi/jitsi-meet/jobs/391160464
I'll try to re-run

guusdk:
Thanks. From my perspective, a considerable amount of black magic is involved with these tests. ;)

paweldomas:
It failed, but is acceptable (iOS physical device issue):
http://pr-reports.jitsi.org/jicofo-PR-test/401/ios-chrome-reports/html/

bgrozev:
Yay!

PR #302, opened 2018-07-30
Update README.md

neilzone:
Added section for jigasi. (I know this is a jicofo README, but it appears to be the main set of instructions for secure domain.)


saghul:
Cheers Neil!

PR #305, opened 2018-08-08
Pass through Jibri application data

bbaldino:


bbaldino:
This will also need to have the pom updated once https://github.com/jitsi/jitsi/pull/517 is in.

paweldomas:
Looks like there's quite a list of options passed to JibriSession... I wonder if it would make sense to make session id, stream id, broadcast id etc. part of the application data. Probably a bigger task though, as it would require changes in the Jibri itself including old Python Jibri.

bbaldino:
> Looks like there's quite a list of options passed to JibriSession... I wonder if it would make sense to make session id, stream id, broadcast id etc. part of the application data. Probably a bigger task though, as it would require changes in the Jibri itself including old Python Jibri.

Yeah I thought the same thing, but agree that it felt like something to do separately.

PR #307, opened 2018-08-13
implement functionality to invite transcriber when requested by client

nikvaessen:
jitsi-desktop-version still needs to be updated

### DIFF
--- a/src/main/java/org/jitsi/jicofo/jigasi/TranscriberManager.java
+++ b/src/main/java/org/jitsi/jicofo/jigasi/TranscriberManager.java
@@ -29,6 +29,8 @@ import org.jivesoftware.smack.packet.IQ;
 import org.jivesoftware.smack.packet.Presence;
 import org.jxmpp.jid.*;
 
+import java.util.concurrent.*;
+
 /**
  * The {@link TranscriberManager} class is responsible for listening to
  * {@link ChatRoomMemberPropertyChangeEvent}s to see whether a
@@ -68,6 +70,12 @@ public class TranscriberManager
     private volatile boolean active;
 
     /**
+     * A single-threaded {@link ExecutorService} to offload inviting the
+     * Transcriber from the smack thread updating presence.
+     */
+    private ExecutorService executorService;
+
+    /**
      * Create a {@link TranscriberManager} responsible for inviting Jigasi as
      * a transcriber when this is desired.
      *
@@ -94,9 +102,16 @@ public class TranscriberManager
      */
     public void init()
     {
+        if(executorService != null)
+        {
+            executorService.shutdown();
+            executorService = null;
+        }
+
+        executorService = Executors.newSingleThreadExecutor();
         chatRoom.addMemberPropertyChangeListener(this);
 
-        logger.info("initialised transcriber manager");
+        logger.debug("initialised transcriber manager");
     }
 
     /**
@@ -105,9 +120,11 @@ public class TranscriberManager
      */
     public void dispose()
     {
+        executorService.shutdown();
+        executorService = null;
         chatRoom.removeMemberPropertyChangeListener(this);
 
-        logger.info("disposed transcriber manager");
+        logger.debug("disposed transcriber manager");
     }
 
     /**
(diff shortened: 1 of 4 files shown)


src/main/java/org/jitsi/jicofo/jigasi/TranscriberManager.java
```diff
@@ -0,0 +1,292 @@
...
+
+    /**
+     * Initialise the manager by starting to listen for
+     * {@link ChatRoomMemberPropertyChangeEvent}s.
+     */
+    public void init()
+    {
+        chatRoom.addMemberPropertyChangeListener(this);
+
+        logger.info("initialised transcriber manager");
```
bgrozev: Can we change the level to debug? Also, do we really need a separate `init` method (instead of adding the listener inline in the constructor)?

src/main/java/org/jitsi/jicofo/jigasi/TranscriberManager.java
```diff
@@ -0,0 +1,292 @@
...
+
+    /**
+     * Stop the manager by stopping to listen for
+     * {@link ChatRoomMemberPropertyChangeEvent}s.
+     */
+    public void dispose()
+    {
+        chatRoom.removeMemberPropertyChangeListener(this);
+
+        logger.info("disposed transcriber manager");
```
bgrozev: Same, log level debug

src/main/java/org/jitsi/jicofo/jigasi/TranscriberManager.java
```diff
@@ -0,0 +1,292 @@
...
+    }
+
+    /**
+     * Method which is able to invite the transcriber by dialing Jigasi
+     */
+    private synchronized void startTranscribing()
+    {
+        logger.info("Attempting to invite transcriber");
+
+        Jid jigasiJid = jigasiDetector.selectJigasi();
```
bgrozev: Check for success

src/main/java/org/jitsi/jicofo/jigasi/TranscriberManager.java
```diff
@@ -0,0 +1,292 @@
...
+
+        RayoIqProvider.DialIq dialIq = new RayoIqProvider.DialIq();
+        dialIq.setDestination("jitsi_meet_transcribe");
+        dialIq.setTo(jigasiJid);
+        dialIq.setType(IQ.Type.set);
+        dialIq.setHeader("JvbRoomName", chatRoom.getName());
+
+        try
+        {
+            IQ response = this.connection.sendPacketAndGetReply(dialIq);
```
bgrozev: Is it safe to block the thread that called up (I think that'd be a smack thread updating presence)?

src/main/java/org/jitsi/jicofo/jigasi/TranscriberManager.java
```diff
@@ -0,0 +1,292 @@
...
+
+            if (response != null)
+            {
+                if (response.getError() == null)
+                {
+                    active = true;
```
bgrozev: We should handle the case where jigasi received the request but failed to join the conference. Perhaps we can leave this for a future improvement

bgrozev:
jenkins add to whitelist

src/main/java/org/jitsi/jicofo/jigasi/TranscriberManager.java
```diff
@@ -0,0 +1,312 @@
...
+            return;
+        }
+
+        if(containsTranscriptionStatus(presence))
+        {
+            maybeStoppedTranscribing(presence);
+        }
+        if(isRequestingTranscriber(presence) && !active)
+        {
+            executorService.submit(this::startTranscribing);
```
bgrozev: We only update the 'active' flag after we get the response from jigasi (over the network). If we receive a second request (e.g. someone clicking the button rapidly) we'll submit two tasks and invite two jigasi's. What do you think about updating 'active' here, before submitting the task (the task can then set it to 'false' if something fails)?

PR #311, opened 2018-08-29
Updates detecting graceful shutdown

damencho:


src/main/java/org/jitsi/jicofo/jigasi/JigasiDetector.java
```diff
@@ -41,6 +41,25 @@
...
+    private static final String STAT_NAME_PARTICIPANTS = "participants";
+
+    /**
+     * The name of the stat that indicates the instance has entered graceful
+     * shutdown mode.
+     * {@code VideobridgeStatistics.SHUTDOWN_IN_PROGRESS}, but is defined
+     * separately to avoid depending on the {@code jitsi-videobridge} maven
+     * package.
+     */
+    public static final String STAT_NAME_SHUTDOWN_IN_PROGRESS
```
bgrozev: Can we have one copy of this constant in jicofo? There's one in JigasiDetector and one in Bridge.java

Issue #309, opened 2018-08-28
Deadlock in AbstractAuthAuthority


jmartine:
## Description
---

I'm seeing a deadlock when using the AbstractAuthAuthority code. This seem to happen when there rapid requests to create a conference are being processed. Here is the out put of a thread dump (I can attach the full dump if that would help):

```
Found one Java-level deadlock:
=============================
"pool-7-thread-17":
  waiting to lock monitor 0x00007f1c780375a8 (object 0x000000070048bdf0, a java.lang.Object),
  which is held by "pool-7-thread-10"
"pool-7-thread-10":
  waiting to lock monitor 0x00007f1c78035ef8 (object 0x0000000782512f08, a java.util.HashMap),
  which is held by "Smack-Single Threaded Executor 0 (0)"
"Smack-Single Threaded Executor 0 (0)":
  waiting to lock monitor 0x00007f1c780375a8 (object 0x000000070048bdf0, a java.lang.Object),
  which is held by "pool-7-thread-10"

Java stack information for the threads listed above:
===================================================
"pool-7-thread-17":
	at org.jitsi.jicofo.auth.AbstractAuthAuthority.processAuthentication(AbstractAuthAuthority.java:347)
	- waiting to lock <0x000000070048bdf0> (a java.lang.Object)
	at org.jitsi.jicofo.xmpp.FocusComponent.processExtensions(FocusComponent.java:292)
	at org.jitsi.jicofo.xmpp.FocusComponent.handleConferenceIq(FocusComponent.java:343)
	at org.jitsi.jicofo.xmpp.FocusComponent.handleIQSetImpl(FocusComponent.java:221)
	at org.jitsi.xmpp.component.ComponentBase.handleIQSet(ComponentBase.java:362)
	at org.xmpp.component.AbstractComponent.processIQRequest(AbstractComponent.java:515)
	at org.xmpp.component.AbstractComponent.processIQ(AbstractComponent.java:289)
	at org.xmpp.component.AbstractComponent.processQueuedPacket(AbstractComponent.java:239)
	at org.xmpp.component.AbstractComponent.access$100(AbstractComponent.java:81)
	at org.xmpp.component.AbstractComponent$PacketProcessor.run(AbstractComponent.java:1051)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
[...]

PR #312, opened 2018-08-30
fix deadlock

jmartine:
I believe this lock to be greedy.  I've tested by rapidly requesting conferences to jicofo.  With this fix jicofo is not running in to deadlock i reported here: https://github.com/jitsi/jicofo/issues/309


bgrozev:
I think Pawel's PR that got merged recently should address the same problem:
https://github.com/jitsi/jicofo/pull/317/files

PR #316, opened 2018-09-14
Add the server (videobridge) region to jingle

bgrozev:


src/main/java/org/jitsi/protocol/xmpp/JingleUtils.java
```diff
@@ -0,0 +1,72 @@
...
+ * Jicofo specific utilities for Jingle.
+ *
+ * @author Boris Grozev
+ */
+public class JingleUtils
+{
+    /**
+     * Adds a group packet extension to a {@link JingleIQ}, and a
+     * {@link BundlePacketExtension} to each of its contents. I.e. adds
+     * everything that we deem necessary to turn enable {@code bundle} in an
```
bbaldino: extra word `turn`?  or maybe i'm misreading this (as I'm not sure I quite follow what it's saying)

src/main/java/org/jitsi/protocol/xmpp/JingleUtils.java
```diff
@@ -0,0 +1,72 @@
...
+    /**
+     * Adds a {@link StartMutedPacketExtension} to a specific {@link JingleIQ}.
+     * @param jingleIQ the {@link JingleIQ} to add extensions to.
+     * @param audioMute the value to set for the {@code audio} attribute.
+     * @param videoMute the value to set for the {@code video} attribute.
+     */
+    public static void addStartMutedExtension(
+        JingleIQ jingleIQ, boolean audioMute, boolean videoMute)
+    {
+        // FIXME Move this to a place where offer's contents are created or
```
bbaldino: what's stopping us from doing this now?

src/main/java/org/jitsi/jicofo/ParticipantChannelAllocator.java
```diff
@@ -363,7 +397,8 @@ else if (reInvite)
             }
             // Existing peers SSRCs
             RtpDescriptionPacketExtension rtpDescPe
-                = JingleUtils.getRtpDescription(cpe);
+                = net.java.sip.communicator.impl.protocol.jabber.jinglesdp
+                    .JingleUtils.getRtpDescription(cpe);
```
bbaldino: i wonder if it's worth renaming this to `JicofoJingleUtils` to 1) avoid the long qualified naming and 2) further denote that these are jicofo-specific jingle util functions

src/main/java/org/jitsi/jicofo/ParticipantChannelAllocator.java
```diff
@@ -233,6 +203,70 @@ else if (reInvite)
...
+     * {@code transport-replace}) and sends it to the {@code participant}.
+     * Blocks until a response is received or a timeout occurs.
+     *
+     * @param address the destination JID.
+     * @param contents the list of contents to include.
+     * @return {@code false} on failure.
+     * @throws OperationFailedException if we are unable to send a packet
+     * because the XMPP connection is broken.
+     */
+    private boolean sendJingleIq(
```
bbaldino: it seems a little odd to me that we have this generic `sendJingleIq` function which has to then look at fields to figure out which JingleIq actually needs to be created and sent.  are there not other places (which presumably trigger/call this) which already know what is happening and can create the necessary more directly?

src/main/java/org/jitsi/protocol/xmpp/JingleUtils.java
bgrozev: That's just a stale comment, I removed it.

src/main/java/org/jitsi/jicofo/ParticipantChannelAllocator.java
bgrozev: > are there not other places (which presumably trigger/call this) which already know what is happening and can create the necessary more directly?

Yes. There's only one place where this is used, and the only reason I extracted it is to avoid having the code inline (it's 50 lines of code in a method that is already over 50 lines)

bbaldino: ok i see now.  is the comment inaccurate, though?  it says "creates either `session-accept` or `transport-replace`" but the code creates either `session-initiate` or `transport-replace`.  and, while fixing that (unless I'm misunderstanding) could you add to the comment how it decides to make one vs the other?  (it looks like it's based on the `reInvite` member variable or the participant having a jingle session?)

PR #320, opened 2018-10-19
handle the new jibri status presence extension

bbaldino:


bbaldino:
depends on https://github.com/jitsi/jitsi/pull/539, will need a pom update

PR #321, opened 2018-10-22
add a helper to get non-health check conferences

bbaldino:
the tests do things like create conferences/channels and then verify how
many were created, but the mock bridge runs health checks which creates
conferences and channels on its own.  this adds a helper to get only the
conferences manually created (not the health-check-created ones)


bbaldino:
test failure is just a known ios test issue

PR #322, opened 2018-10-23
Try to fallback to another Jibri in case of an error

bbaldino:
If a jibri session encounters an error, we'll try and find another
available jibri to continue the current service.

### DIFF
--- a/src/main/java/org/jitsi/jicofo/recording/jibri/JibriSession.java
+++ b/src/main/java/org/jitsi/jicofo/recording/jibri/JibriSession.java
@@ -20,6 +20,7 @@ package org.jitsi.jicofo.recording.jibri;
 import net.java.sip.communicator.impl.protocol.jabber.extensions.jibri.*;
 import net.java.sip.communicator.impl.protocol.jabber.extensions.jibri.JibriIq.*;
 import net.java.sip.communicator.service.protocol.*;
+import org.jetbrains.annotations.Nullable;
 import org.jitsi.eventadmin.*;
 import org.jitsi.jicofo.*;
 import org.jitsi.osgi.*;
@@ -231,6 +232,7 @@ public class JibriSession
             {
                 jibriEventHandler.start(FocusBundleActivator.bundleContext);
                 sendJibriStartIq(jibriJid);
+                logger.info("Starting session with Jibri " + jibriJid);
                 return true;
             }
             catch (Exception e)
@@ -272,8 +274,7 @@ public class JibriSession
                     stopRequest,
                     stanza -> {
                         JibriIq resp = (JibriIq)stanza;
-                        setJibriStatus(resp.getStatus(), null);
-                        cleanupSession();
+                        handleJibriStatusUpdate(currentJibriJid, resp.getStatus(), resp.getFailureReason());
                     },
                     exception -> {
                         logger.error("Error sending stop iq: " + exception.toString());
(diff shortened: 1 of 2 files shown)


src/main/java/org/jitsi/jicofo/recording/jibri/JibriSession.java
```diff
@@ -443,24 +440,62 @@ private void reschedulePendingTimeout()
     }
 
     /**
-     * Stores current Jibri status and notifies {@link #owner}.
-     * @param newStatus the new Jibri status to be set
-     * @param failureReason optional error for failed state.
+     * Handle a Jibri status update (this could come from an IQ response, a new IQ from Jibri, an XMPP event, etc.).
+     * This will handle:
+     * 1) Retrying with a new Jibri in case of an error
```
paweldomas: can you explain how will Jicofo try another Jibri exactly ? how the `jibriDetector` will know to try another one ?

src/main/java/org/jitsi/jicofo/recording/jibri/JibriSession.java
```diff
@@ -411,7 +408,7 @@ private void sendJibriStartIq(final Jid jibriJid)
         try
         {
             JibriIq result = (JibriIq)xmpp.sendPacketAndGetReply(startIq);
-            setJibriStatus(result.getStatus(), null);
+            handleJibriStatusUpdate(jibriJid, result.getStatus(), result.getFailureReason());
```
paweldomas: I'm not sure I remember things correctly, but isn't it the case that the `result` might be `null` in case of a timeout ? or will that trigger OperationFailedException ? then in case of a timeout we won't try another one neither we would mark the one that is not responding as unhealthy in any way, right ?

src/main/java/org/jitsi/jicofo/recording/jibri/JibriSession.java
bbaldino: do you mean just elaborate in the comment?  we just call `start` here which contains the logic to find an available jibri

paweldomas: I mean it's very likely that the selector will return the same one unless it would report a failure immediately

paweldomas: or do we rely on that it would be in the busy state ?

bbaldino: from jicofo's side we assume that a jibri that has caused a failed session will have marked itself as no longer available for use.  from jibri, we bubble the error up (the same one that resulted in the failure here) to our status and mark ourselves as unhealthy.  (note that this is new behavior for jibri and it has just started doing this as of https://github.com/jitsi/jibri/pull/159)

src/main/java/org/jitsi/jicofo/recording/jibri/JibriSession.java
bbaldino: i assume we'll get the exception--at least, if there's an issue there we're no worse off than we were before (and i don't believe we've seen any NPE here due to that).  the pending timeout would come in and clean things up.

paweldomas: It might be worth trying another one if the first one refuses to respond, but I guess we can do this incrementally, so I guess it's fine this way for now.

paweldomas:
The test are good, it's the iOS build still failing due to Travis permission issue I haven't solved yet.

PR #325, opened 2018-11-05
Fix index out of bounds

torreylyons:
Fix index out of bounds crash when description name does not exist or there
is no channel for the content. Example: Data content does not have a channel
but has SctpConnection instead.


bgrozev:
Thanks for your contribution. I don't think the issue is valid anymore (colibri1 has been removed)

PR #329, opened 2018-12-04
Updates libjitsi and jvb dependencies.

damencho:
Fixes tests of newly created jvb PRs.


damencho:
After merging the systemd PR in jvb we can update this one with the latest jvb maven version, but it is not a must.

PR #332, opened 2019-01-07
Retries sending rayo iq.

damencho:
Retries sending rayo iq two times. Hitting timeout waiting for response
returns null reply, then we retry till we exhaust the attempts, receive
a reply, there is an exception or there are no more instances to try. We
 filter the instances we already tried when selecting new one.

### DIFF
--- a/src/main/java/org/jitsi/jicofo/jigasi/JigasiDetector.java
+++ b/src/main/java/org/jitsi/jicofo/jigasi/JigasiDetector.java
@@ -105,6 +105,8 @@ public class JigasiDetector
     /**
      * Selects the jigasi instance that is less loaded.
      *
+     * @param filter a list of <tt>Jid</tt>s to be filtered from the list of
+     * available Jigasi instances. List that we do not want as a result.
      * @return XMPP address of Jigasi instance or <tt>null</tt> if there are
      * no Jigasis available currently.
      */
--- a/src/main/java/org/jitsi/jicofo/jigasi/JigasiDetector.java
+++ b/src/main/java/org/jitsi/jicofo/jigasi/JigasiDetector.java
@@ -22,6 +22,8 @@ import org.jitsi.jicofo.*;
 import org.jitsi.jicofo.xmpp.*;
 import org.jxmpp.jid.*;
 
+import java.util.*;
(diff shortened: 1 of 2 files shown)


src/main/java/org/jitsi/jicofo/jigasi/JigasiDetector.java
```diff
@@ -96,11 +98,28 @@ protected void notifyInstanceOffline(Jid jid)
...
+        return this.selectJigasi(null);
+    }
+
+    /**
+     * Selects the jigasi instance that is less loaded.
+     *
+     * @return XMPP address of Jigasi instance or <tt>null</tt> if there are
+     * no Jigasis available currently.
+     */
+    public Jid selectJigasi(List<Jid> filter)
```
paweldomas: can you document the 'filter' argument ?

src/main/java/org/jitsi/jicofo/jigasi/JigasiDetector.java
```diff
@@ -96,11 +98,28 @@ protected void notifyInstanceOffline(Jid jid)
      * no Jigasis available currently.
      */
     public Jid selectJigasi()
```
paweldomas: is this method still being used anywhere after this changes ?

src/main/java/org/jitsi/jicofo/jigasi/JigasiDetector.java
damencho: Ups. I will do that now.

src/main/java/org/jitsi/jicofo/jigasi/JigasiDetector.java
damencho: Yes, for transcriptions.

src/main/java/org/jitsi/jicofo/jigasi/JigasiDetector.java
paweldomas: Thanks ! LGTM otherwise

damencho: Done.

PR #335, opened 2019-01-17
Adds support for signalling the video content type RTP header extension.

gpolitis:


gpolitis:
Do we care about the IDs we assign to the RTP header extensions?

bbaldino:
> Do we care about the IDs we assign to the RTP header extensions?

We should probably have a centralized place where they're defined so it's less likely we accidentally introduce a conflict.  I assume chrome doesn't care/is flexible as to which ext uses which ID? Though it'd be nice to follow along with whatever chrome's defaults are.

bbaldino:
were the RIDs causing issues?  also I think it'd be nice if we separated out that change to a separate PR, it seems like an independent piece of work?

bgrozev:
LGTM

gpolitis:
@bbaldino why do we have RIDs on today? I thought they were not used anywhere

bbaldino:
> @bbaldino why do we have RIDs on today? I thought they were not used anywhere

Yeah, I think it was just an oversight to not have a flag for them.

gpolitis:
>> were the RIDs causing issues?

>no issues afaik, just figured we should disable it because it's not used

I disabled the framemarkings header extension for the same reason (just reverted the first commit in this PR)

bbaldino:
> > also I think it'd be nice if we separated out that change to a separate PR, it seems like an independent piece of work?
> 
> I'm not sure I understand, you would like me to open a separate PR for the RID flag?

I think it's nice to at least have separate commits for independent pieces (RID flag vs frame markings work) where possible but it's fine.

gpolitis:
> I think it's nice to at least have separate commits for independent pieces (RID flag vs frame markings work) where possible but it's fine.

Agreed, I'll re-arrange the commits.

bbaldino:
shoot, i did not notice the failed build and looks like it is legitimate.  Re-running it again just to double check, if it fails again I'll revert.

bbaldino:
```
20:52:34 [ERROR] /home/jenkins/jicofo/src/main/java/org/jitsi/jicofo/util/JingleOfferFactory.java:[373,60] cannot find symbol
20:52:34   symbol:   variable VIDEO_CONTENT_TYPE_URN
20:52:34   location: class org.jitsi.service.neomedia.RTPExtension
```

PR #337, opened 2019-01-17
Max jibri retries

bbaldino:
don't merge until https://github.com/jitsi/jicofo/pull/336 is in

### DIFF
--- a/src/main/java/org/jitsi/jicofo/recording/jibri/JibriSession.java
+++ b/src/main/java/org/jitsi/jicofo/recording/jibri/JibriSession.java
@@ -302,6 +302,7 @@ public class JibriSession
     {
         logger.info("Cleaning up current JibriSession");
         currentJibriJid = null;
+        numRetries = 0;
         try
         {
             jibriEventHandler.stop(FocusBundleActivator.bundleContext);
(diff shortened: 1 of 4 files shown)


src/main/java/org/jitsi/jicofo/recording/jibri/JibriSession.java
```diff
@@ -481,6 +499,7 @@ private void handleJibriStatusUpdate(
             {
                 // The fallback to another Jibri succeeded.
                 logger.info("Successfully resumed session with another Jibri");
+                numRetries++;
```
paweldomas: It's hard to understand what's the flow here. Can you explain why this works ?

paweldomas: I just don't get why it increases the numRetries on a successful start() ?

bbaldino: The `start()` here is in response to an error with the current Jibri and an attempted fallback to another Jibri.  So we're tracking here how many times a Jibri had an error and we fell back to another one.

bbaldino: Yeah, both before and after if the fallback fails, then it's game over.  I ran into a scenario where a Jibri can fail, but it's a "soft" failure, so it doesn't go unhealthy.  This means that Jicofo could try Jibri A -> soft failure, then it tries Jibri B -> soft failure, then it tries Jibri A again -> soft failure, etc. etc.  Clearly this indicates the failure may not actually be a "soft failure", but Jibris don't know that.  So I added this just to make sure we don't fall into some endless fallback loop.

> Those retries are how many times the session will be moved to another Jibri during one recording session ? But it's not a retry on a per request basis to the same Jibri, do I get it right ?

Yeah, exactly.  Retries/fallbacks in JibriSession refer to trying another Jibri.

bbaldino: Jicofo doesn't mark Jibri as not working after a soft failure, the idea is that Jibri knows best whether or not a failure is hard or soft, so that's up to Jibri.  There are some cases where neither will know, though.

> Anyway this code is kind of hard to understand, so to have it merged there must be a comment which explains how this works.

I'll add a comment and see if that helps

paweldomas: >Jicofo could try Jibri A -> soft failure, then it tries Jibri B -> soft failure
So in that case Jicofo may as well keep trying the same one 5 times ?

bbaldino: yes, that could happen.  and could be legitimate.  for example, if the web server was having network problems, that's not an issue with Jibri at all so we'd want to keep retrying until one was able to load the page successfully.

PR #342, opened 2019-01-30
Leaking conferences

paweldomas:
See commit messages for more details


paweldomas:
I'm merging because the failing tests are also failing on other, totally unrelated PRs the same way

PR #343, opened 2019-02-01
Revert "Adds support for signalling the video content type RTP header extension."

bbaldino:
Reverts jitsi/jicofo#335


bbaldino:
believe the failure here is just a flakey test.

PR #357, opened 2019-05-09
Adds support for java 11.

damencho:


damencho:
Here I can update jvb binary used for test after merging https://github.com/jitsi/jitsi-videobridge/pull/827

PR #358, opened 2019-05-09
Use bitrate instead of videostreams for load balancing.

bgrozev:


paweldomas:
If we know it's a better metric and will not be going back to video streams, maybe we should get rid of the logics for estimating the video stream count ?

bgrozev:
I think that makes sense, Pawel. But let's take a look at how well this works first. I'll add this to my list

paweldomas:
Btw. is the bitrate statistic relatively consistent over time ? Im wondering if there are big variations over time. Also I suspect it may not behave perfectly when starting up a lot of conferences/participants at the same time. That's why the "video stream" estimation was introduced, to handle well bursts of channel allocations

paweldomas:
I don't remember though why handling burst of channels was important. One case I can think of is conference migration when bridge goes down, but it's a corner case.

bgrozev:
Screenshot 2019-05-09 at 17 28 33

It seems pretty stable. We get updated stats every 5 seconds, so it should be fine

bgrozev:
> One case I can think of is conference migration when bridge goes down, but it's a corner case.

Oh, yes, that's a good point!

PR #367, opened 2019-06-13
Fix : use of health-checks REST API with Shibboleth authentication

daimoc:
When we enable the shibboleth authentication the health check REST API stop to work.
All HTTP requests are first processed by the Shibboleth Jetty Handler and this handler rejects the /about/health request du to missing auth mandatory parameter. 
The solution to enable health REST API and shibboleth is to switch jetty handlers order so HTTP request are firstly computed with the health-check handler and if the request is not an health-check it is then computed by the shibboleth part.


daimoc:
I signed the CLA.

damencho:
jenkins test this please.

PR #369, opened 2019-06-25
Octo fixes

bgrozev:


src/main/java/org/jitsi/jicofo/FocusManager.java
```diff
@@ -485,7 +485,8 @@ private String generateConferenceId()
             {
                 id
                     = jicofoShortId +
-                        Integer.toHexString(RANDOM.nextInt(0x1_0000));
+                        String.format(
+                            "%4h", RANDOM.nextInt(0x1_0000)).replace(' ', '0');
```
bbaldino: what was the issue here?

src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
```diff
@@ -1433,12 +1434,34 @@ public XMPPError onSessionAccept(
         participant.setRTPDescription(answer);
         participant.addTransportFromJingle(answer);
 
+        MediaSourceMap sourcesAdvertised
+                = MediaSourceMap.getSourcesFromContent(answer);
+        MediaSourceGroupMap sourceGroupsAdvertised
+                = MediaSourceGroupMap.getSourceGroupsForContents(answer);
+        if (sourcesAdvertised.isEmpty())
```
bbaldino: i know we talked about this, but there is a reason we *can't* just do this on the client, right?  i worry we'll introduce new issues by injecting fake ssrcs that the client isn't aware of.

bgrozev: I don't know, I haven't looked at the client at all. I agree with your concern, but I think it also applies to some extent if we make client changes.

src/main/java/org/jitsi/jicofo/FocusManager.java
bgrozev: The conference ID consists of one byte for an ID of the jicofo instance plus two bytes for the conference within jicofo. If the random number was < 0x1000 we'd get an ID like "ff123" (with jicofo ID "ff"). This gets serialized the same as 0ff123 which could clash with jicofo ID "0f" and conf ID "f123".

This change pads the random ID so for the case above we get "ff0123".

src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
bbaldino: the client would be a single source of truth with regards to its sources though.

src/main/java/org/jitsi/jicofo/FocusManager.java
bbaldino: it looks like `String.format("%04x", ...)` does what you need and i think is cleaner than the replace:

```
String.format("%04x", 5); // --> 0005
String.format("%04x", 0xFFFF); // --> ffff
```

src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
gpolitis: > the client would be a single source of truth with regards to its sources though.

True, at least temporarily, until ssrcs fade away completely.

bbaldino:
I don't follow, it would be the client who removes them, right? So it'd
still be the single source of truth.

bgrozev:
Okay, I did the digging. The client intentionally does NOT generate an SSRC when it's receive only. It was disabled [here](https://github.com/jitsi/lib-jitsi-meet/pull/466) as a workaround for (this chrome bug)[https://bugs.chromium.org/p/webrtc/issues/detail?id=7555]. The Chrome bug is still not fixed, and simulcast fails to be enabled if we generate the SSRC early. 

So, I'll go ahead with this "fix".

PR #375, opened 2019-08-05
disable  tests relying on mockjvb

bbaldino:
MockVideobridge relies on classes in jvb that are no longer defined
(with jvb 2.0) and so it fails to compile during PR tests (where we
don't use the explicit  jvb dependency in the pom but instead use the
snapshot).  we need to adapt MockVideobridge to use the new types,
or, look at mocking the bridge entirely and  no longer relying on the
real one.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -206,12 +206,12 @@
       <version>1.6</version>
       <scope>test</scope>
     </dependency>
-    <dependency>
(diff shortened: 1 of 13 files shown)


bbaldino:
superseded by https://github.com/jitsi/jicofo/pull/376/files

PR #377, opened 2019-08-14
Less verbose logs

paweldomas:
I've been trying to investigate one issue in Jicofo lately and those logs are flooding the SEVERE level with not important things.

### DIFF
--- a/src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
+++ b/src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
@@ -1407,7 +1407,7 @@ public class JitsiMeetConferenceImpl
         {
             String errorMsg
                 = "No participant found for: " + participantJid;
-            logger.error(errorMsg);
+            logger.warn(errorMsg);
             return XMPPError.from(XMPPError.Condition.item_not_found,
                     errorMsg).build();
         }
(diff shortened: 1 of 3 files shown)


src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
```diff
@@ -1407,7 +1407,7 @@ public XMPPError onSessionAccept(
         {
             String errorMsg
                 = "No participant found for: " + participantJid;
-            logger.error(errorMsg);
+            logger.warn(errorMsg);
```
bbaldino: this seems pretty bad to me, shouldn't it be error?  what's the situation it's happening where it isn't bad?

paweldomas: When looking at the code it can happen after participant was marked as expired by Jicofo(removed from participants list), because it didn't reply with an ack to the session-initiate. So lets say participant lost connectivity for a while and then came back, but as the current code goes this one is not recoverable - jicofo will not re-invite this participant(maybe should).

bbaldino: i just worry this is going to make real issues harder to find.

paweldomas: Yeah, the point is that currently those messages appear in the logs and nobody does anything about them. So do you think anyone would do in future? How would you know it's a bigger issue this time? It will still appear on the warning level.

paweldomas: I wanted to have the error level reserved for cases which when happen really mean something's wrong and there's no ambiguity. This case can happen during "normal" operation lets say.

PR #379, opened 2019-08-22
Exclude json-simple's compile dependency on junit.

JonathanLennox:
(This is fixed in json-simple's github repo, but they haven't pushed a maven release with the fix yet.)


bgrozev:
Jenkins, add to whitelist

PR #381, opened 2019-08-29
Separate sip & transcribe and announce jvb version

damencho:


src/main/java/org/jitsi/jicofo/jigasi/TranscriberManager.java
```diff
@@ -201,14 +201,26 @@ private void startTranscribing()
             return;
         }
 
+        selectTranscriber(2, null);
+    }
+
+    /**
+     * Sends the message to the selected jigasi (from brewery muc).
```
bgrozev: What message does this send? Does it also do the selection logic like the name suggests?

src/main/java/org/jitsi/jicofo/jigasi/TranscriberManager.java
```diff
@@ -201,14 +201,26 @@ private void startTranscribing()
...
+     * @param retryCount the number of attempts to be made for sending this iq,
+     * if no reply is received from the remote side.
+     * @param filterJigasi <tt>null</tt> or a list of jigasi Jids which
```
bgrozev: Can we rename to 'exclude' or something like that?

damencho:
@bgrozev I think I've addressed all the comments and moved the constants to jitsi-xmpp-extensions.

src/main/java/org/jitsi/jicofo/jigasi/JigasiDetector.java
```diff
@@ -91,45 +111,121 @@ protected void onInstanceStatusChanged(
     protected void notifyInstanceOffline(Jid jid)
     {}
 
+    /**
+     * Selects the jigasi instance that is less loaded.
```
bgrozev: How about "Selects a jigasi instance which supports transcription"?

src/main/java/org/jitsi/jicofo/jigasi/JigasiDetector.java
```diff
@@ -91,45 +111,121 @@ protected void onInstanceStatusChanged(
...
+     * @param preferredRegions a list of preferred regions.
+     * @param transcriber Whether we need to select a transcriber or sipgw
+     * jigasi instance.
      * @return XMPP address of Jigasi instance or <tt>null</tt> if there are
      * no Jigasis available currently.
      */
-    public Jid selectJigasi(List<Jid> filter)
+    public static Jid selectJigasi(
+        List<BrewInstance> instances,
+        List<Jid> filter,
```
bgrozev: How about renaming to "exclude"?

src/main/java/org/jitsi/jicofo/jigasi/JigasiDetector.java
```diff
@@ -91,45 +111,121 @@ protected void onInstanceStatusChanged(
...
+        boolean transcriber)
     {
-        BrewInstance lessLoadedInstance = null;
-        int numberOfParticipants = Integer.MAX_VALUE;
-        for (BrewInstance jigasi : instances)
+        // let's filter using the provided filter and those in graceful shutdown
+        List<BrewInstance> filteredInstances = instances.stream()
+            .filter(j -> filter == null || !filter.contains(j.jid))
+            .filter(j -> j.status == null
+                || !Boolean.parseBoolean(j.status.getValueAsString(
```
bgrozev: Can we add a method to `BrewInstance` which makes this check? For example `isInGracefulShutdown`?

src/main/java/org/jitsi/jicofo/jigasi/JigasiDetector.java
```diff
@@ -91,45 +111,121 @@ protected void onInstanceStatusChanged(
...
+                        STAT_NAME_SHUTDOWN_IN_PROGRESS)))
+            .collect(Collectors.toList());
+
+        // let's select by type, is it transcriber or sipgw
+        List<BrewInstance> selectedByCap =
+            filteredInstances.stream()
+            .filter(j ->  j.status != null
+                    && transcriber ?
+                        Boolean.parseBoolean(
```
bgrozev: Can we refactor the have these in the `BrewInstance`? Ideally it would be a jigasi specific `BrewInstance` class

src/main/java/org/jitsi/jicofo/jigasi/JigasiDetector.java
```diff
@@ -91,45 +111,121 @@ protected void onInstanceStatusChanged(
...
+
+        // let's select by region
+        // Prefer a jigasi in the participant's region.
+        List<BrewInstance> filteredByRegion = new ArrayList<>();
+        if (preferredRegions != null && !preferredRegions.isEmpty())
+        {
+            filteredByRegion = selectedByCap.stream()
+                .filter(j -> j.status == null
+                            || preferredRegions.contains(
+                                j.status.getValueAsString(STAT_NAME_REGION)))
```
bgrozev: Same as above. Or at least extract to a utility function somewhere

src/main/java/org/jitsi/jicofo/jigasi/JigasiDetector.java
```diff
@@ -91,45 +111,121 @@ protected void onInstanceStatusChanged(
...
+                .collect(Collectors.toList());
+        }
+
+        // Otherwise, prefer a jigasi in the local region.
+        if (filteredByRegion.isEmpty() && localRegion != null)
+        {
+            filteredByRegion = selectedByCap.stream()
+                .filter(j -> j.status == null
+                    || j.status.getValueAsString(STAT_NAME_REGION)
```
bgrozev: Use `localRegion.equals...` to avoid a possible NPE?

src/main/java/org/jitsi/jicofo/jigasi/JigasiDetector.java
```diff
@@ -91,45 +111,121 @@ protected void onInstanceStatusChanged(
...
      * Selects the jigasi instance that is less loaded.
      *
+     * @param filter a list of <tt>Jid</tt>s to be filtered from the list of
+     * available Jigasi instances. List that we do not want as a result.
+     * @param preferredRegions a list of preferred regions.
      * @return XMPP address of Jigasi instance or <tt>null</tt> if there are
      * no Jigasis available currently.
      */
-    public Jid selectJigasi()
+    public Jid selectJigasi(List<Jid> filter, List<String> preferredRegions)
```
bgrozev: We don't care about the order of the preferred regions, right? Can we make it a `Collection` or a `Set` and make sure it's non-null and doesn't contain null elements? 

In the future we will be using more than one bridge in a region, and not every deployment has regions configured (in which case the bridges' regions are null).

src/main/java/org/jitsi/jicofo/VersionBroadcaster.java
```diff
@@ -171,18 +172,16 @@ public void handleEvent(Event event)
...
 
-        // Videobridge
-        // It is not be reported for FOCUS_JOINED_ROOM_TOPIC
-        Jid bridgeJid
-                = (Jid)event.getProperty(EventFactory.BRIDGE_JID_KEY);
-        Version jvbVersion
-            = bridgeJid == null
-                ? null : meetServices.getBridgeVersion(bridgeJid);
-        if (jvbVersion != null)
+        String jvbVersions = conference.getBridges().stream()
```
bgrozev: Fun fact: `CONFERENCE_ROOM_TOPIC` is used when a new colibri conference is allocated. I had to follow 4 references (including an exaustive search on uses of a constant) to find this out, and it is not what I expected it to be based on the name. Also, (a comment I left 2 years ago)[https://github.com/jitsi/jicofo/blob/master/src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java#L2423] suggests that at the time I failed to understand why we fire the event. Luckily I failed to convince myself that we don't need it either, so I left it in. Using `EventAdmin` leads to hard to maintain code. 

I'm just ranting, I don't expect any change. Though I would like to refactor this eventually.

src/main/java/org/jitsi/jicofo/VersionBroadcaster.java
```diff
@@ -171,18 +172,16 @@ public void handleEvent(Event event)
...
+            .map(b -> b.getVersion())
+            .filter(Objects::nonNull)
+            .collect(Collectors.joining(", "));
```
bgrozev: I think we should treat it as a set (remove duplicates). Otherwise we'll see it change every time we add/remove a bridge to the conference.

src/main/java/org/jitsi/jicofo/MeetExtensionsHandler.java
```diff
@@ -240,8 +240,10 @@ private IQ handleRayoIQ(RayoIqProvider.DialIq dialIq, int retryCount,
         }
 
 
-        List<String> bridgeRegions = conference.getBridges()
-            .stream().map(b -> b.getRegion()).collect(Collectors.toList());
+        List<String> bridgeRegions = conference.getBridges().stream()
+            .map(b -> b.getRegion())
+            .filter(Objects::nonNull)
+            .collect(Collectors.toList());
```
bgrozev: Why not `toSet`? We don't care about the order, and we don't want duplicates, right?

PR #382, opened 2019-08-30
Move tests to use constants from jitsi-xmpp-extensions lib.

damencho:


damencho:
Unrelated flakey test failed on ff.

PR #383, opened 2019-09-03
update jitsi-utils

bbaldino:


bbaldino:
one test repeatedly failing (avatartest).  i feel confident at this point that the failure is unrelated.

PR #385, opened 2019-09-19
Communicate with jitsi-videobridge going through the MUC.

damencho:


src/main/java/org/jitsi/jicofo/BridgeMucDetector.java
```diff
@@ -69,10 +69,6 @@ public BridgeMucDetector(
               ColibriStatsExtension.ELEMENT_NAME,
               ColibriStatsExtension.NAMESPACE);
 
-        // We prefer to communicate with jitsi-videobridge without going
-        // through the MUC.
-        setUseOccupantJid(false);
```
bbaldino: do we have a use case for this to be true anymore? can the `useOccupantJid` member go away altogether?

damencho: Yep, I just removed it, thanks.

PR #386, opened 2019-09-19
update jvb version

bbaldino:


bbaldino:
this has a problem in that we get a conflict with the ice4j version since jicofo pulls in things that pull in prior ice4j versions, so we'll need to figure that out.

damencho:
I think at one moment we had explicit ice4j version in jicofo's pom xml because of that.

damencho:
And it was removed here: https://github.com/jitsi/jicofo/commit/ee35bbbf36f3e3ee5dbfeb1c27a081d56d66c068#diff-600376dffeb79835ede4a0b285078036

bbaldino:
thanks @damencho .  I tried re-adding the explicit ice4j dependency (just in the test scope) and that fixed the tests (locally at least, we'll see if the build passes).  we can talk about whether or not we want to do this, just want to verify that it works.

Issue #30, opened 2015-11-30
Build failed


matthias-w:
Hello,
I have a problem building the latest version. At first I tried to build as I was used to with 'ant dist.lin64'. I got a maven error and afterwards saw the comment in build.xml to download maven ant tasks.

I tried running the given 'mvn dependency...' command but got lots of error messages. These are also described here: http://lists.jitsi.org/pipermail/dev/2015-October/025520.html (seems to be the same issue with jitsi videobridge)

Then I downloaded 'maven-ant-tasks-2.1.3.jar' directly from apache website and run:
'ant -lib maven-ant-tasks-2.1.3.jar dist.lin64'

Result was:
...
Not a v4.0.0 POM. for project org.sonatype.oss:oss-parent at /home/jitsi/.m2/repository/org/sonatype/oss/oss-parent/3/oss-parent-3.pom

BUILD FAILED
/home/jitsi/jicofo-master/build.xml:54: Unable to resolve artifact: Unable to get dependency information: Unable to read the metadata file for artifact 'com.googlecode.concurrentlinkedhashmap:concurrentlinkedhashmap-lru:jar': Cannot find parent: org.sonatype.oss:oss-parent for project: com.googlecode.concurrentlinkedhashmap:concurrentlinkedhashmap-lru:jar:1.0_jdk5 for project com.googlecode.concurrentlinkedhashmap:concurrentlinkedhashmap-lru:jar:1.0_jdk5
  com.googlecode.concurrentlinkedhashmap:concurrentlinkedhashmap-lru:jar:1.0_jdk5

from the specified remote repositories:
  jitsi-maven-repository-snapshots (https://github.com/jitsi/jitsi-maven-repository/raw/master/snapshots/),
  central (http://repo1.maven.org/maven2),
  sonatype-nexus-snapshots (https://oss.sonatype.org/content/repositories/snapshots),
  jitsi-maven-repository-releases (https://github.com/jitsi/jitsi-maven-repository/raw/master/releases/)

Path to dependency:
        1) org.jitsi:jicofo:jar:1.0-SNAPSHOT
        2) org.igniterealtime:tinder:jar:1.2.3

I am running Ubuntu Server 14.04. Did I miss something obvious?

Thanks in advance

lyubomir:
Maven should be preferred to Ant for Jicofo.

[...]

PR #387, opened 2019-09-19
Adding instructions for building the package from source

dmikushin:
It took me some time to get through https://github.com/jitsi/jicofo/issues/30 and figure out what steps are needed for successful command-line compilation. This PR proposes to add at least minimal directions into the README to simplify the developer experience, especially for those who are not native Java-speakers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -180,3 +180,25 @@ configured with the jitsi-meet scripts, then you can find the certificate in:
 /var/lib/prosody/$JICOFO_AUTH_DOMAIN.crt 
 ```
 
+## Building from source
+
+The build system is based on `maven` and `ant`. Additionally, `maven-ant-tasks` package is needed:
+
+```
+mvn dependency:get -DartifactId=maven-ant-tasks -DgroupId=org.apache.maven -Dversion=2.1.3
+```
+
+Goto the `jicofo` source folder and resolve package dependencies:
+
+```
+mvn dependency:resolve
+```
+
+Finally, build package for the given target platform, e.g. for 64-bit Linux:
+
+```
+ant -lib ~/.m2/repository/org/apache/maven/maven-ant-tasks/2.1.3 dist.lin64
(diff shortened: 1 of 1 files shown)


damencho:
For having just the jicofo.jar, why not just use `mvn install` the default command for every maven project ... ?

damencho:
Ant is just used for creating debian packages, the rest is in maven.

bgrozev:
I would also suggest using `mvn install -Dassembly.skipAssembly=false` to generate the zipped packages (which can be used manually as a replacement for `/usr/share/jicofo` in a debian installation).

dmikushin:
Thank you for your feedback! I've used ant specifically in order to make a
local build for development purposes. That is, jicofo.jar is generated in
the build folder and can be manually copied into existing Jitsi
installation, without any packaging. Do you mean that mvn install can also
be used for this?

ср, 25 сент. 2019 г. в 13:16, bgrozev <notifications@github.com>:

bgrozev:
Yes, we recommend always using maven instead of ant. Just running `mvn install` should be enough to produce a new `jicofo.jar`. However, replacing `jicofo.jar` alone may not be sufficient because some of the dependencies may have changed (especially if you are working on some of the underlying libraries instead).

My procedure for manually building a custom version and deploying it to an existing remote installation is the following:
```
local$ mvn install -Dassembly.skipAssembly=false
local$ scp target/jicofo-linux-x64-1.1-SNAPSHOT.zip $remote

remote$ unzip jicofo-linux-x64-1.1-SNAPSHOT.zip
remote$ sudo mv /usr/share/jicofo /usr/share/jicofo-orig
remote$ sudo ln -s $HOME/jicofo-linux-x64-1.1-SNAPSHOT /usr/share/jicofo
remote$ sudo service jicofo restart
```

bgrozev:

PR #394, opened 2019-10-15
Bridge selection max load

bbaldino:
allow a configurable 'max participant count' for bridge selection such that no bridge with that many participants will be considered for selection.  i did this as a decorator, so it's totally orthogonal to the strategy itself, it just wraps whatever strategy is configured.

### DIFF
--- a/src/main/java/org/jitsi/jicofo/Bridge.java
+++ b/src/main/java/org/jitsi/jicofo/Bridge.java
@@ -294,7 +294,7 @@ public class Bridge
         return this.lastReportedBitrateKbps - o.lastReportedBitrateKbps;
     }
 
-    private int getEstimatedVideoStreamCount()
+    public int getEstimatedVideoStreamCount()
     {
         return videoStreamCount + videoStreamCountDiff;
     }
--- a/src/main/java/org/jitsi/jicofo/BridgeSelector.java
+++ b/src/main/java/org/jitsi/jicofo/BridgeSelector.java
@@ -1007,4 +1007,77 @@ public class BridgeSelector
                 .orElse(selectFrom.get(0));
         }
     }
+
+    /**
+     * Implements a {@link BridgeSelectionStrategy} which
+     * tries to find a bridge in the users region with a load
+     * of less than the number of participants defined by
+     * {@link RegionBasedMaxLoadBridgeSelectionStrategy#MAX_NUM_BRIDGE_PARTICIPANTS_PNAME}
+     * participants
+     */
+    private static class RegionBasedMaxLoadBridgeSelectionStrategy
+        extends RegionBasedBridgeSelectionStrategy
+    {
+        /**
+         * The name of the property to define the amount of participants
+         * after which a bridge will not be considered
+         */
+        public static final String MAX_NUM_BRIDGE_PARTICIPANTS_PNAME
+            = "org.jitsi.jicofo.BridgeSelector.MAX_NUM_BRIDGE_PARTICIPANTS";
+
+        /**
+         * The default number of participants after which the bridge
+         * will not be considered for adding a participant
+         */
+        private static int DEFAULT_MAX_NUM_BRIDGE_PARTICIPANTS = 25;
+
+        protected final int maxNumBridgeParticipants;
+
+        /**
+         * Default constructor.
+         */
+        RegionBasedMaxLoadBridgeSelectionStrategy()
+        {
+            ConfigurationService config = FocusBundleActivator.getConfigService();
+            if (config != null)
+            {
+                maxNumBridgeParticipants = config.getInt(
+                    MAX_NUM_BRIDGE_PARTICIPANTS_PNAME, DEFAULT_MAX_NUM_BRIDGE_PARTICIPANTS);
+            }
+            else
+            {
+                maxNumBridgeParticipants = DEFAULT_MAX_NUM_BRIDGE_PARTICIPANTS;
+            }
+        }
+
+        /**
+         * {@inheritDoc}
+         * </p>
+         * Always selects the bridge already used by the conference.
+         */
+        @Override
+        public Bridge doSelect(
+            List<Bridge> bridges,
+            List<Bridge> conferenceBridges,
+            JitsiMeetConference conference,
+            String participantRegion)
+        {
+            List<Bridge> filteredBridges = bridges.stream()
+                .filter(b -> b.getEstimatedVideoStreamCount() < maxNumBridgeParticipants)
+                .collect(Collectors.toList());
+
+            List<Bridge> filteredConferenceBridges = conferenceBridges.stream()
+                .filter(b -> b.getEstimatedVideoStreamCount() < maxNumBridgeParticipants)
+                .collect(Collectors.toList());
+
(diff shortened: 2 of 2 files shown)


bgrozev:
The PR description mentions number of participants, but the code uses the number of "video streams". I think number of participants makes more sense.

I think the strategy should fallback to non-filtered values when all bridges are loaded higher than the threshold.

bbaldino:
Thinking about this more, I think it'd be even better to let the _videobridge_ define whether or not it is fully-loaded: this gives us the flexibility to have a bridge take into account things like enabled-debug features (which would add load) when determining whether or not it is fully-loaded.

As for this "fallback when a bridge says it's loaded thing", I still think we're looking at this the wrong way.  If 'fully-loaded' is defined as "we believe this bridge has reached a reasonable limit to handle data", then we should *not* add more to that bridge.  The solution here shouldn't be to just say "oh well, throw more load on the bridge", it should be to make sure we never get into a position where we don't have bridges for participants: we should be anticipating this and scaling up new bridge appropriately.  Do we really feel like it's impossible to anticipate and scale up in such a way that we'll be causing experience issues for users?  I think the discussion should focus around _that_ problem--at least at first--before we fall back to just piling more load on a bridge we've already deemed "full".

damencho:
But there should always be a fallback. If you prefer not to load more the bridges, we need to implement proper signalling that jicofo will send back to client and the ui should show the appropriate message that we are full and to offer the user to retry in few minutes. It is always possible that something can go wrong...

bbaldino:
> But there should always be a fallback. If you prefer not to load more the bridges, we need to implement proper signalling that jicofo will send back to client and the ui should show the appropriate message that we are full and to offer the user to retry in few minutes. It is always possible that something can go wrong...

Sure, I don't think I've said anything against this sort of thing.

bgrozev:
> The solution here shouldn't be to just say "oh well, throw more load on the bridge", it should be to make sure we never get into a position where we don't have bridges for participants

We already do this elsewhere (aws autoscaling rules based on system load). We can never bring everything here in the github projects, because spinning up new machines is highly specific to the deployment. So the advantage of having more of this logic in this project is limited. I.e., you'll always need additional logic, so you might as well use system load.

JonathanLennox:
It sounds to me like bridges should have both a "yellow" state (I'm getting loaded, give load to green bridges if they're available, or spin up new bridges) and also a "red" state (don't give me any more load at all).

bbaldino:
> > The solution here shouldn't be to just say "oh well, throw more load on the bridge", it should be to make sure we never get into a position where we don't have bridges for participants
> 
> We already do this elsewhere (aws autoscaling rules based on system load). We can never bring everything here in the github projects, because spinning up new machines is highly specific to the deployment. So the advantage of having more of this logic in this project is limited. I.e., you'll always need additional logic, so you might as well use system load.

I don't follow; I'm not proposing this logic move anywhere?  There will be something monitoring the load of the bridges (be it some `system_load` stat or some custom metric the bridge provides) and potentially scaling up new bridges based on that.  But, separately, Jicofo will need to read some metric to decide which bridge to select.

bgrozev:
> I don't follow; I'm not proposing this logic move anywhere?

You said you think we should focus on the problem of making sure we scale up appropriately and always have enough bridges. That's a problem we solve elsewhere, and I think with the latest improvements it works reasonably well.

The question I brought up is what to do when this fails. The code in the PR just throws an error, which is not handled well in the other components. This is new behavior. The current behavior is to select the least loaded bridge (even if it is loaded above the threshold), and I think this is reasonable since we have a margin, and new bridges will be spun up soon.

bbaldino:
> > I don't follow; I'm not proposing this logic move anywhere?
> 
> You said you think we should focus on the problem of making sure we scale up appropriately and always have enough bridges. That's a problem we solve elsewhere, and I think with the latest improvements it works reasonably well.
> 
> The question I brought up is what to do when this fails. The code in the PR just throws an error, which is not handled well in the other components. This is new behavior. The current behavior is to select the least loaded bridge (even if it is loaded above the threshold), and I think this is reasonable since we have a margin, and new bridges will be spun up soon.

If I'm understanding this right, this policy is: "choose the least-loaded bridge" (in addition to geography or whatever other policy we may be applying).  We can do this, but we should be conscious that the decision we're making is: in the scenario where scale-up fails to happen, rather than give an error to a new participant we'll--with some probability--negatively affect the experience of all users on the bridge we decide to put the new participant on.

bgrozev:
Yes, which is what we do today.

bbaldino:
> Yes, which is what we do today.

And we're happy with this? Degrading experience for many users to give a degraded experience for another user?

damencho:
> And we're happy with this? Degrading experience for many users to give a degraded experience for another user?

But why you think its that case? We normally do not put the bar on the limit, we always keep it a little bit lower just in case. This gives us this time waiting for the new ones to spin up and to take the load.
And that's why these are configurable so we can move the bar when we think we need to.

bbaldino:
> > And we're happy with this? Degrading experience for many users to give a degraded experience for another user?
> 
> But why you think its that case? We normally do not put the bar on the limit, we always keep it a little bit lower just in case. This gives us this time waiting for the new ones to spin up and to take the load.
> And that's why these are configurable so we can move the bar when we think we need to.

Yes but this is true regardless.  The limit at which the auto-scaling logic decides to spin up new bridges is totally independent of the limit at which Jicofo would decide whether or not to put a participant on a bridge.  

For example, auto-scaling might scale up new bridges if no bridge had a "load" of less than `70`, but Jicofo would still put a participant on a bridge of load `70`.  If all bridges had a load of `100` though, adding another participant would degrade the experience for all those on that bridge, so Jicofo would not put a new participant here.  The only time all bridges have loads above 100 is if auto-scaling has failed to work as needed.

bgrozev:
There's a difference between the degraded experience when a bridge is temporarily overloaded, and the complete failure to join a conference. There is sufficient flow of users*, that after new bridges ar espun up an overloaded bridge will recover relatively quickly. 

*With 100 participants per bridge and an average conference duration of 20 min we shold see around 5 participants join/leave per minute.

bbaldino:
> There's a difference between the degraded experience when a bridge is temporarily overloaded, and the complete failure to join a conference.

Ok, so can we agree that both a "soft" limit and a "hard" limit exist?  That's my main point here.

gpolitis:
There's room for improvement in this area and plenty of good ideas have already been mentioned by you all, so I find it's unfortunate that this topic has stalled. I merely wanted to share a couple of thoughts on the subject and maybe revitalise interest.

I like Brian's decorator approach and agree with him that system load is a more appropriate indicator than the number of streams (or participants). I also agree with Boris and Damyan in that we should not change the existing functionality and introduce any hard limits. What we need, it seems, is a bridge prioritizer (nicely implemented by a decorator), after the main strategy has executed.

bgrozev:
> What we need, it seems, is a bridge prioritizer (nicely implemented by a decorator), after the main strategy has executed.

I agree.

Note that the current implementation uses the bitrate and not the number of streams or participants. This is precesely because it is much more closely related to system load.

PR #397, opened 2019-10-24
Add a /stats REST endpoint to Jicofo, expose some data

bbaldino:
This PR adds a `/stats` REST endpoint in Jicofo to expose conference count, largest conference size, and a conference sizes histogram.  Previously, this data was being retrieved from the bridge, but due to Octo the bridge is unable to give reliable information, as it doesn't have the complete picture.

I added Jersey here to handle this; I've already updated all the REST code in the bridge to use it and I think it'd make sense to port over the other Jicofo endpoints, but I refrained from doing that as part of this work--though I can take a look at doing that in an immediate follow-up PR for consistency.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -184,6 +184,36 @@
       <artifactId>ice4j</artifactId>
       <version>3.0-1-ge854e4e</version>
     </dependency>
+    <dependency>
+      <groupId>org.glassfish.jersey.containers</groupId>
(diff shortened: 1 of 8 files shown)


bbaldino:
@aaronkvanmeerten i haven't been able to verify this with an octo call yet as i don't have an easy way to test, i was thinking since this is all-new maybe we could just do that on beta?  let me know what you think.

PR #398, opened 2019-10-25
Port Jicofo's health to Jersey

bbaldino:
Since the new stats uses Jersey I thought I'd start porting over the other REST endpoints (health, version) to Jersey as well, so here's the `/about/health` endpoint and I added some tests as well.  I tried to just reproduce what already existed, but I think moving the 'health check' logic to a function in `FocusManager` to make the health endpoint class a bit dumber would be a good thing to revisit at some point.


src/main/java/org/jitsi/jicofo/auth/AuthBundleActivator.java
```diff
@@ -146,16 +146,16 @@ protected Handler initializeHandlerList(
     {
         List<Handler> handlers = new ArrayList<>();
 
-        // FIXME While Shibboleth is optional, the health checks of Jicofo (over
-        // REST) are mandatory at the time of this writing. Make the latter
-        // optional as well (in a way similar to Videobridge, for example).
-        handlers.add(new org.jitsi.jicofo.rest.HandlerImpl(bundleContext));
-
```
damencho: Why did you do this change? This breaks health checks when Shibboleth is enabled
https://github.com/jitsi/jicofo/pull/367

bbaldino: this didn't actually get touched, the diff just looks like this because it thinks it got 'moved' (see a little below on the right side).  is that what you mean?

damencho: Yes, that move breaks health checks when Shibboleth is enabled. Look at the PR I pasted.

bbaldino: i don't follow, from what i read there it's important that the `/about/health` handler is above shibboleth, and that's still the case.  this change is just making it so the new jersey code runs before the old rest handler code, since `AbstractJSONHandler` still tries to handle `/about/health`.  once version gets ported as well the `HandlerImpl` will go away entirely.

damencho: I'm also confused cause your change is reverting that PR. 
So you say the requests are now split and don't follow the same path as before so the change in that PR is not relevant with your changes today?

bbaldino: The change in that PR moves the handler for `/about/health` to before adding the Shibboleth handler, which is still the case here.  After #367, but before the jersey changes, it went:

1) Add HandlerImpl (which handles `/about/health`, `/about/version`, etc.
2) Add Shibboleth handler (if enabled)

After this PR, it goes:

1) Add the `Application` handler, which handles `stats` and `health`
2) Add HandlerImpl, which still handles `version`
3) Add Shibboleth handler, if enabled

eventually version will get ported to `Application` as well, and the old `HandlerImpl` will go away, and it will just be:

1) Add the `Application` handler, which handles `stats` and `health`
2) Add Shibboleth handler, if enabled

damencho: Ok, thanks, for the explanation. I see it now.

damencho: @bbaldino Can you take a look again, cause Shibboleth is broken, as I initially suspected. Not completely sure whether it is this change or some other around the jetty, but the fact is that those changes broke it. I tested it, and you got either Shibboleth or healthcheck/version.
https://github.com/jitsi/jicofo/issues/404

PR #402, opened 2019-11-08
Track num participants/total num participants

bbaldino:
(This is built on top of the code in https://github.com/jitsi/jicofo/pull/401, so the first 3 commits here will go away once that is merged and I rebase).

This adds tracking for the current number of participants and the total number of participants since start.  I don't know if `FocusManager` is the best place to be the central location for stats, but it seemed reasonable so let me know if somewhere else would be appropriate.

### DIFF
--- a/src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
+++ b/src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
@@ -3034,7 +3034,7 @@ public class JitsiMeetConferenceImpl
         return includeInStatistics;
     }
 
-    public FocusManager getFocusManager()
+    protected FocusManager getFocusManager()
     {
         return ServiceUtils.getService(FocusBundleActivator.bundleContext, FocusManager.class);
     }
--- a/src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
+++ b/src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
@@ -661,6 +661,7 @@ public class JitsiMeetConferenceImpl
             logger.info(
                     "Member "
                         + chatRoomMember.getContactAddress() + " joined.");
+            getFocusManager().getStatistics().totalParticipants.incrementAndGet();
 
             if (!isFocusMember(chatRoomMember))
             {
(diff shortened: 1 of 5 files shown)


bbaldino:
cc @aaronkvanmeerten

src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
```diff
@@ -3033,6 +3034,11 @@ public boolean includeInStatistics()
         return includeInStatistics;
     }
 
+    public FocusManager getFocusManager()
+    {
+        return ServiceUtils.getService(FocusBundleActivator.bundleContext, FocusManager.class);
```
paweldomas: would it be a better place for this to be a `static` getter in the `FocusManager` class?

src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
```diff
@@ -661,6 +661,7 @@ protected void onMemberJoined(final XmppChatMember chatRoomMember)
             logger.info(
                     "Member "
                         + chatRoomMember.getContactAddress() + " joined.");
+            getFocusManager().getStatistics().totalParticipants.incrementAndGet();
```
paweldomas: would it be better to pass the statistics instance in the constructor instead?

src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
paweldomas: I'm deleting this comment. I think it's better to pass instance instead of reaching to a global

src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
bbaldino: I'm not sure.  On one hand I'd say yes, but in general I think Jicofo embraces the osgi approach in favor of passing a bunch of services in to the constructor, so this felt more consistent.  If you think it'd be better as a constructor param I'll change it, I don't feel too strongly about it.

paweldomas: I don't like OSGi and those static globals, but it doesn't matter, so I guess it's fine(actually checked and there are bunch of similar usages there already). Maybe at least not make it public or have it moved to the FocusManager class? It would be weird to start seeing other classes using this one to get FocusManager and it encourages just that.

bbaldino: Ah, yes that was a mistake that it was public.  Thanks.

Issue #404, opened 2019-11-13
Problems connecting to Shibboleth


lobosco:
Hi, 

Since last week I am experiencing problems connecting jitsi to a SAML IdP (e.g. Shibboleth).

## Description
I am running jitsi on a Ubuntu 16.04 machine and added support for Shibboleth according to the official guide (https://github.com/jitsi/jicofo/blob/master/doc/shibboleth.md). Until last week, everything was working fine. Unfortunately, on any version of jicofo other than 1.0-485-1 (unstable) this is no longer the case 

## Current behavior
Whenever I try to open a conference, I receive a 'something went wrong' error, instead of the usual relaying to the Identity Provider and than back again to the conference. 

This error is unfortunately not logged. Even with log level 'FINEST' I only receive the following lines in jicofo.log: 

`WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by org.dom4j.io.SAXContentHandler (file:/usr/share/jicofo/lib/dom4j-1.6.1.jar) to method com.sun.org.apache.xerces.internal.parsers.AbstractSAXParser$LocatorProxy.getEncoding()
WARNING: Please consider reporting this to the maintainers of org.dom4j.io.SAXContentHandler
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release`

Any help would be very appreciated, since I can unfortunately no longer simply revert to the known working version of 1.0-485-1.

Regards, 

Wolfgang Knopki, Stuttgart University of Applied Sciences

damencho:
What is the js error you see, probably an xmpp error package, expand it and what is the error? 
Which are the versions you had tried, latest from unstable doesn't work?

lobosco:
Thank you for your quick reply.

I have tried the latest version from unstable, as well as versions 1.0-504-1, *506*, *501*, *488* and versions 1.0-257-1, 1.0-268-1 and 1.0-181-1 from stable. I am currently preparing to test the latest versions from git.

The errors I get in js are the following: 

`2019-11-13T12:09:47.471Z [JitsiMeetJS.js] <getGlobalOnErrorHandler>:  UnhandledError: null Script: null Line: null Column: null StackTrace:  Error: "Focus error, retry after 4000"
    _allocateConferenceFocusError moderator.js:451
    allocateConferenceFocus moderator.js:378
    s strophe.js:3388
    run strophe.js:2556
    _dataRecv strophe.js:3822
[...]

PR #408, opened 2019-11-27
fix the shibboleth web handler

bbaldino:
since the servlet has a path spec of '/*', it matches all urls and, if
it doesn't end up handling a request, that request is still marked as
'handled' so it doesn't fall through to the next handler.

now, we'll move the shibboleth handler first but make sure it handles
only the shibboleth urls: all others will not be marked as handled and
will fall through to the next handler.

addresses https://github.com/jitsi/jicofo/issues/404

### DIFF
--- a/src/main/java/org/jitsi/jicofo/auth/ShibbolethHandler.java
+++ b/src/main/java/org/jitsi/jicofo/auth/ShibbolethHandler.java
@@ -51,9 +51,13 @@ class ShibbolethHandler
     private static final Logger logger
             = Logger.getLogger(ShibbolethHandler.class);
 
+    /**
+     * The target on which shibboleth requests are made
+     */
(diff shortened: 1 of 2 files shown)


src/main/java/org/jitsi/jicofo/auth/ShibbolethHandler.java
```diff
@@ -53,6 +53,8 @@
 
     private final ShibbolethAuthAuthority shibbolethAuthAuthority;
 
+    private static final String SHIBBOLETH_TARGET = "/login";
```
paweldomas: I feel bad for asking about it, but can you move it above the `shibbolethAuthAuthority` so that `static final` are grouped together?

bbaldino: sure, added a javadoc as well.

PR #409, opened 2019-11-29
Fix health returned jvb list

damencho:
Before the change health check that returns jvb list (curl -v http://localhost:8888/about/health?list_jvb=true), in case of components returns:
`The JID 'jitsi-videobridge.jitsi.example.com' has no resourcepart (through reference chain: org.json.simple.JSONObject["jvbs"]->java.util.ArrayList[0]->org.jxmpp.jid.impl.DomainpartJid["resourceOrThrow"])`.
And in case of mucs returns:
`{"jvbs":[{"resourceOrNull":"damencho","localpartOrNull":"jvbbrewery","resourcepart":"damencho","localpart":"jvbbrewery","domain":"internal.auth.jitsi.example.com","domainBareJid":false,"entityBareJid":false,"entityJid":true,"domainFullJid":false,"entityFullJid":true,"resourceOrEmpty":"damencho","resourceOrThrow":"damencho","localpartOrThrow":"jvbbrewery"}]}`

After the change will return:
`{"jvbs":["jvbbrewery@internal.auth.jitsi.example.com/damencho"]}`
or
`{"jvbs":["jitsi-videobridge.jitsi.example.com"]}`

Reporrted by Bui_Xuan_Thu_n https://community.jitsi.org/t/jicofo-lastest-version-jicofo-1-0-508-1-i386-deb-error-api-about-health

### DIFF
--- a/src/main/java/org/jitsi/jicofo/auth/AuthBundleActivator.java
+++ b/src/main/java/org/jitsi/jicofo/auth/AuthBundleActivator.java
@@ -159,11 +159,12 @@ public class AuthBundleActivator
         // FIXME While Shibboleth is optional, the health checks of Jicofo (over
         // REST) are mandatory at the time of this writing. Make the latter
         // optional as well (in a way similar to Videobridge, for example).
-        ServletContextHandler appHandler = new ServletContextHandler(ServletContextHandler.NO_SESSIONS);
+        ServletContextHandler appHandler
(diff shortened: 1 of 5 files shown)


bbaldino:
lgtm, but the tests need to be updated as well.

damencho:
> lgtm, but the tests need to be updated as well.

Fixed the test which revealed another bug. Oh I need to rebase, will force push.

PR #411, opened 2019-12-03
fix: Distinguishes between actual bad-requests and "conference ID not found" errors from the bridge.

bgrozev:


src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
```diff
@@ -2358,8 +2358,20 @@ private void onBridgeUp(Jid bridgeJid)
...
     /**
-     * Handles on bridge down event by shutting down the conference if it's the
-     * one we're using here.
+     * Notifies this conference that one of its channel allocators failed to
+     * allocate channels, and that the participants on a specific bridge need
+     * to be re-invited.
+     * @param bridgeJid the JID of the bridge on which participants need to be
+     * re-invited.
+     */
+    void channelAllocationFailed(Jid bridgeJid)
```
bbaldino: is this called anywhere?

bgrozev: It is now (I just fixed it)

bbaldino: i still don't see it?

bgrozev: It's here: https://github.com/jitsi/jicofo/pull/411/commits/a57d14ed382fb902ea94d0b8acdecd89579304fc#diff-653d615bc7b6b3b1f8196be7f8491c8eR313

PR #414, opened 2020-01-14
fix: validate conference state on source removal

paweldomas:
When processing client's source remove notification, first tries to remove on conference state copy and prevents the modification if it results in a broken state.

### DIFF
--- a/src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
+++ b/src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
@@ -1865,10 +1865,8 @@ public class JitsiMeetConferenceImpl
         MediaSourceGroupMap sourceGroupsToRemove
             = MediaSourceGroupMap.getSourceGroupsForContents(contents);
 
-        removeSources(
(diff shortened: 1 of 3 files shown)


bbaldino:
What was the scenario that was causing a bug?

paweldomas:
This one: https://github.com/jitsi/jicofo/pull/414/commits/ab0b97ce190c6b8adc02f7e273281974e5a46514#diff-e3805cc088d661a2b852885acfeff77cR719

I haven't figured out exactly why the client was doing that, but it removed main video SSRC, SIM and one FIR group. Probably when the client would try too replace video tracks in one go and the main source changed? I think it then followed with source add for the new main SSRC.

That weird client state was caused by blocked GUM calls not handled correctly for presenter mode: https://github.com/jitsi/jitsi-meet/pull/4986
While I was able to repro the errors around the presenter feature I was not able to repro that weird tracks state. I only hope client will be fine without going into invalid state. This PR should prevent from breaking entire conference in the future.

bgrozev:
lgtm

paweldomas:
@bgrozev can I get a stamp?

PR #418, opened 2020-01-24
Takes the number of estimated video streams into consideration when s…

gpolitis:
…electing a bridge.

The goal is to determine whether the bridge is in survival mode and ideally
this information comes directly from the bridge in question. This is
just a first implementation based on the number of estimated video
streams.


src/main/java/org/jitsi/jicofo/BridgeSelector.java
```diff
@@ -1002,9 +1017,20 @@ private Bridge findFirst(
...
+                    .filter(not(Bridge::isInSurvivalMode))
+                    .findFirst()
+                    .orElse(order.stream()
+                        .filter(selectFrom::contains)
+                        .findFirst()
+                        .orElse(selectFrom.get(0))));
         }
     }
+
+    public static <T> Predicate<T> not(Predicate<T> t) {
```
bbaldino: This is cool!

src/main/java/org/jitsi/jicofo/BridgeSelector.java
```diff
@@ -767,22 +773,28 @@ private Bridge selectInitial(List<Bridge> bridges,
         {
             Bridge bridge = null;
 
-            // Prefer a bridge in the participant's region.
-            if (participantRegion != null)
+            for (boolean includeBridgesInSurvivalMode : FALSE_TRUE)
             {
-                bridge = findFirstOperationalInRegion(bridges, participantRegion);
-            }
+                // Prefer a bridge in the participant's region.
```
bbaldino: How about breaking out of the loop when we find a bridge?

src/main/java/org/jitsi/jicofo/BridgeSelector.java
gpolitis: more like legacy code if you think that Java11 already has this baked in :/

bgrozev:
The bridges are already ordered by load (see `BridgeSelector#selectBridge`), so we are already  selecting the least loaded bridge when there are multiple options. I found a bug in the existing code while looking at the PR, and while fixing it I'll also rename the variables to make the order clear (conflicts should be easy to fix).

I like the idea of an `overloaded` flag coming from the bridge, but I think adding a second load metric to jicofo is problematic. If we have to estimate load, let's do it on the existing metric of bitrate. We already know that the "videostreams" metric is NOT a good indication of load in practice:
Screenshot 2020-01-27 at 11 30 59


What is the problem you are trying to solve? Is it that jicofo's `bitrate` stats lag behind because the bridge only sends updates at 5 second intervals? If so, note that:
1. This is mostly a problem in a testing environment where a large number of clients join simultaneously. In live environments participants don't join in huge bursts.
2. Having an `overloaded` flag come from the bridge will not solve the problem, because it will also lag (so this code is not a step towards the proper solution, but only a patch)
3. Jicofo's correction of the "videostreams" metric is not accurate when a bridge is connected to more than one jicofo.

Perhaps increasing the bridges' reporting rate and adding a small interval between clients joining will be a better solution for performance testing.

paweldomas:
> This is mostly a problem in a testing environment where a large number of clients join simultaneously. In live environments participants don't join in huge bursts.

This is true, unless one of the bridges goes bad and the conferences are  being migrated. That was the use case that the video stream estimation was trying to solve.

gpolitis:
Thanks to both for your comments. @bgrozev the initial motivation for this PR is what we saw last week together when we were looking at the datadog graphs, i.e. the load isn't distributed evenly and a single bridge may end up taking most of the load produced in a large call.

If there's a large call scheduled, it may happen that many participants join "simultaneously" (or within a very short period of time, i.e. < 30 seconds). Malleus creates participants faster than that, so maybe I need to tweak it to spread participants across X seconds and test again? @bgrozev Do you expect to see better load distribution in this new test?

Then of course we have the case that @paweldomas is describing that isn't currently handled well.

bgrozev:
> hanks to both for your comments. @bgrozev the initial motivation for this PR is what we saw last week together when we were looking at the datadog graphs, i.e. the load isn't distributed evenly and a single bridge may end up taking most of the load produced in a large call.

Which selection strategy was used during the test? I thought it was `SplitBridgeSelectionStrategy`, which would chose a bridge at random (and not round-robin).



> If there's a large call scheduled, it may happen that many participants join "simultaneously" (or within a very short period of time, i.e. < 30 seconds).

The bridges' reporting interval is just 5 seconds, I think that's low enough even for scheduled meetings.


> Malleus creates participants faster than that, so maybe I need to tweak it to spread participants across X seconds and test again? @bgrozev Do you expect to see better load distribution in this new test?

Yes, I think this together with lowering the reporting interval should help with the test.



> Then of course we have the case that @paweldomas is describing that isn't currently handled well.

I agree this is a problem that needs to be addressed.

gpolitis:
I rebased my changes over @bgrozev recent fixes and now the PR looks quite different. Do you guys (@bbaldino @bgrozev @paweldomas) prefer I force push or should I close this one and open a new PR?

bbaldino:
> I rebased my changes over @bgrozev recent fixes and now the PR looks quite different. Do you guys (@bbaldino @bgrozev @paweldomas) prefer I force push or should I close this one and open a new PR?

force push is fine with me.

bgrozev:
The only reason "videostreams" is useful is because we keep track of the diff locally. The absolute number is not a good indicator of load, so the thresholds don't have any meaning. It would be nice to capture the concept of "load that we recently added to this bridge" without the noise of the absolute number of "videostreams".

How about this:
We define a new load metric as `load = bitrate + [videoStreamCountDiff](https://github.com/jitsi/jicofo/blob/master/src/main/java/org/jitsi/jicofo/bridge/Bridge.java#L79) * C` (C estimates the average bitrate of a video stream). 

This metric updates immediately when we add new participants, and is easier to reason about because its a single value expressed in bps. We order the bridges by `load`. Your stress thresholds consider only `load`.

We can then finally remove "videostreams" from the bridge (since we only care about the diff), and we can improve the estimate by e.g. taking simulcast into account.

bbaldino:
Can we have the bridge itself just return how 'stressed' it is?  That way we can have all sorts of things impact that decision on the bridge without having to involve Jicofo at all.  For example, maybe we turn on a debug mode that has overhead but wouldn't be reflected at all by counting the number of participants/streams/etc.

bgrozev:
> Can we have the bridge itself just return how 'stressed/loaded' it is?

This doesn't address the problem of updates from the bridge coming rarely, leading to jicofo dumping everyone on the momentarily lowest loaded bridge when a large number of participants join together.

bbaldino:
> > Can we have the bridge itself just return how 'stressed/loaded' it is?
> 
> This doesn't address the problem of updates from the bridge coming rarely, leading to jicofo dumping everyone on the momentarily lowest loaded bridge when a large number of participants join together.

Bridges could push updates whenever their states transitioned rather than at some interval?  I think there'd be solutions to that.  I'm just trying to see if we can keep Jicofo dumb here and not need to be changed every time we want to evolve this calculation on the bridge, or, perhaps, support different numbers for different bridge versions?

bgrozev:
> Bridges could push updates whenever their states transitioned rather than at some interval?

Not sure if that's good enough for the case where a bridge failed and jicofo transfers all of its participants to other bridges.

What if the bridge publushes its thresholds in terms of bitrate?

bbaldino:
> > Bridges could push updates whenever their states transitioned rather than at some interval?
> 
> Not sure if that's good enough for the case where a bridge failed and jicofo transfers all of its participants to other bridges.
> 

I almost mentioned here that we could still do some minimum interval, but thought it wouldn't be needed.  Looks like this case would require it, though, so we still have it transmit at least as often as before?

gpolitis:
> Why not use an enum?

@bbaldino I imagined the stress level reported by the bridge to be a float (not necessarily the system load value, but an aggregate of system and network load: for example the average or the max of system load and network load) so defining the constants as integers allowed me to use the less-than operator (i.e. consider the bridges that are less than MEDIUM_HIGH loaded). But you're right, we should simplify this, the bridge can simply report an enum value?

> The only reason "videostreams" is useful is because we keep track of the diff locally. 

@bgrozev in fact I think both are useless right now, at least I don't see any other use other uses than in this new load distribution scheme, or maybe I'm wrong?

> The absolute number is not a good indicator of load, so the thresholds don't have any meaning.

@bgrozev true, because only the bridge knows its stress level, as @bbaldino has pointed out. His suggestion to have the bridge push its stress level to Jicofo can also simplily things when in the future we move to a different instance type for the bridge or if we ever decide to have a mixed instance type environment. 

>  Not sure if that's good enough for the case where a bridge failed and jicofo transfers all of its participants to other bridges.

@bgrozev that's a good point, but just to be clear, this PR doesn't attempt to solve that problem. I assume that 100s of participants join a large call within 30 seconds. As we've discussed last week, the bridge-crash problem exists even outside the context of large calls (which I'm frantically working on and trying to get done). Of course it would be fascinating if we could come up with a solution that addresses the bridge-crash issue, but I think @bbaldino's suggestion to have bridges push updates whenever their states transitioned rather than at some interval would work well here.

> What if the bridge publushes its thresholds in terms of bitrate?

@bgrozev I'm not sure if it's a good idea to use bitrate as the only measure of stress or to make Jicofo smart about this as it doesn't have to care about why a bridge is stressed, just its stress level.

paweldomas:
About the issue of many candidates joining or being migrated in very short time. Whatever  metric for the stress level you guys come up with here, it can be returned in the channel allocation response. It's still not as accurate as being able to calculate in advance on the time when Jicofo queues the request, but maybe a bit faster than sending periodical updates?

bgrozev:
> @bgrozev in fact I think both are useless right now, at least I don't see any other use other usages outside this new load distribution scheme, or maybe I'm wrong?

Useful in the sense that "it can be used for something". Bitrate is useful because it correlates with load. Videostreams is useful, because we keep a local estimate when we allocate channels, avoiding the lag present for the rest of the stats (and I claim that's the only reason it's useful).

> @bgrozev true, because only the bridge knows its stress level

No, this is true regardless of whether it is the bridge or another entity that looks at the videostreams. The thresholds on videostreams don't have any meaning because the absolute number of videostreams does not correlate well with load. Depending on a bunch of variables (simulcast use, p2p, muted video, etc) the exact same machine with 1250 videostreams may be overloaded or very lightly loaded. Look at this graph again: we balance the videostreams, but the load is NOT balanced. 
Screenshot 2020-01-27 at 11 30 59


> @bgrozev that's a good point, but just to be clear, this PR doesn't attempt to solve that problem

Then I am uterrly confused why you chose to use videostreams. 


You didn't adress my actual suggestion at all. Is there a case where it fails?

bgrozev:
> Whatever metric for the stress level you guys come up with here, it can be returned in the channel allocation response.

That's a good idea, but I don't think it will solve the bridge-crash problem, because the load on the bridge itself will not update immediately.

paweldomas:
>> Whatever metric for the stress level you guys come up with here, it can be returned in the channel allocation response.

>That's a good idea, but I don't think it will solve the bridge-crash problem, because the load on the bridge itself will not update immediately.

Then maybe something around random round robin is better on the jicofo side when it knows it's allocating a lot

bbaldino:
> > Whatever metric for the stress level you guys come up with here, it can be returned in the channel allocation response.
> 
> That's a good idea, but I don't think it will solve the bridge-crash problem, because the load on the bridge itself will not update immediately.

Are you talking about the scenario when bridge A fails and Jicofo moves everyone to a new bridge?  And are you saying that Jicofo can be smarter in this situation because it knows how many endpoints are about to be allocated vs the bridge who doesn't?

paweldomas:
> Are you talking about the scenario when bridge A fails and Jicofo moves everyone to a new bridge? And are you saying that Jicofo can be smarter in this situation because it knows how many endpoints are about to be allocated vs the bridge who doesn't?

yes and yes

gpolitis:
Depends on https://github.com/jitsi/jitsi-xmpp-extensions/pull/20

src/main/java/org/jitsi/jicofo/bridge/Bridge.java
```diff
@@ -50,6 +50,11 @@
     private static final ColibriStatsExtension EMPTY_STATS
         = new ColibriStatsExtension();
 
+    /**
+     *
+     */
```
bbaldino: Are you going to add a doc here?

src/main/java/org/jitsi/jicofo/bridge/Bridge.java
```diff
@@ -361,99 +349,14 @@ public String getRegion()
...
-     * know that it misbehaves (i.e. it drops packets and/or introduces jiitter),
-     * and therefore jicofo needs to route traffic to less stressed bridges.
-     *
-     * Having multiple stress levels allows us to act before a bridge has become
-     * highly stressed.
-     *
-     * @return the "stress" level of the bridge.
-     */
-    public float getStressLevel()
+    public double getStress()
```
bbaldino: Docs here would be very useful as well

src/main/java/org/jitsi/jicofo/bridge/BridgeSelectionStrategy.java
```diff
@@ -85,39 +86,55 @@ public Bridge doSelect(List<Bridge> bridges,
...
+
         Bridge bridge = null;
-        for (float stressLevel : StressLevels.ALL_LEVELS)
+        List<Double> bridgesStress = bridges
+            .stream()
+            .map(Bridge::getStress)
+            .sorted()
+            .collect(Collectors.toList());
+
+        // Try to group into groups of two, so #bridge/2
```
bbaldino: I don't follow this comment, what's being done here?

src/test/java/org/jitsi/jicofo/bridge/BridgeSelectionStrategyTest.java
```diff
@@ -191,17 +190,11 @@ public void preferRegionWhenStressLevelIsEqual()
             highStressRegion = region1;
 
         highStressBridge.setStats(
-            createJvbStats(StressLevels.HIGH, 100, highStressRegion));
-        assertEquals(
-            highStressBridge.getStressLevel(), StressLevels.HIGH, precision);
+            createJvbStats(750_000, 75_000, highStressRegion));
```
bbaldino: What are these numbers?

bbaldino: Same below

gpolitis: I've added some comments. Please let me know if it doesn't make sense.

src/main/java/org/jitsi/jicofo/bridge/Bridge.java
gpolitis: Added.

src/main/java/org/jitsi/jicofo/bridge/Bridge.java
gpolitis: Added.

src/main/java/org/jitsi/jicofo/bridge/RegionBasedBridgeSelectionStrategy.java
```diff
@@ -15,68 +27,4 @@
     RegionBasedBridgeSelectionStrategy()
```
bgrozev: This is now a no-op? Why not remove that class if that's the case?

src/main/java/org/jitsi/jicofo/bridge/BridgeSelectionStrategy.java
```diff
@@ -67,79 +69,120 @@ public Bridge select(
...
-            bridge = findFirstOperationalInRegion(bridges, localRegion);
-        }
+        double maxStress = findStressLimit(bridges);
 
-        // Otherwise, just find the first operational bridge.
-        if (bridge == null)
-        {
-            bridge = findFirstOperationalInRegion(bridges, null);
-        }
+        // Try the first operational bridge in the participant region and
```
bgrozev: "operational" has another meaning in jicofo. Maybe change it to "lightly loaded" or something that mentions stress?

src/main/java/org/jitsi/jicofo/bridge/BridgeSelectionStrategy.java
```diff
@@ -67,79 +69,120 @@ public Bridge select(
...
+                    .orElse(bridges.stream()
+                        .filter(stressIsLessThanOrEqual(maxStress))
+                        .filter(inRegion(localRegion))
+                        .findFirst()
+                        // Otherwise, try to add the least loaded bridge
+                        // that we know of.
+                        // TODO: perhaps use a bridge in a nearby region (if
+                        // we have data about the topology of the regions).
+                        .orElse(bridges.stream()
+                            .filter(stressIsLessThanOrEqual(maxStress))
```
bgrozev: This never considers bridges with `stress > maxStress`, failing instead. At least as a start we should preserve the current behavior and fallback to a stressed bridge instead of failing.

bgrozev:
The purpose of considering stress when selecting a bridge is to avoid
overloading it. However, making the wrong decision (incorrectly considering a
bridge overloaded) has negative impact:
1. Conferences are split, using more resources overall.
1. Participants are connected to a remote region.

This definition of `stress` is way too conservative and will result in false positives.
By only considering relative load, some of the bridges are *always* overloaded.

As an example consider the behavior when the system has just been initialized, with 2 bridges per region and all bridges have stress 0. Participants start joining from the same region.
The first one is sent to a bridge in their region (as it should be), which now
has load 500Kbps. This bridge is now overloaded. The second
participant is sent to a different bridge in the same region. Now both bridges in
this region are overloaded, so participants are sent to other regions randomly....this degenerates into random selection of regions and splits.


Having a pre-defined threshold for bitrate above which a bridge is overloaded
(the first solution we discussed last week) is not perfect, but it should work
reasonably well in practice. The second solution that we discussed last time
(the bridge reporting its own stress as `cpu + videoStreamDiff * C`) goes a
step further and makes it more generic.


A couple more thoughts:

1. It never makes sense to split a conference with just two participants in the
same region. We should handle this case separately in the code.

2. While the code is pretty simple, the overall behavior of the system is
complex, and we have no visibility into how it performs. Perhaps this is the time to add
some statistics to keep track of what decisions are made.

src/main/java/org/jitsi/jicofo/bridge/BridgeSelectionStrategy.java
gpolitis: this part has changed and we're now considering all bridges, even those that are very highly stressed

src/main/java/org/jitsi/jicofo/bridge/RegionBasedBridgeSelectionStrategy.java
gpolitis: I'm not sure we want this as a no-op; now the scheme is implemented in the base class and so it's inherited by the specialised classes, which may not be a good idea

gpolitis:
Thanks for your comments Boris, some inline answers:

> It never makes sense to split a conference with just two participants in the
same region. We should handle this case separately in the code.

Since we don't have 1-1 traffic through the bridge do we need to handle this case?

> While the code is pretty simple, the overall behavior of the system is
complex, and we have no visibility into how it performs. Perhaps this is the time to add
some statistics to keep track of what decisions are made.

I'm going to be working on that tomorrow, do you have any specific ideas in mind?

PR #420, opened 2020-01-29
Allow to disable vp8, vp9 and h264 with properties.

bgrozev:


JonathanLennox:
So you could disable either by setting the disable explicitly or by setting the pt to -1?

bgrozev:
> So you could disable either by setting the disable explicitly or by setting the pt to -1?

Yes. Unfortunately I had to do it this way in order to allow new infrastructure scripts to disable h264 in a way that doesn't break old jicofo versions...

PR #427, opened 2020-03-05
Enable mute requests from sip gateway.

theunafraid:


damencho:
Jenkins, add to whitelist, please!

damencho:
You want to put that behind flag, I suppose?

theunafraid:
> You want to put that behind flag, I suppose?

Added flag for mute request.

src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
```diff
@@ -2294,11 +2294,9 @@ boolean handleMuteRequest(Jid fromJid,
             return false;
         }
 
-        // we do not want to remotely mute jigasi as currently
-        // there is no way for it to unmute
-        if (doMute && participant.isSipGateway())
+        if (doMute && (globalConfig.isSipMuteEnabled() == false))
```
damencho: Tests are failing. There is an error in the logic I think ... Isn't this the correct if statement:
```
if (doMute && participant.isSipGateway() && !globalConfig.isSipMuteEnabled())
```
?

damencho: You are not checking this whether it is jigasi and you always forbid remote muting, which tests caught.

theunafraid: Corrected condition for jigasi

PR #428, opened 2020-03-06
Bridge selection improvements.

gpolitis:
This PR supersedes https://github.com/jitsi/jicofo/pull/418 and implements the changes discussed in the bridge selection document.

UPDATE: You can see how the behavior changes here https://github.com/jitsi/jicofo/pull/428/commits/cbfde03c4473082951cc2edc475309d7f498d341 as this commit adapts the test case (that captures the behavior) from the previous implementation.

### DIFF
--- a/src/main/java/org/jitsi/jicofo/bridge/Bridge.java
+++ b/src/main/java/org/jitsi/jicofo/bridge/Bridge.java
@@ -18,6 +18,7 @@
 package org.jitsi.jicofo.bridge;
 
 import org.jitsi.jicofo.util.*;
+import org.jitsi.utils.stats.*;
 import org.jitsi.xmpp.extensions.colibri.*;
 import static org.jitsi.xmpp.extensions.colibri.ColibriStatsExtension.*;
 
@@ -104,11 +105,9 @@ public class Bridge
     private final Jid jid;
 
     /**
-     * Estimates the number of new video streams added (or removed if negative)
-     * on this bridge by this jicofo instance since the last statistics update
-     * was received.
+     * Keep track of the recently allocated or removed channels.
      */
-    private int videoStreamCountDiff = 0;
+    private final RateStatistics videoChannelsRate = new RateStatistics(10000);
 
     /**
      * The last reported bitrate in Kbps.
@@ -156,11 +155,6 @@ public class Bridge
      */
     void setStats(ColibriStatsExtension stats)
     {
-        // Reset the counter for video streams added/removed since the last
-        // stats update. TODO: this, if used at all, needs to not be tied to
-        // stats.
-        videoStreamCountDiff = 0;
-
         if (stats == null)
         {
             this.stats = EMPTY_STATS;
@@ -304,15 +298,24 @@ public class Bridge
         return Double.compare(this.getStress(), o.getStress());
     }
 
-    void onVideoStreamsChanged(Integer videoStreamCount)
+    void onVideoChannelsChanged(Integer diff)
     {
-        if (videoStreamCount == null || videoStreamCount == 0)
+        if (diff == null)
         {
-            logger.error("videoStreamCount is " + videoStreamCount);
+            logger.error("diff is null");
             return;
         }
 
-        videoStreamCountDiff += videoStreamCount;
+        videoChannelsRate.update(diff, System.currentTimeMillis());
+    }
+
+    /**
+     * Returns the net number of video channels recently allocated or removed
+     * from this bridge.
+     */
+    private long getRecentVideoChannelChange()
+    {
+        return videoChannelsRate.getAccumulatedCount();
     }
 
     public Jid getJid()
@@ -359,7 +362,10 @@ public class Bridge
      */
     public double getStress()
     {
-        double stress = (lastReportedPacketRatePps + videoStreamCountDiff * AVG_PARTICIPANT_PACKET_RATE_PPS) / MAX_TOTAL_PACKET_RATE_PPS;
+        double stress =
+            (lastReportedPacketRatePps
+                + Math.max(0, getRecentVideoChannelChange()) * AVG_PARTICIPANT_PACKET_RATE_PPS)
+            / MAX_TOTAL_PACKET_RATE_PPS;
         return Math.min(1, stress);
     }
 
--- a/src/main/java/org/jitsi/jicofo/bridge/Bridge.java
+++ b/src/main/java/org/jitsi/jicofo/bridge/Bridge.java
@@ -335,8 +335,8 @@ public class Bridge
      */
     public double getStress()
     {
-        return Math.min(MAX_TOTAL_BITRATE_BPS,
-            (lastReportedBitrateKbps + videoStreamCountDiff * avgVideoStreamBitrateKbps)) / MAX_TOTAL_BITRATE_BPS;
+        double stress = (lastReportedBitrateKbps + videoStreamCountDiff * avgVideoStreamBitrateKbps) / MAX_TOTAL_BITRATE_BPS;
+        return Math.min(1, stress);
     }
 
     /**
--- a/src/main/java/org/jitsi/jicofo/bridge/Bridge.java
+++ b/src/main/java/org/jitsi/jicofo/bridge/Bridge.java
@@ -55,12 +55,12 @@ public class Bridge
      * We assume that each recently added participant contributes this much
      * to the bridge's packet rate.
      */
-    private static final int AVG_PARTICIPANT_PACKET_RATE_PPS = 500;
+    private static int AVG_PARTICIPANT_PACKET_RATE_PPS;
 
     /**
      * We assume this is the maximum packet rate that a bridge can handle.
      */
-    private static final double MAX_TOTAL_PACKET_RATE_PPS;
+    private static double MAX_TOTAL_PACKET_RATE_PPS;
 
     static
     {
@@ -72,14 +72,24 @@ public class Bridge
             5 /* numberOfLocalReceivers */
         );
 
-        MAX_TOTAL_PACKET_RATE_PPS =
+        setMaxTotalPacketRatePps(
             packetRateCalculator.computeIngressPacketRatePps()
-                + packetRateCalculator.computeEgressPacketRatePps();
+                + packetRateCalculator.computeEgressPacketRatePps());
 
-        logger.info("Using max total packet rate of " + MAX_TOTAL_PACKET_RATE_PPS);
-        logger.info("Using average participant packet rate of " + AVG_PARTICIPANT_PACKET_RATE_PPS);
+        setAvgParticipantPacketRatePps(500);
     }
 
+    static void setMaxTotalPacketRatePps(int maxTotalPacketRatePps)
+    {
+        MAX_TOTAL_PACKET_RATE_PPS = maxTotalPacketRatePps;
+        logger.info("Setting max total packet rate of " + MAX_TOTAL_PACKET_RATE_PPS);
+    }
+
+    static void setAvgParticipantPacketRatePps(int avgParticipantPacketRatePps)
+    {
+        AVG_PARTICIPANT_PACKET_RATE_PPS = avgParticipantPacketRatePps;
+        logger.info("Setting average participant packet rate of " + AVG_PARTICIPANT_PACKET_RATE_PPS);
+    }
     /**
      * The stress-level beyond which we consider a bridge to be
      * overloaded/overstressed.
--- a/src/main/java/org/jitsi/jicofo/bridge/Bridge.java
+++ b/src/main/java/org/jitsi/jicofo/bridge/Bridge.java
@@ -64,15 +64,15 @@ public class Bridge
 
     static
     {
-        MaxPacketRateCalculator calc = new MaxPacketRateCalculator(
-            2 /* numberOfConferenceBridges */,
+        MaxPacketRateCalculator packetRateCalculator = new MaxPacketRateCalculator(
+            4 /* numberOfConferenceBridges */,
             20 /* numberOfGlobalSenders */,
-            5 /* numberOfSpeakers */
+            2 /* numberOfSpeakers */
         );
 
         MAX_TOTAL_PACKET_RATE_PPS =
-            calc.computeIngressPacketRatePps(20 /* numberOfLocalSenders */)
-                + calc.computeEgressPacketRatePps(20 /* numberOfLocalSenders */, 5 /* numberOfLocalReceivers */);
+            packetRateCalculator.computeIngressPacketRatePps(20 /* numberOfLocalSenders */)
+                + packetRateCalculator.computeEgressPacketRatePps(20 /* numberOfLocalSenders */, 5 /* numberOfLocalReceivers */);
     }
 
     /**
--- a/src/main/java/org/jitsi/jicofo/bridge/Bridge.java
+++ b/src/main/java/org/jitsi/jicofo/bridge/Bridge.java
@@ -54,19 +54,25 @@ public class Bridge
      * A conservative estimate of the average packet rate (in kbps) of a video
      * stream flowing through the bridge.
      */
-    private static final int avgVideoStreamPacketRatePps = 500;
+    private static final int AVG_VIDEO_STREAM_PACKET_RATE_PPS = 500;
 
     /**
      * This is the worst case scenario that the bridge must be able to handle
-     * for a 100-peeps call: 20 local senders and 5 local receivers. Most bits
-     * are wasted in octo traffic.
+     * in a 100-peeps call with 20 local senders and 5 local receivers.
      */
     private static final double MAX_TOTAL_PACKET_RATE_PPS;
 
     static
     {
-        MaxPacketRateCalculator calc = new MaxPacketRateCalculator(2, 20, 5);
-        MAX_TOTAL_PACKET_RATE_PPS = calc.computeMaxDownload(20) + calc.computeMaxUpload(20, 5);
+        MaxPacketRateCalculator calc = new MaxPacketRateCalculator(
+            2 /* numberOfConferenceBridges */,
+            20 /* numberOfGlobalSenders */,
+            5 /* numberOfSpeakers */
+        );
+
+        MAX_TOTAL_PACKET_RATE_PPS =
+            calc.computeIngressPacketRatePps(20 /* numberOfLocalSenders */)
+                + calc.computeEgressPacketRatePps(20 /* numberOfLocalSenders */, 5 /* numberOfLocalReceivers */);
     }
 
     /**
@@ -341,7 +347,7 @@ public class Bridge
      */
     public double getStress()
     {
-        double stress = (lastReportedPacketRatePps + videoStreamCountDiff * avgVideoStreamPacketRatePps) / MAX_TOTAL_PACKET_RATE_PPS;
+        double stress = (lastReportedPacketRatePps + videoStreamCountDiff * AVG_VIDEO_STREAM_PACKET_RATE_PPS) / MAX_TOTAL_PACKET_RATE_PPS;
         return Math.min(1, stress);
     }
 
--- a/src/main/java/org/jitsi/jicofo/bridge/Bridge.java
+++ b/src/main/java/org/jitsi/jicofo/bridge/Bridge.java
@@ -337,13 +337,13 @@ public class Bridge
 
     /**
      * Returns the "stress" of the bridge. The stress is computed based on the
-     * total bitrate reported by the bridge and the video stream estimation
-     * since the last update from the bridge.
+     * total packet rate reported by the bridge and the video stream diff
+     * estimation since the last update from the bridge.
      *
-     * @return the sum of the last total reported bitrate (in kbps) and an
-     * estimation of the bitrate of the streams that we estimate that the bridge
+     * @return the sum of the last total reported packet rate (in pps) and an
+     * estimation of the packet rate of the streams that we estimate that the bridge
      * hasn't reported to Jicofo yet. The estimation is the product of the
-     * number of unreported streams and a constant C (which we set to 1000 kbps).
+     * number of unreported streams and a constant C (which we set to 500 pps).
      */
     public double getStress()
     {
--- a/src/main/java/org/jitsi/jicofo/bridge/Bridge.java
+++ b/src/main/java/org/jitsi/jicofo/bridge/Bridge.java
@@ -50,6 +50,25 @@ public class Bridge
         = new ColibriStatsExtension();
 
     /**
+     * A conservative estimate of the average bitrate (in kbps) of a video
+     * stream flowing through the bridge.
+     */
+    private static final int avgVideoStreamBitrateKbps = 1000;
+
+    /**
+     * This is the worst case scenario that the bridge must be able to handle
+     * for a 100-peeps call: 20 local senders and 5 local receivers. Most bits
+     * are wasted in octo traffic.
+     */
+    private static final double MAX_TOTAL_BITRATE_BPS = 560_000;
+
+    /**
+     * The stress-level beyond which we consider a bridge to be
+     * overloaded/overstressed.
+     */
+    private static final double OVERSTRESSED_THRESHOLD = .8;
+
+    /**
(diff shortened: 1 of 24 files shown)


src/main/java/org/jitsi/jicofo/bridge/Bridge.java
```diff
@@ -343,9 +363,35 @@ public String getRegion()
...
+     * since the last update from the bridge.
+     *
+     * @return the sum of the last total reported bitrate (in kbps) and an
+     * estimation of the bitrate of the streams that we estimate that the bridge
+     * hasn't reported to Jicofo yet. The estimation is the product of the
+     * number of unreported streams and a constant C (which we set to 1000 kbps).
+     */
+    public double getStress()
+    {
+        return Math.min(MAX_TOTAL_BITRATE_BPS,
```
bbaldino: I'm confused by the scale between these two things we're taking the minimum of.  Isn't the goal of this to return a number between 0 and 1?  But the first arg is  a maximum total bitrate in bps?

gpolitis: @bbaldino I'm just making sure the number doesn't become greater than 1 (in case our bitrate estimation is greater than MAX_TOTAL_BITRATE_BPS).

bbaldino: So `getStress` should return a number between 0 and 1, right?  But `MAX_TOTAL_BITRATE_BPS` is `560_000` ?

gpolitis: Yeah but we scale by that max, there should be a division there: `Math.min(max, estimate) / max` is always between 0 and 1, right?

bbaldino: Sorry!  I misread where the closing parenthesis was!  I wonder if it would be clearer to do:
```
double stress = (lastReportedBitrateKbps + videoStreamCountDiff * avgVideoStreamBitrateKbps) / MAX_TOTAL_BITRATE_BPS;
return Math.min(1, stress);
```

gpolitis:
I've adapted the stress computation to be based on the packet rate and I've added the stats you suggested @bgrozev. /cc @bbaldino @JonathanLennox

src/main/java/org/jitsi/jicofo/bridge/BridgeSelectionStrategy.java
```diff
@@ -86,61 +93,248 @@ public Bridge select(
...
-     * available.
+     * @param bridges
+     * @param conferenceBridges
+     * @param participantRegion
+     *
+     * @return
      */
-    private Bridge selectInitial(List<Bridge> bridges,
-                                 String participantRegion)
+    private Optional<Bridge> a1(List<Bridge> bridges,
```
bbaldino: I think it'd be clearer to try and come up with descriptive names for these helpers.  Like `findBridgeAlreadyInConferenceInRegion`.  Personally I'd rather have a long, descriptive name.  Then the cascade of calls ends up looking like:

```
findBridgeAlreadyInConferenceInRegion(bridges, conferenceBridges, participantRegion).orElse(
    findNewBridgeInRegion(bridges, conferenceBridges, participantRegion).orElse(
        ...
```

Which I think reads better.  (I think the fact that we're looking for a non-overloaded bridged in each case is implicit and could just be documented on the method).

src/main/java/org/jitsi/jicofo/bridge/BridgeSelectionStrategy.java
```diff
@@ -86,61 +93,248 @@ public Bridge select(
...
+     * A2. (Happy case 2): There is a NOL bridge in the region, and the
+     * conference has no bridges in the region. Use the least loaded bridge in
+     * the region.
+     *
+     * @param bridges
+     * @param participantRegion
+     *
+     * @return
+     */
+    private Optional<Bridge> a2(List<Bridge> bridges,
```
bbaldino: Also, do all these methods belong in the abstract base class?  It seems to me like these should be implemented in a specific strategy (especially since they're private, so doesn't seem like they're intended for re-use by other implementations?)

src/main/java/org/jitsi/jicofo/bridge/BridgeSelectionStrategy.java
```diff
@@ -168,4 +372,19 @@ void setLocalRegion(String localRegion)
     {
         this.localRegion = localRegion;
     }
+
+    public JSONObject getStats()
+    {
+        JSONObject json = new JSONObject();
+
+        json.put("total_a1", totalA1);
```
bbaldino: I think these should be more descriptive as well, otherwise we'll have to look them up every time.

src/main/java/org/jitsi/jicofo/bridge/Bridge.java
```diff
@@ -49,6 +50,31 @@
     private static final ColibriStatsExtension EMPTY_STATS
         = new ColibriStatsExtension();
 
+    /**
+     * A conservative estimate of the average packet rate (in kbps) of a video
+     * stream flowing through the bridge.
```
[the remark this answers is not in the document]
gpolitis: Judging from where this is coming from (ColibriConferenceImpl.trackVideoChannelsAddedRemoved), I think it's equivalent to a participant.

src/main/java/org/jitsi/jicofo/bridge/BridgeSelectionStrategy.java
gpolitis: I restructured the code because the previous version was always evaluating the expression in the orElse() method call, even if it wasn't necessary.  So unfortunately now the method name findBridgeAlreadyInConferenceInRegionOrRestOfDecisionTree which is too long I think. Any suggestions how to deal with that?

src/main/java/org/jitsi/jicofo/bridge/BridgeSelectionStrategy.java
gpolitis: Maybe we should encapsulate this decision tree in a separate class? It's common to all strategies: in the single bridge selection strategy for example we use the logic to select the single bridge.

src/main/java/org/jitsi/jicofo/bridge/BridgeSelectionStrategy.java
gpolitis: yeah agreed, it doesn't mean anything like this. Any good name suggestions?

src/main/java/org/jitsi/jicofo/bridge/BridgeSelectionStrategy.java
```diff
@@ -86,61 +93,248 @@ public Bridge select(
     }
 
     /**
-     * Selects a bridge to be used for a new participant in a conference,
-     * assuming that no other bridge is used by the conference (i.e. this
-     * is the initial selection of a bridge for the conference).
+     * A1. (Happy case 1): There is a NOL bridge in the conference and in the
```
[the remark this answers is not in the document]
gpolitis: I don't mind, either is fine to me

src/main/java/org/jitsi/jicofo/bridge/Bridge.java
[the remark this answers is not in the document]
bbaldino: Ok, I just think we should add to the comment & variable name to make that clearer--that isn't what I would expect when reading "video stream".

src/main/java/org/jitsi/jicofo/bridge/Bridge.java
```diff
@@ -54,19 +54,25 @@
...
     private static final double MAX_TOTAL_PACKET_RATE_PPS;
 
     static
     {
-        MaxPacketRateCalculator calc = new MaxPacketRateCalculator(2, 20, 5);
-        MAX_TOTAL_PACKET_RATE_PPS = calc.computeMaxDownload(20) + calc.computeMaxUpload(20, 5);
+        MaxPacketRateCalculator calc = new MaxPacketRateCalculator(
+            2 /* numberOfConferenceBridges */,
+            20 /* numberOfGlobalSenders */,
+            5 /* numberOfSpeakers */
```
bbaldino: 👍

src/main/java/org/jitsi/jicofo/util/MaxPacketRateCalculator.java
```diff
@@ -34,43 +34,48 @@ public MaxPacketRateCalculator(int numberOfConferenceBridges, int numberOfGlobal
...
+    public int computeIngressPacketRatePps(int numberOfLocalSenders)
     {
-        // regardless of the participant distribution, in a 100 people call each
-        // sender receivers 19 other senders.
-        return numberOfLocalSenders*Arrays.stream(packetRatePps).sum();
+        return numberOfLocalSenders*Arrays.stream(maxPacketRatePps).sum();
     }
 
-    public int computeMaxOctoSendBitrate(int numberOfLocalSenders)
+    public int computeOctoEgressPacketRate(int numberOfLocalSenders)
```
bbaldino: I'm a little confused by which variables are fixed (set in the constructor) and which are passed to these methods dynamically.  If the number of local senders can change, how is it that the number of global senders (passed in constructor) is fixed?

src/main/java/org/jitsi/jicofo/bridge/BridgeSelectionStrategy.java
```diff
@@ -327,14 +355,10 @@ public Bridge selectLeastLoadedNearby(List<Bridge> bridges,
...
-        return a1(bridges, conferenceBridges, participantRegion).orElse(
-            a2(bridges, conferenceBridges, participantRegion).orElse(
-                b1(bridges, conferenceBridges, participantRegion).orElse(
-                    c1(bridges, conferenceBridges, participantRegion).orElse(
-                        c2(bridges, conferenceBridges, participantRegion).orElse(
-                            d1(bridges, conferenceBridges, participantRegion).orElse(
-                                d2(bridges, conferenceBridges, participantRegion).orElse(null)))))));
+        else
+        {
+            return a1(bridges, conferenceBridges, participantRegion);
```
bbaldino: Using `orElseGet` instead of `orElse` will give you lazy evaluation

src/main/java/org/jitsi/jicofo/bridge/Bridge.java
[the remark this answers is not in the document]
gpolitis: 👍

src/main/java/org/jitsi/jicofo/bridge/BridgeSelectionStrategy.java
gpolitis: I reverted the commit that refactored the code and replaced the orElse with orElseGet.

src/main/java/org/jitsi/jicofo/bridge/BridgeSelectionStrategy.java
gpolitis: I brought back the previous code structure and tried to come up with better method names. I also removed rule b1 which identical to a2 code-wise.

src/main/java/org/jitsi/jicofo/util/MaxPacketRateCalculator.java
gpolitis: I refactored the calculator a bit.

src/main/java/org/jitsi/jicofo/bridge/Bridge.java
```diff
@@ -49,6 +50,49 @@
...
+            5 /* numberOfLocalReceivers */
+        );
+
+        MAX_TOTAL_PACKET_RATE_PPS =
+            packetRateCalculator.computeIngressPacketRatePps()
+                + packetRateCalculator.computeEgressPacketRatePps();
+
+        // This currently evaluates to 546. It assumes that the new participant
+        // will be sending video. We're using the "Max" packet rate calculator
+        // so we divide by 3 (an arbitrary factor) to figure out a more
```
bbaldino: I don't follow the `3`.  It's just based on assuming a participant will send 1/3rd the max on average?  Is this based on the idea that many people are video muted?

src/main/java/org/jitsi/jicofo/bridge/Bridge.java
```diff
@@ -49,6 +50,49 @@
...
+
+    /**
+     * This is the worst case scenario that the bridge must be able to handle
+     * in a 100-peeps call with 20 local senders and 5 local receivers.
+     */
+    private static final double MAX_TOTAL_PACKET_RATE_PPS;
+
+    static
+    {
+        MaxPacketRateCalculator packetRateCalculator = new MaxPacketRateCalculator(
```
bbaldino: So this is intended to reflect a "worst case" packet calculator?

src/main/java/org/jitsi/jicofo/bridge/BridgeSelectionStrategy.java
```diff
@@ -86,61 +93,216 @@ public Bridge select(
     }
 
     /**
-     * Selects a bridge to be used for a new participant in a conference,
-     * assuming that no other bridge is used by the conference (i.e. this
-     * is the initial selection of a bridge for the conference).
+     * A1. (Happy case 1): There is a NOL bridge in the conference and in the
+     * region. Use the least loaded of them.
```
bbaldino: I think we should update these docs to reflect what the method does, rather than the scenario from the doc.  E.g.:
```
Find a not-overloaded bridge already handling this conference in the participant's region
```

src/main/java/org/jitsi/jicofo/bridge/BridgeSelectionStrategy.java
```diff
@@ -86,61 +93,216 @@ public Bridge select(
...
-     * available.
+     * @param bridges
+     * @param conferenceBridges
+     * @param participantRegion
+     *
+     * @return
      */
-    private Bridge selectInitial(List<Bridge> bridges,
-                                 String participantRegion)
+    private Optional<Bridge> findNotLoadedBridgeAlreadyInConferenceInRegion(List<Bridge> bridges,
```
bbaldino: Is there any point in taking in or searching `bridges` here?  We filter based on `conferenceBridges` anyway, so it seems like we should just start with `conferenceBridges` directly.

src/main/java/org/jitsi/jicofo/bridge/BridgeSelectionStrategy.java
```diff
@@ -86,61 +93,216 @@ public Bridge select(
...
+            .filter(selectFrom(conferenceBridges))
+            .filter(inRegion(participantRegion))
+            .findFirst();
 
-        // Prefer a bridge in the participant's region.
-        if (participantRegion != null)
+        if (result.isPresent())
         {
-            bridge = findFirstOperationalInRegion(bridges, participantRegion);
+            logger.info("bridge_selected,rule=a1,bridge=" + result.get() +
```
bbaldino: I don't think we need to refer to `rule=a1` here (the method name will now give enough context)

src/main/java/org/jitsi/jicofo/bridge/BridgeSelectionStrategy.java
```diff
@@ -86,61 +93,216 @@ public Bridge select(
...
+     * A2. (Happy case 2): There is a NOL bridge in the region, and the
+     * conference has no bridges in the region. Use the least loaded bridge in
+     * the region.
+     *
+     * @param bridges
+     * @param participantRegion
+     *
+     * @return
+     */
+    private Optional<Bridge> findNotLoadedBridgeInRegion(List<Bridge> bridges,
```
bbaldino: Here it looks like we don't need `conferenceBridges`.  I don't think making the arguments consistent (even when they're not used) across these methods is valuable.

src/main/java/org/jitsi/jicofo/bridge/BridgeSelectionStrategy.java
```diff
@@ -86,61 +93,216 @@ public Bridge select(
...
+                                                         List<Bridge> conferenceBridges,
+                                                         String participantRegion)
+    {
+        // NOTE that A2 is equivalent to B1 but we include it as a separate
```
bbaldino: Is this comment still relevant?  Looks like all references to B1 can go away (and I think we should move away from the A1, A2, etc. references in general)

src/main/java/org/jitsi/jicofo/bridge/BridgeSelectionStrategy.java
bbaldino: I think the names you used for the methods are good (tweaked a bit...without the leading `find`, for example)

src/main/java/org/jitsi/jicofo/bridge/Bridge.java
gpolitis: Yeah, all that + we assumed a 20 sender call in the packet rate calculation. Maybe it'd be cleaner to assume a smaller call and just take the result of that calculation? Or keep the large call and just take the result of the calculation (no division by 3)

src/main/java/org/jitsi/jicofo/bridge/Bridge.java
gpolitis: Yes

src/main/java/org/jitsi/jicofo/bridge/BridgeSelectionStrategy.java
gpolitis: No point whatsoever other than to have uniform logging that can be useful when debugging. Shall I prune the args list?

src/main/java/org/jitsi/jicofo/bridge/BridgeSelectionStrategy.java
gpolitis: 👍

src/main/java/org/jitsi/jicofo/bridge/Bridge.java
```diff
@@ -56,7 +56,7 @@
      * "video stream" comes from {@link ColibriConferenceImpl#trackVideoChannelsAddedRemoved},
      * so it's equivalent to a participant.
      */
-    private static final int AVG_VIDEO_STREAM_PACKET_RATE_PPS = 500;
+    private static final int AVG_VIDEO_STREAM_PACKET_RATE_PPS;
```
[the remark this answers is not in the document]
gpolitis: I tweaked the constant name but now there's this multiplication that looks like we're multiplying apples by oranges `videoStreamCountDiff * AVG_PARTICIPANT_PACKET_RATE_PPS`. We'd have to go all the way up to the source to be consistent (I don't mind doing that if that's what we want to do).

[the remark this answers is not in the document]
bbaldino: So you're saying a `videoStreamCountDiff` value of 1, for example, represents a single participant?

I don't think we necessarily need to address that part now.  I'm glad we go this constant fixed, at least.

src/main/java/org/jitsi/jicofo/bridge/BridgeSelectionStrategy.java
```diff
@@ -86,61 +98,205 @@ public Bridge select(
...
+     * is already handling the conference.
+     *
+     * @param bridges the list of operational bridges, ordered by load.
+     * @param conferenceBridges the set of bridges that are already used in the conference.
+     * @param participantRegion the participant region.
+     *
+     * @return an optional that contains the least loaded bridge that is already
+     * in the conference and that is in the participant region if it exists.
+     */
+    private Optional<Bridge> leastLoadedAlreadyInConferenceInRegion(List<Bridge> bridges,
```
bbaldino: We don't need `bridges` here, and can just start with `conferenceBridges` since we filter to those anyway.

src/main/java/org/jitsi/jicofo/bridge/Bridge.java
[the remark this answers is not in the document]
gpolitis: > So you're saying a videoStreamCountDiff value of 1, for example, represents a single participant?

yeah

src/main/java/org/jitsi/jicofo/bridge/BridgeSelectionStrategy.java
gpolitis: The bridges list is prioritized by load whereas the conferenceBridges list is not. We should probably make this more obvious by changing the conferenceBridges argument to be a `Set`.

src/test/java/org/jitsi/jicofo/bridge/BridgeTest.java
```diff
@@ -0,0 +1,82 @@
...
+        Jid jvb1Jid = JidCreate.from("jvb.example.com");
+        Bridge bridge = new Bridge(null, jvb1Jid, null);
+
+        // In the most balanced scenario, in a 100 people call over 4 bridges,
+        // each bridge handles 25 participants. Note that we can't know in
+        // advance whether a participant will be sendrecv or a recvonly.
+
+        /*
+         * Where do the expected values come from? Are they hard-coded to match
+         * the implementation (as it was in a previous commit)?
```
gpolitis: The values come from the implementation. It gives you an idea of how the system will behave in practice and you can adjust the code accordingly.

src/main/java/org/jitsi/jicofo/bridge/RegionBasedBridgeSelectionStrategy.java
```diff
@@ -1,82 +1,76 @@
...
+ * and in the region: use the least loaded of them.
+ * (Happy case 2): If there is a non-overloaded bridge in the region,
+ * and the conference has no bridges in the region: use the least loaded bridge
+ * in the region.
+ *
+ * (Split case 1): If there is a non-overloaded bridge in the region, the
+ * conference has bridges in the region but all are overloaded: Use the least
+ * loaded of the bridges in the region.
+ *
+ * (Overload case 1): If all bridges in the region are overloaded, and the
```
gpolitis: I think B1 is the same as A2, at least they're identical in the code https://github.com/jitsi/jicofo/pull/428/commits/3e075fb194500e1feb9c43036f286e1bc2998688

src/main/java/org/jitsi/jicofo/bridge/BridgeSelectionStrategy.java
```diff
@@ -86,63 +139,215 @@ public Bridge select(
...
-     * @param bridges
-     * @param region
-     * @return
+     * @return an optional that contains the least loaded bridge if it exists.
      */
-    private Bridge findFirstOperationalInRegion(
+    Optional<Bridge> leastLoaded(
             List<Bridge> bridges,
-            String region)
+            Map<Bridge, Integer> conferenceBridges,
```
bbaldino: It's a bummer we have to take in `conferenceBridges` and `participantRegion` here just to call `updateSplitStats`...is there another place we could call `updateSplitStats` from? Or another way to track that?

bgrozev: It's also for consistency -- all these methods in the base class that make a selection *and* update the stats (i.e. assume their selection was used) have the same prototype. 

Since calling `updateSplitStats` only happens in some cases, we would have to move this logic to the `doSelect` implementations which would be quite awkward and confusing:
https://github.com/jitsi/jicofo/pull/428/files#diff-d6bf891e4163817e6f64797ce48fbb1aR47
https://github.com/jitsi/jicofo/pull/428/files#diff-9c48bdfd8bc771eb097b1aa7f372a011R69

bbaldino: Ok, I had commented to george before that I thought it was misleading to pass values (e.g. `conferenceBridges` and `participantRegion`) to methods that don't actually take them into account.

src/main/java/org/jitsi/jicofo/bridge/Bridge.java
gpolitis: Marking this as resolved as it was boomed.

src/main/java/org/jitsi/jicofo/bridge/BridgeSelectionStrategy.java
gpolitis: Marking this conversation as resolved as the code was refactored.

PR #432, opened 2020-03-13
Use packet rate and add stats.

bgrozev:


src/main/java/org/jitsi/jicofo/bridge/BridgeSelectionStrategy.java
```diff
@@ -41,10 +41,9 @@ public Bridge select(
         {
             Bridge bridge
                 = selectInitial(bridges, participantRegion);
-            if (logger.isDebugEnabled())
-            {
-                logger.debug("Selected initial bridge " + bridge);
-            }
+            logger.info("Selected initial bridge " + bridge
```
bbaldino: Should we log some indication of the loads (packet rates) of the other bridges?  Are we exposing that in any way?

src/main/java/org/jitsi/jicofo/bridge/Bridge.java
```diff
@@ -348,4 +302,10 @@ public String toString()
                      getRelayId(),
                      getRegion());
     }
+
+    public int getLastReportedPacketRatePps()
```
bbaldino: How is it this doesn't seem to get called by the selection code?  Or anywhere exception for logging/stats?

src/main/java/org/jitsi/jicofo/bridge/Bridge.java
```diff
@@ -61,26 +60,17 @@
...
-     * <tt>null</tt>.
-     *
-     * Is is set back to <tt>null</tt> when new value from the bridge
-     * arrives.
+     * Estimates the number of new video streams added (or removed if negative)
+     * on this bridge by this jicofo instance since the last statistics update
+     * was received.
      */
+    @SuppressWarnings("unused")
     private int videoStreamCountDiff = 0;
```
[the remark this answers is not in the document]
bgrozev: It will be used when we rebase George's PR

src/main/java/org/jitsi/jicofo/bridge/Bridge.java
```diff
@@ -167,6 +176,11 @@ void setStats(ColibriStatsExtension stats)
             lastReportedBitrateKbps = bitrate;
         }
 
+        if (packetRateDown != null && packetRateUp != null)
```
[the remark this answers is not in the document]
bgrozev: Yes. I think this might happen once when the bridge advertises stats for the first time, when only a small subset of them are included. All subsequent reports will have both

src/main/java/org/jitsi/jicofo/bridge/Bridge.java
bgrozev: Becaus `compareTo` uses the field directly. I can fix if you like

src/main/java/org/jitsi/jicofo/bridge/BridgeSelectionStrategy.java
bgrozev: We can, but we expect to have 600 of them...

bbaldino: Ok, so maybe we say this is nice to have but we won't be able to fully answer the question of "why did jicofo pick this bridge and not another" until we move this to the bridge-side?

src/main/java/org/jitsi/jicofo/bridge/Bridge.java
bbaldino: Gotcha, I see it now.  I think it's fine as-is.

PR #437, opened 2020-03-18
Don't consider failed and non-operational bridges for Octo.

JonathanLennox:
Note: it's possible more places need to use the new `operationalBridges()` method rather than use `bridges` directly, but this should be a start?

### DIFF
--- a/src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
+++ b/src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
@@ -822,6 +822,16 @@ public class JitsiMeetConferenceImpl
     }
 
     /**
+     * Get a stream of those bridges which are operational.
+     * Caller should be synchronized on bridges.
+     */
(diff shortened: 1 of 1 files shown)


src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
```diff
@@ -821,6 +821,16 @@ private Bridge selectBridge(Participant participant)
         return bridge;
     }
 
+    /**
+     * Get a stream of those bridges which are operational.
+     * Caller should be synchronized on bridges.
+     */
+    private Stream<BridgeSession> operationalBridges()
```
bgrozev: Can we synchronize access to `bridges` to prevent new code from calling the function without synchronizing?

JonathanLennox: No, because we return a `Stream<>` that iterates over `bridges`.

PR #444, opened 2020-03-24
Fix jibri iq cast

paweldomas:


src/main/java/org/jitsi/jicofo/recording/jibri/CommonJibriStuff.java
```diff
@@ -262,7 +263,10 @@ final synchronized IQ handleIQRequest(JibriIq iq)
                 // (so it isn't XMPPError.Condition.service_unavailable), so it
                 // must be that they're all busy.
                 logger.info("Failed to start a Jibri session, all Jibris were busy");
-                return IQ.createErrorResponse(iq, XMPPError.Condition.resource_constraint);
+                return ErrorResponse.create(
```
bbaldino: What is the difference between `ErrorResponse.create` and `IQ.createErrorResponse` ?

bbaldino: Ah, I see it's a new helper now.

src/main/java/org/jitsi/jicofo/recording/jibri/JibriSession.java
```diff
@@ -781,4 +796,23 @@ void onSessionStateChanged(
                 JibriIq.Status    newStatus,
                 JibriIq.FailureReason         failureReason);
     }
+
+    public enum ErrorReason {
```
bbaldino: What about calling this something like `JibriResult` and including a 'Success' value in it that we can use instead of returning null?

paweldomas: okay

paweldomas: I don't know if it feels better or worse. What I don't like is that now Result mixes success with errors, but on the other hand it reads better when checks for Result.SUCCESS instead of null.

paweldomas: maybe Exception would be best

paweldomas:
I've squashed and force pushed the last 3 commits into f53e4de

PR #447, opened 2020-03-25
Shibboleth Authentification Fix

### LINKED ISSUE #446: Regression bug for Shibboleth Authentication
Setting up Jicofo with Shibboleth Authentication on Apache in the latest build we came across an error which we think might have been introduced by #408 as it sets a stringent check on the target, while we have found the target to come with a trailing forward slash at the end. This might also be partially responsible for #441 although an incorrect configuration is more likely given the limited documentation for implementing Shibboleth. This also means that #404 was only partially fixed.

SimonMaenaut:
Replacing equals with startsWith should fix #446 for Shibboleth.

### DIFF
--- a/src/main/java/org/jitsi/jicofo/auth/ShibbolethHandler.java
+++ b/src/main/java/org/jitsi/jicofo/auth/ShibbolethHandler.java
@@ -81,7 +81,7 @@ class ShibbolethHandler
             throws IOException,
             ServletException
     {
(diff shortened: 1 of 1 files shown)


SimonMaenaut:
Signed the CLA

damencho:
I'm good. Target cannot be null right?

bertvandepoel:
From our testing, target is at least "/". But feel free to check!

PR #449, opened 2020-03-28
Add optional sentry error tracking

sapkra:
Because Jitsi is using `java.util.logging` it is quite easy to implement:
https://docs.sentry.io/clients/java/integrations/#javautillogging

To configure the Sentry client the SENTRY_DSN environment variable has to be set. But there are also some other possibilities.

The same changes for the other jitsi repositories can be found here:
- https://github.com/jitsi/jigasi/pull/247
- https://github.com/jitsi/jitsi-videobridge/pull/1152


sapkra:
It's working fine. I can see all the errors in sentry now.
image

PR #458, opened 2020-04-06
feat: Bumps jicoco version, bringing muc retry on connect.

damencho:


paweldomas:
It doesn't hurt to update, but I don't think jicofo uses the `MucClient`

damencho:
Oooh you are right. Brian was hitting the item-not-found issue today when restarting prosody ...

PR #459, opened 2020-04-07
Clarify how to add a "key=value" in the .properties file

famzah:
For people not experienced with Java, it's not clear how to append the additional "key=value" config.


famzah:
I've just signed the CLA and received confirmation by email.

bgrozev:
Thanks you

PR #462, opened 2020-04-09
Fixes room destruction after duration expires.

michal-stoma:
This is to fix behaviour described at: https://community.jitsi.org/t/room-not-being-destroyed-after-duration-exceeded/29768
When reservation API sets duration for conference, this is not respected and room is not destroyed upon duration's expiry,
with error like that:
java.lang.RuntimeException: org.jxmpp.stringprep.XmppStringprepException: XmppStringprepException caused by 'null': org.jxmpp.stringprep.XmppStringprepException: Input 'domain' must not be null

### DIFF
--- a/src/main/java/org/jitsi/impl/protocol/xmpp/ChatRoomImpl.java
+++ b/src/main/java/org/jitsi/impl/protocol/xmpp/ChatRoomImpl.java
@@ -877,7 +877,9 @@ public class ChatRoomImpl
     {
         try
         {
-            muc.destroy(reason, JidCreate.entityBareFrom(alternateAddress));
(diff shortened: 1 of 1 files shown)


amit1576:
@damencho ,

Is this fix available at ### https://download.jitsi.org stable. or we need to wait.

damencho:
This is still not in stable.

amit1576:
Ok, any idea when can i get it. or should i use testing ?

damencho:
Next week or the week after. You can switch to unstable repo and update from there.

amit1576:
Ok, Thanks.

PR #464, opened 2020-04-11
Update README.md

ongolaboy:
The previous command is ejabberd-centric. It's better to use the native one of prosody when creating a new user.

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,8 +131,9 @@ org.jitsi.jicofo.auth.URL=XMPP:jitsi-meet.example.com
 
 4 To create users use the command:
 ```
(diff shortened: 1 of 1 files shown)


gpolitis:
I don't see anything ejabberd specific in the code that you're trying to modify but your version has  the advantage that it doesn't require typing the password in cleartext on command line, potentially storing it in your history.

ongolaboy:
Actually @gpolitis, here is what you have in the man of prosodyctl(1) , starts at line 126 https://hg.prosody.im/0.11/file/tip/man/prosodyctl.markdown#l126 .

PR #465, opened 2020-04-11
Add note on service restart.

jeannekamikaze:
Add a 6th step to the secure domain setup process instructing the user to restart services after applying the relevant changes to the configuration files.


jeannekamikaze:
Hello. When following the instructions to set up a secure domain, I was restarting jicofo, but it was not obvious to me that I also had to restart prosody. I left a note on the manual instructing the user to restart both services after applying all of the configuration changes. Thanks.

bgrozev:

PR #471, opened 2020-04-15
feat(stats): add info about Jibri sessions

paweldomas:
Adds the following to the JSON stats:

"live_streaming_active" - active Jibri live streaming sessions
"recording_active" - active Jibri recording sessions
"sip_call_active" - active Jibri SIP calls
"live_streaming_pending" - pending live streaming Jibri sessions
"recording_pending" - pending recording Jibri sessions
"sip_call_pending" - pending SIP call Jibri sessions

### DIFF
--- a/src/main/java/org/jitsi/jicofo/JitsiMeetConference.java
+++ b/src/main/java/org/jitsi/jicofo/JitsiMeetConference.java
@@ -19,6 +19,7 @@ package org.jitsi.jicofo;
 
 import net.java.sip.communicator.service.protocol.*;
 import org.jitsi.jicofo.bridge.*;
+import org.jitsi.jicofo.recording.jibri.*;
 import org.jitsi.protocol.xmpp.*;
 import org.jitsi.utils.logging.*;
 import org.jxmpp.jid.*;
@@ -86,6 +87,12 @@ public interface JitsiMeetConference
      * array is copied.
      */
     void setStartMuted(boolean[] startMuted);
+
+    /**
+     * @return a stats snapshot for all {@link JibriSession}s used in this
+     * conference.
+     */
+    JibriSessionStats getJibriSessionStats();
 
     /**
      * Gets the role of a member in the conference.
--- a/src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
+++ b/src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
@@ -3122,6 +3122,27 @@ public class JitsiMeetConferenceImpl
      * {@inheritDoc}
      */
     @Override
+    public JibriSessionStats getJibriSessionStats()
+    {
+        List<JibriSession> sessions = new ArrayList<>();
+
+        if (jibriRecorder != null)
+        {
+            sessions.addAll(jibriRecorder.getJibriSessions());
+        }
+
+        if  (jibriSipGateway != null)
+        {
+            sessions.addAll(jibriSipGateway.getJibriSessions());
+        }
+
+        return new JibriSessionStats(sessions);
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
     public String toString()
     {
         return
--- a/src/main/java/org/jitsi/jicofo/recording/jibri/CommonJibriStuff.java
+++ b/src/main/java/org/jitsi/jicofo/recording/jibri/CommonJibriStuff.java
@@ -157,6 +157,11 @@ public abstract class CommonJibriStuff
     protected abstract JibriSession getJibriSessionForMeetIq(JibriIq iq);
 
     /**
+     * @return a list with all {@link JibriSession}s used by this instance.
+     */
+    public abstract List<JibriSession> getJibriSessions();
+
+    /**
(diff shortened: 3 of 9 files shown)


src/main/java/org/jitsi/jicofo/JitsiMeetConference.java
```diff
@@ -87,6 +88,12 @@
      */
     void setStartMuted(boolean[] startMuted);
 
+    /**
+     * @return the list of all Jibri sessions(of any type) used by this
+     * conference.
+     */
+    List<JibriSession> getJibriSessions();
```
bbaldino, line 95: Instead of adding code to reach in and extract all Jibri sessions, can we just have the code push stats up to somewhere which would cache the counts and then we can query that cache for the stats?

paweldomas: do you mean to track the events when sessions starts etc. ? I'd rather not do that as it's more error prone

bbaldino: Yes I mean the code that starts a session fires a stat up that a session has started, etc.   What makes it more error prone?

paweldomas: That listeners must always be in place and the events have to be fired correctly, missing one event for transition to ended or failed will cause a leak. Querying the current state is also easier in my opinion. Why do you think events are better? Is it the list of all sessions that bothers you?

bbaldino: I was assuming we'd have a single static listener which always existed and would capture these and also be query-able for stats.  But yeah, I just thought it was unfortunate to have to plumb a call all the way down to get the list of sessions.

paweldomas: It is a bit unfortunate, but it's either that global query or global events.

bbaldino: Maybe I'm misreading this, but I thought the stats were already global...the `eventAdmin` calls?

paweldomas: I also think that counting events creates a second source of truth. Why not just generate the stats based on the real current state.

bbaldino: Just because it breaks encapsulation more to reach down and grab them.

bbaldino: That may also require locking, etc. to grab things which can conflict with code execution.  There's still a chance we just miss the start or end of a session.

paweldomas: if instead of returning the entire list I've passed a selector? is that better? then it's api just for querying and does not expose the whole set?

bbaldino: I'm also confused because we already had a `JibriStats` class which was tracking failures...why can't we just add these to that using the same method? Why push some stats up one way, and query for others?

paweldomas: because failures are not stored in any state, so it  creates a new state that stores them

bbaldino: So I guess I'm wondering: if we already have a mechanism for pushing Jibri-related events for the purposes of stat tracking, why not re-use it?

bbaldino: Thinking about it more, I do see how maybe we'd need to include a jibri ID or some common identifier for state  to track, for example, when a jibri goes from "available" to "busy" or something.  I do see how that's a bit more complicated than tracking error counts, and iterating over the current state makes it easier.  I don't know if it would be easy to deal with that (with a jibri id or whatever)

paweldomas: there are no events really for tracking  JibriSession's current status, so I'd have to add global events, then create a tracking object to update the current state and then count them anyway

paweldomas: yes, what you wrote above, it is more work and it creates another source of truth

bbaldino: I don't agree with the source of truth argument, because I don't see why pushing information up wouldn't be reliable, but I see that it can be complex to track the current state of sessions (primarily tracking which state they _left_).

paweldomas: my proposal is to change the getter of all session to a query method  for  counting them

bbaldino: If you mean adding something like a `getJibriSessionStats` method which returned some stats object or a `JSONObject`, then yeah I do think that would be better.

paweldomas: ok

PR #474, opened 2020-04-20
fix(JvbDoctor): release the lock for feature discovery

paweldomas:
verifyHealthCheckSupport() calls to DiscoveryUtil.checkFeatureSupport
which is a blocking operation while waiting for disco info response.

### DIFF
--- a/src/main/java/org/jitsi/jicofo/bridge/JvbDoctor.java
+++ b/src/main/java/org/jitsi/jicofo/bridge/JvbDoctor.java
@@ -456,23 +456,24 @@ public class JvbDoctor
                 return;
             }
 
-            // Sync on start/stop and bridges state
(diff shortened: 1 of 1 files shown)


bgrozev:
What about the `synchronized` block on line 509? It calls `notifyHealthCheckFailed` which in turns calls `eventAdmin.postEvent()` all while holding the lock on `JvbDoctor`?

paweldomas:
> `What about the synchronized block on line 509? It calls notifyHealthCheckFailed which in turns calls eventAdmin.postEvent() all while holding the lock on JvbDoctor?`

`postEvent()` is non  blocking: https://github.com/jitsi/jicoco/blob/master/jicoco/src/main/java/org/jitsi/eventadmin/EventAdmin.java#L33

PR #475, opened 2020-04-20
fix: re-arrange shared thread pools

paweldomas:
The scheduled ExecutorService pool shared by all Jicofo
components had a silly size of 20, so bump it to 200. Any timing out
requests can easily block the pipeline. This pool is used
by components that run things at fixed rate or schedule tasks
with delays.

Adds a cached ExecutorService and make channel allocator threads
use it instead of the scheduled executor pool which has limited size.

### DIFF
--- a/src/main/java/org/jitsi/jicofo/FocusBundleActivator.java
+++ b/src/main/java/org/jitsi/jicofo/FocusBundleActivator.java
@@ -17,8 +17,6 @@
  */
 package org.jitsi.jicofo;
 
-import net.java.sip.communicator.impl.protocol.jabber.caps.*;
-
 import org.jitsi.eventadmin.*;
 import org.jitsi.jicofo.util.*;
 import org.jitsi.service.configuration.*;
@@ -37,9 +35,10 @@ public class FocusBundleActivator
     implements BundleActivator
 {
     /**
-     * The number of threads available in the thread pool shared through OSGi.
+     * The number of threads available in the scheduled executor pool shared
+     * through OSGi.
      */
-    private static final int SHARED_POOL_SIZE = 20;
+    private static final int SHARED_SCHEDULED_POOL_SIZE = 200;
 
     /**
      * OSGi bundle context held by this activator.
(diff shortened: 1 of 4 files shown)


src/main/java/org/jitsi/jicofo/FocusBundleActivator.java
```diff
@@ -169,10 +211,20 @@ public static EventAdmin getEventAdmin()
...
+
+    /**
+     * Returns a cached {@link ExecutorService} shared by all components through
+     * OSGi.
      */
-    public static ScheduledExecutorService getSharedThreadPool()
+    public static ExecutorService getSharedThreadPool()
     {
-        return sharedThreadPool;
+        return ServiceUtils2.getService(bundleContext, ExecutorService.class);
```
bgrozev, line 228: Registering a service with such a generic class seems like a bad idea. It seems like all usage is restricted to this class (`FocusBundleActivator`), so why not bypass OSGi and return a reference to `cachedPool` directly?

The same applies `getSharedScheduledThreadPool()`

bgrozev: Actually, is there any code which uses `getSharedThreadPool`?

paweldomas: https://github.com/jitsi/jicofo/blob/0b0905e34fb5114b04b5263817e11e3de7ed350c/src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java#L926
and
https://github.com/jitsi/jicofo/blob/0b0905e34fb5114b04b5263817e11e3de7ed350c/src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java#L3094
So it's both places that create channel allocators, which  is what the commit msg says: moves channel allocators from bounded size pool to unbounded, cached pool.

paweldomas: > so why not bypass OSGi and return a reference to cachedPool directly?

It's just that there is code that doesn't deal with the `FocusBundleActivator` directly, but retrieves it through OSGi.

paweldomas: I have removed registration of `ExecutorService` as the only usages were going through the bundle activator anyway.

PR #476, opened 2020-04-20
Fail health checks when no bridges are operational.

bgrozev:


bgrozev:
Jicofo will report unhealthy until the briges connect, and then recover (since we have `stickyFailures=false`). That was also the behavior prior to #467

PR #478, opened 2020-04-21
fix(octo): Prevents leaking of OctoEndpoint.

gpolitis:
Cancels any pending octo allocations when the a bridge session is
terminated. This is fixes the following race as reported by @bgrozev:

1. A participant joins
2. We start an OctoChannelAllocator for a new bridge
3. We start a ParticipantChannelAllocator
4. The participant leaves (~150 ms after joining)
5. We cancel the ParticipantChannelAllocator

### DIFF
--- a/src/main/java/org/jitsi/jicofo/AbstractChannelAllocator.java
+++ b/src/main/java/org/jitsi/jicofo/AbstractChannelAllocator.java
@@ -208,9 +208,6 @@ public abstract class AbstractChannelAllocator implements Runnable
         catch (OperationFailedException e)
         {
             logger.error("Failed to invite participant: ", e);
-            // expire any channels that were successfully created, if the invite
-            // failed.
-            bridgeSession.colibriConference.expireChannels(colibriChannels);
         }
     }
 
--- a/src/main/java/org/jitsi/jicofo/AbstractChannelAllocator.java
+++ b/src/main/java/org/jitsi/jicofo/AbstractChannelAllocator.java
@@ -98,6 +98,12 @@ public abstract class AbstractChannelAllocator implements Runnable
     protected final boolean reInvite;
 
     /**
+     * The colibri channels that this allocator has allocated. They'll be
+     * cleaned up if the allocator is canceled or failed at any point.
+     */
+    private ColibriConferenceIQ colibriChannels;
+
+    /**
      * Initializes a new {@link AbstractChannelAllocator} instance which is to
      * invite a specific {@link Participant} into a specific
      * {@link JitsiMeetConferenceImpl} (using a specific jitsi-videobridge
@@ -145,9 +151,15 @@ public abstract class AbstractChannelAllocator implements Runnable
         catch (Throwable e)
         {
             logger.error("Channel allocator failed: ", e);
+            cancel();
         }
         finally
         {
+            if (canceled && colibriChannels != null)
+            {
+                bridgeSession.colibriConference.expireChannels(colibriChannels);
+            }
+
             if (participant != null)
             {
                 participant.channelAllocatorCompleted(this);
@@ -166,15 +178,9 @@ public abstract class AbstractChannelAllocator implements Runnable
             return;
         }
 
-        ColibriConferenceIQ colibriChannels = allocateChannels(offer);
+        colibriChannels = allocateChannels(offer);
         if (canceled)
         {
-            // expire any channels that were successfully created, if the
-            // allocator was canceled while waiting for a reply.
-            if (colibriChannels != null)
-            {
-                bridgeSession.colibriConference.expireChannels(colibriChannels);
-            }
             return;
         }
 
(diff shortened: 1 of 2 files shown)


src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
```diff
@@ -2903,6 +2903,10 @@ public boolean terminate(AbstractParticipant participant,
 
             if (octo)
             {
+                if (participant != null)
+                {
+                    participant.setChannelAllocator(null);
```
paweldomas, line 2908: I'm wondering why is it only in the Octo case?

gpolitis: We cancel the channel allocator for the non-octo case [here](https://github.com/jitsi/jicofo/blob/e8d68b021ea84ec18ebe09770d9f518b1e1ed4ec/src/main/java/org/jitsi/jicofo/Participant.java#L530).

bgrozev: I can't follow all the logic, but I think the leak still exists: if the octo participant is terminated while the channel allocator is running, the channels it eventually allocates will not be released: https://github.com/jitsi/jicofo/blob/master/src/main/java/org/jitsi/jicofo/AbstractChannelAllocator.java#L169

paweldomas: is it that the octo channels never expire?

paweldomas: what  I mean is that for regular participants they eventually expire and that was not a big deal

bgrozev: Yes, currently the octo channels/endpoints on the bridge never expire. We should definitely fix that. 
Do you think if we expire them on the bridge, then we can leave jicofo as it is? If so, do we still need this PR?

paweldomas: either that or extra step that would expire channels if the allocator is canceled

paweldomas: in general I think it's better to expire them explicitly in jicofo even fi the bridge will be smar  enough  to expire

src/main/java/org/jitsi/jicofo/AbstractChannelAllocator.java
```diff
@@ -199,6 +208,9 @@ private void doRun()
         catch (OperationFailedException e)
         {
             logger.error("Failed to invite participant: ", e);
+            // expire any channels that were successfully created, if the invite
+            // failed.
+            bridgeSession.colibriConference.expireChannels(colibriChannels);
```
paweldomas, line 213: I think ParticipantChannel allocator has some logic about expiring channels if invite fails, it calls to `meetConference.onInviteFailed(this);` maybe it's a good time to have it unified? I think channel allocator is a better place to do it

paweldomas: actuallyI'm not sure where to do, but it would be good to have it in one place that would cover both and regular participants

bgrozev: We don't need this code, because `OctoChannelAllocator.invite()` is a no-op. 

I agree it would be nice to centralize error handling, but changing this code worries me because the side effects are so hard to follow. Perhaps the `ChannelAllocator` can check its status after `doRun` completes and clean up (expire channels) if necessary.

gpolitis: You're right @paweldomas. This doesn't apply to Octo because there's no invite step, so I've removed the redundant code.

gpolitis: @bgrozev I'm sorry I top-posted and resolved the conversation, I missed your latest message.

> Perhaps the ChannelAllocator can check its status after doRun completes and clean up (expire channels) if necessary.

I like this idea, I'll push a commit that implements this.

src/main/java/org/jitsi/jicofo/AbstractChannelAllocator.java
```diff
@@ -145,9 +151,15 @@ public void run()
         catch (Throwable e)
         {
             logger.error("Channel allocator failed: ", e);
+            cancel();
         }
         finally
         {
+            if (canceled && colibriChannels != null)
+            {
+                bridgeSession.colibriConference.expireChannels(colibriChannels);
```
bgrozev, line 160: There's a path that expires the channels:
https://github.com/jitsi/jicofo/blob/master/src/main/java/org/jitsi/jicofo/ParticipantChannelAllocator.java#L204

I don't think `colibriConference.expireChannels` is idempotent (I'm pretty sure it will at least send another colibri request). Have you tested this case?

gpolitis: I haven't tested this case but the bridge handles it gracefully [here](https://github.com/jitsi/jitsi-videobridge/blob/06b3dfe5d53306b6959c7bfb65a53d7fc4c0bf56/src/main/java/org/jitsi/videobridge/shim/ContentShim.java#L291), as it should.

Issue #477, opened 2020-04-21
Better distribution over videobridges in case of restart


krombel:
## Description
In case of jicofo restart it happens that nearly all conferences are pushed to the same videobridge. The update ist to slow to take all videobridges into account.

## Current behavior
After jicofo restart all existing conferences are distributed on one (sometimes two) videobridge(s)

## Expected Behavior
After jicofo restart it should be distributed more equal over all videobridges

## Possible Solution
In case of assigning a conference on a videobridge increase the internal count of assigned conferences (or sth. similar) to not count this videobridge as least loaded anymore.
It would be overwritten by incoming stats reports either way. It would be just to handle this case better. It seems that the stats update is too slow to handle 50 requests in one second which cause conference=>bridge assignments.

## Steps to reproduce
- Have 50 Conferences distributed over 10 videobridges
- restart jicofo
- have all conferences on one (sometimes two) videobridges

# Environment details
latest jicofo with several videobridges

bgrozev:
Jicofo counts newly added participants (in the last 10 seconds) with a certain weight towards the load of a bridge. You can control the weight using this property:
```
org.jitsi.jicofo.BridgeSelector.AVG_PARTICIPANT_PACKET_RATE
```

The default is 500.

krombel:
I don't think this is making any difference in that situation.
More details in my comment on #480

bgrozev:
Fixed in #480/#530

PR #480, opened 2020-04-21
Add expectedAdditionalChannels

krombel:
This is my first try to implement my idea to enhance #477.

I don't know if [videoChannelsRate](https://github.com/krombel/jicofo/blob/4e5d8afd99dc134061e8ac75bc11c48705a34e38/src/main/java/org/jitsi/jicofo/bridge/Bridge.java#L112) would might be used for tracking sth similar, but as I don't know the mechanics of that I thought, it might be better/simpler to track this channels independently.

So after a restart - before the stats data received jicofo - this will cause some round-robin for the next unused bridge. And after that the "normal" behaviour as before takes place again.

### DIFF
--- a/src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
+++ b/src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
@@ -907,6 +907,7 @@ public class JitsiMeetConferenceImpl
             }
 
             bridgeSession.participants.add(participant);
+            bridgeSession.bridge.expectAnotherParticipant();
             participant.setBridgeSession(bridgeSession);
             logger.info("Added participant jid= " + participant.getMucJid()
                             + ", bridge=" + bridgeSession.bridge.getJid());
--- a/src/main/java/org/jitsi/jicofo/bridge/Bridge.java
+++ b/src/main/java/org/jitsi/jicofo/bridge/Bridge.java
@@ -300,6 +300,10 @@ public class Bridge
         return Double.compare(this.getStress(), o.getStress());
     }
 
+    public void expectAnotherParticipant() {
+        onVideoChannelsChanged(1);
+    }
+
     void onVideoChannelsChanged(Integer diff)
     {
         if (diff == null)


bgrozev:
Thanks for the patch. 

The `videoChannelsRate` is meant to accomplish the same result. It tracks the number of channels added in the last 10 seconds, instead of the number of channels addded  since the last time we received updated stats from the bridge. Do you find that it doesn't work as expected? Have you tried changing the `AVG_PARTICIPANT_PACKET_RATE` constant?

krombel:
Can you then probably explain how this value is updated?

My research made me assume it is updated an BridgeEvent [here](https://github.com/jitsi/jicofo/blob/36dc201/src/main/java/org/jitsi/jicofo/bridge/BridgeSelector.java#L613), which is created [here](https://github.com/jitsi/jicofo/blob/36dc201/src/main/java/org/jitsi/jicofo/event/BridgeEvent.java#L130).
I stopped following the path once I saw that the only path seems to get to ColibriConferenceImpl [here](https://github.com/jitsi/jicofo/blob/36dc201/src/main/java/org/jitsi/impl/protocol/xmpp/colibri/ColibriConferenceImpl.java#L1047). This makes me assume it is only updated when there is an event coming in via xmpp. And this is exactly why we are getting issues on restart: It is simply too slow.

What I would expect (or what I propose) is to increase it as mentioned in this PR or post an event e.g. before [this return statemend](https://github.com/jitsi/jicofo/blob/36dc201/src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java#L837).

As hopefully got clear: The stress level of a bridge needs to be increased before the next request for a new bridge is handled (at most a few until requests until an updated value is taken into consideration)

Note: As I expect bridge initialization to be more expensive than just running a conference (in terms of videobridge) I am OK with assuming two streams for one participant for the short amount of time between "report via colibri, that value increased" and "timeout of internal increase".

bgrozev:
> This makes me assume it is only updated when there is an event coming in via xmpp.

No, it's updated before we send an allocation request over XMPP. There's no network delay, but it does [execute on another thread](https://github.com/jitsi/jicofo/blob/36dc201/src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java#L926). 

If you'd like to rip out the `EventAdmin` code and replace it with a direct call to `bridge.onVideoChannelsChanged(1)` [somewhere around here]([execute on another thread](https://github.com/jitsi/jicofo/blob/36dc201/src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java#L909)) that would be welcome! I don't even think we need to worry about decrementing the counter when participants are removed.

krombel:
Sorry for responding just now. I did not find time to change it as you mentioned.
Either way: Today we had some networking issues (as you probably had noticed as well) so we decided to update in a hurry. As I was confident in my approach we therefore updated to the build I built last week.
TL;DR> This is now live on our production environment. Distribution over the video bridges happened really smooth with this.
You are able to have a look [on our public stats page](https://stats.ffmuc.net/d/VZoq4FjZk/meet-stats?orgId=1&from=1588258396803&to=1588259378622)

krombel:
Instead of just merging this PR the same approach got implemented in 11810444efc08b1daa5c6408013647d301fb90e5...
This PR is therefore obsolete

bgrozev:
Hey @krombel, my apologies for the negative experience (I just saw your thread on twitter, I'm not active there). I didn't realize it at the time, but I should have communicated better.

This is what happened from my perspective. I didn't and still don't see this PR as solving an issue, because as I wrote earlier the counter was updated before we send a request, not after we receive a response. The first version of your PR is not correct. It adds a second counter, duplicating what `videoChannelsRate` does, and fails to ever decrement it. 

On April 22 I suggested that you replace our existing code with a direct call to `bridge.onVideoChannelsChanged(1)` in `JitsiMeetConferenceImpl` . You replied that you didn't have time to do it, and later pushed another version that implements a small part of what I suggested. It is still incorrect. It doesn't remove the existing code, leading to the same counter being incremented twice. 

This double counting is probably why you see a difference in your production system, and I as wrote earlier the same can be achieved by changing a constant.

When I worked on this code later, I followed my own suggestion and replaced the `EventAdmin` code with a call to `bridge.onVideoChannelsChanged(1)`. The commit you cite is [one small part of that](https://github.com/jitsi/jicofo/pull/530). The other relevant commit is [this](https://github.com/jitsi/jicofo/pull/530/commits/d84038698b766b402b85ae7541453d60c4ae0ce5).

PR #481, opened 2020-04-21
Use code blocks for two conf snippet in the README

raphj:


raphj:
I have signed the CLA in the past :-)

PR #483, opened 2020-04-21
feat(JibriDetector): available jibris count

paweldomas:


src/main/java/org/jitsi/jicofo/recording/jibri/JibriDetector.java
```diff
@@ -191,7 +191,11 @@ private void notifyJibriStatus(Jid jibriJid, boolean available)
     public JSONObject getStats()
     {
         JSONObject stats = new JSONObject();
-        stats.put("count", getInstanceCount());
+        stats.put("count", getInstanceCount(null));
+        stats.put(
+            "available",
+            getInstanceCount(
+                    brewInstance -> brewInstance.status.isAvailable()));
```
bgrozev, line 198: I see we have null checks for  `status` in most places it's used. Are  you sure it's not  necessary?

paweldomas: no, Im not, I will add that

PR #491, opened 2020-04-24
Propagate sources upon ice restart + some.

gpolitis:
@JonathanLennox discovered an issue with Octo that could lead to broken rendering of Octo participants because we're not propagating the sources/source groups of a re-invited participants. That's fixed in the first commit.

In the second commit I remove sources/groups from octo, when a participant is terminated. I can't think of any side-effects if we don't do that, but it seems like a good idea to maintain a clean Octo state.

### DIFF
--- a/src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
+++ b/src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
@@ -1327,6 +1327,18 @@ public class JitsiMeetConferenceImpl
 
         if (bridgeSession != null)
         {
+            MediaSourceMap removedSources = participant.getSourcesCopy();
+            MediaSourceGroupMap removedGroups = participant.getSourceGroupsCopy();
+
+            synchronized (bridges)
+            {
+                operationalBridges()
+                    .filter(bridge -> !bridge.equals(bridgeSession))
+                    .forEach(
+                        bridge -> bridge.removeSourcesFromOcto(
+                            removedSources, removedGroups));
+            }
+
             maybeExpireBridgeSession(bridgeSession);
         }
     }
--- a/src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
+++ b/src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
@@ -1800,6 +1800,15 @@ public class JitsiMeetConferenceImpl
                 errorMsg).build();
         }
 
+        JingleSession jingleSessionCopy = participant.getJingleSession();
+        if (jingleSessionCopy != null && jingleSession != jingleSessionCopy)
+        {
+            //FIXME: we should reject it ?
+            logger.error(
+                "Reassigning jingle session for participant: "
+                    + participantJid);
+        }
+
         // XXX We will be acting on the received session-accept bellow.
         // Unfortunately, we may have not received an acknowledgment of our
         // session-initiate yet and whatever we do bellow will be torn down when
@@ -1808,6 +1817,7 @@ public class JitsiMeetConferenceImpl
         // occurs, we may as well ignore the timeout.
         jingleSession.setAccepted(true);
 
+        participant.setJingleSession(jingleSession);
 
         // Extract and store various session information in the Participant
         participant.setRTPDescription(contents);
--- a/src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
+++ b/src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
@@ -1463,7 +1463,7 @@ public class JitsiMeetConferenceImpl
     {
         logger.info("Got session-accept from: " + jingleSession.getAddress());
 
-        return updateParticipant(jingleSession, answer);
+        return onSessionAcceptInternal(jingleSession, answer);
     }
 
     /**
@@ -1652,7 +1652,10 @@ public class JitsiMeetConferenceImpl
     {
         logger.info("Got transport-accept from: " + jingleSession.getAddress());
 
-        return updateParticipant(jingleSession, contents);
+        // We basically do the same processing as with session-accept by just
+        // forwarding transport/rtp information to the bridge + propagate the
+        // participants sources & source groups to remote bridges.
+        return onSessionAcceptInternal(jingleSession, contents);
     }
 
     /**
@@ -1783,7 +1786,7 @@ public class JitsiMeetConferenceImpl
      * @param contents
      * @return
      */
-    private XMPPError updateParticipant(
+    private XMPPError onSessionAcceptInternal(
         JingleSession jingleSession,
         List<ContentPacketExtension> contents)
     {
--- a/src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
+++ b/src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
@@ -1825,14 +1825,12 @@ public class JitsiMeetConferenceImpl
         if (sourcesAdvertised.isEmpty()
             && globalConfig.injectSsrcForRecvOnlyEndpoints)
         {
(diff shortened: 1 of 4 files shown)


src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
```diff
@@ -1768,15 +1650,9 @@ public XMPPError onTransportAccept(
...
-        jingleSession.setAccepted(true);
-
         logger.info("Got transport-accept from: " + jingleSession.getAddress());
 
-        // We basically do the same processing as with transport-info by just
-        // forwarding transport/rtp information to the bridge
-        onTransportInfo(jingleSession, contents);
-
-        return null;
+        return updateParticipant(jingleSession, contents);
```
JonathanLennox, line 1655: Should this run always, or only if the bridge has changed?

src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
```diff
@@ -1898,6 +1774,146 @@ public XMPPError onRemoveSource(JingleSession sourceJingleSession,
...
+            = MediaSourceMap.getSourcesFromContent(contents);
+        MediaSourceGroupMap sourceGroupsAdvertised
+            = MediaSourceGroupMap.getSourceGroupsForContents(contents);
+        if (sourcesAdvertised.isEmpty()
+            && globalConfig.injectSsrcForRecvOnlyEndpoints)
+        {
+            // We inject an SSRC in order to insure that the participant has
+            // at least one SSRC advertised. Otherwise, non-local bridges in the
+            // conference will not be aware of the participant. We intentionally
+            // use a negative value, because this is an invalid SSRC and will
```
JonathanLennox, line 1826: Do we? It doesn't look like it. (I know this is an old, moved comment...)

src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
```diff
@@ -1898,6 +1774,146 @@ public XMPPError onRemoveSource(JingleSession sourceJingleSession,
...
+        if (participant == null)
+        {
+            String errorMsg
+                = "No participant found for: " + participantJid;
+            logger.warn(errorMsg);
+            return XMPPError.from(XMPPError.Condition.item_not_found,
+                errorMsg).build();
+        }
+
+        // XXX We will be acting on the received session-accept bellow.
```
JonathanLennox, line 1803: This comment is now out-of-date since this is not just used for session-accept.  Also, spelling: "bellow" should be "below".

src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
```diff
@@ -1898,6 +1774,146 @@ public XMPPError onRemoveSource(JingleSession sourceJingleSession,
...
+        // Unfortunately, we may have not received an acknowledgment of our
+        // session-initiate yet and whatever we do bellow will be torn down when
```
JonathanLennox, line 1805: Likewise.

src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
```diff
@@ -1898,6 +1774,146 @@ public XMPPError onRemoveSource(JingleSession sourceJingleSession,
...
+        // the acknowledgement timeout occurs later on. Since we will have
+        // acted on the session-accept by the time the acknowledgement timeout
+        // occurs, we may as well ignore the timeout.
+        jingleSession.setAccepted(true);
```
JonathanLennox, line 1809: I don't understand much about jicofo state but is this really correct to do for transport-accept?

paweldomas: that accepted flag is cleared on ice restart, and transport-accept is equivalent of 'session-accept' even though it may make no sense  from Jingle spec perspective. With 'transport-accept' the client will include the latest source description, so that's why the participant needs  to be updated. It wasn't a problem without Octo(I think), because those are fresh channels and jicofo makes sure that the previous ones have been expired by doing sync expire(due to out of order xmpp processing on the jvb).

src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
paweldomas: always I think

src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
```diff
@@ -1768,15 +1638,9 @@ public XMPPError onTransportAccept(
         JingleSession jingleSession,
         List<ContentPacketExtension> contents)
     {
-        jingleSession.setAccepted(true);
-
         logger.info("Got transport-accept from: " + jingleSession.getAddress());
 
-        // We basically do the same processing as with transport-info by just
-        // forwarding transport/rtp information to the bridge
-        onTransportInfo(jingleSession, contents);
```
bgrozev, line 1777: Do we update the participant's local bridge anywhere now?

src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
gpolitis: The code seems to, indeed, inject an ssrc in the advertised sources. The comment indicates that this is necessary for octo operation. It feels risky and unnecessary to change this behavior without deeper understanding or without the original authors "blessing"?

src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
gpolitis: `onTransportInfo` seems like a strict subset of `onSessionAccept` (if you unwrap `updateColibriChannels`).

src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
gpolitis: Is it out-of-date? It was written back in 2015 in c975f1a and it seems the dance is still present in the code (if you look at another comment in today's code [here](https://github.com/jitsi/jicofo/blob/2696bb4c901682b8fcb8670fccc75436634a2c55/src/main/java/org/jitsi/protocol/xmpp/AbstractOperationSetJingle.java#L229)).

src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
JonathanLennox: Sorry, the "do we" comment was in reference to the "negative value" - I don't think this part is right?

src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
JonathanLennox: I meant that it was saying "on the received session-accept" whereas this now applies to both session-accept and transport-accept.

JonathanLennox: (But you fixed it.)

src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
gpolitis: @bgrozev shall we remove the reference to the "negative value" or update the code to reflect the behavior described in the comment?

bgrozev: I don't remember why I wrote that, but my best guess is that I tried using a negative value, something broke, and then I changed the code but forgot to fix my comment. Let's keep the code and update the comment.

PR #495, opened 2020-04-29
Fix npe

paweldomas:


gpolitis:
Is this about a participant announcing his/her sources but there are no bridge channels, so we skip the update? When/how can it happen? I'm wondering if this can potentially lead to broken video.

paweldomas:
> Is this about a participant announcing his/her sources but there are no bridge channels, so we skip the update? When/how can it happen? I'm wondering if this can potentially lead to broken video.

I  will have to go through the "paths" to figure this out...

The other commit probably is safe to go in, because it's just ad-hoc mute operation. So I will split it into another PR? I have actually seen this one much more often.

paweldomas:
https://github.com/jitsi/jicofo/pull/496

PR #497, opened 2020-04-30
Secure domain + Coturn - Need adapted instructions

Commifreak:
Since I enabled LDAP auth, the TURN/TURNS did not work. I came across this post: https://community.jitsi.org/t/cant-get-jitsi-meet-to-work-with-turn-solved/32300/14?u=webcf and this did the trick.

This PR adds this information.

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,7 +102,7 @@ from anonymous domain. Here's what has to be configured:
  ```
 (Note that guest.jitsi-meet.example.com is internal to jitsi, and you do not need to (and should not) create a DNS record for it, or generate an SSL/TLS certificate, or do any web server configuration.)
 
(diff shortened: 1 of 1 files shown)


Commifreak:
> 
> 
> Hi, thanks for your contribution!
> If you haven't already done so, could you please make sure you sign our CLA (https://jitsi.org/icla for individuals and https://jitsi.org/ccla for corporations)? We would unfortunately be unable to merge your patch unless we have that piece :(.

Done.

damencho:
Can you slightly modify and make it more generic. Like: make sure you enable desired modules to the guest virtual host like `turncredentials ` and `conference_duration ` ...

Commifreak:
Tried it.

But turncredetiansl should be enabled by default in the 1st example, since Jitsi installs coturn now by default, right?

PR #499, opened 2020-04-30
ref(Health):  auto enable on first REST request

paweldomas:
...if org.jitsi.jicofo.health.ENABLE_HEALTH_CHECKS
is not set. This also preserves the previous behavior
from before the most recent health checks refactor.

Also fixes an issue where the health checks would return
500 error instead of always success if not enabled what
was written in the docs.

### DIFF
--- a/src/main/java/org/jitsi/jicofo/health/Health.java
+++ b/src/main/java/org/jitsi/jicofo/health/Health.java
@@ -69,6 +69,12 @@ public class Health
      */
     private static final String ROOM_NAME_PREFIX = "__jicofo-health-check";
 
+    /**
+     * Config property to control health checks:
+     * <li>true - enabled</li>
+     * <li>false - disabled</li>
+     * <li>(null - not set) - auto enable on first REST request(default)</li>
+     */
     private static final String ENABLE_HEALTH_CHECKS_PNAME
             = "org.jitsi.jicofo.health.ENABLE_HEALTH_CHECKS";
 
@@ -79,9 +85,10 @@ public class Health
 
     /**
      * Whether internal health checks are enabled. If not enabled, the rest
-     * API will always return 200.
+     * API will always return 500 error. If {@link #ENABLE_HEALTH_CHECKS_PNAME}
+     * is not set then the first request will auto enable them.
      */
-    private boolean enabled = false;
+    private Boolean enabled = null;
 
     /**
      * FIXME: Temporary override for max health check duration.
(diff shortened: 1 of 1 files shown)


bbaldino:
This seems kinda weird...if we're just going to automatically turn them on, what's the point of the configuration value at all?

paweldomas:
> This seems kinda weird...if we're just going to automatically turn them on, what's the point of the configuration value at all?

To be able to disable.

paweldomas:
> This seems kinda weird...if we're just going to automatically turn them on, what's the point of the configuration value at all?

We use health checks by default, but the community complains about those checks enabled by default. In order to have less config properties I want to do it this way.

bbaldino:
> > This seems kinda weird...if we're just going to automatically turn them on, what's the point of the configuration value at all?
> 
> To be able to disable.

But they aren't really disabled, because a request turns them on?

paweldomas:
@bbaldino 
It may seem weird, but I want to have as little config as possible.

bbaldino:
Ok, I guess I don't have all the context, but it seems like the health checks should just either be enabled or not.  If community doesn't want them, then they can just disable them in the config? I don't follow why we need the config value, but also this 'auto enable' thing if a request is made.

paweldomas:
> Ok, I guess I don't have all the context, but it seems like the health checks should just either be enabled or not. If community doesn't want them, then they can just disable them in the config? I don't follow why we need the config value, but also this 'auto enable' thing if a request is made.

What I said above, to have less config properties to set.

paweldomas:
> then they can just disable them in the config?
Well they can, but it was a recent change in the behavior where they didn't have to do that.

bbaldino:
> > Ok, I guess I don't have all the context, but it seems like the health checks should just either be enabled or not. If community doesn't want them, then they can just disable them in the config? I don't follow why we need the config value, but also this 'auto enable' thing if a request is made.
> 
> What I said above, to have less config properties to set.

I'm not sure I understand why we want to avoid this, as config properties aren't a big deal (it'd be on by default and disable-able by the community) and this is exactly the sort of thing config values are useful for?

Anyway, it's done now, but this seems like an appropriate use of config to me.  Changing this value on upgrades doesn't seem like the worst "breakage" to me, either.  But, again, it's done so I can drop it.

paweldomas:
> I'm not sure I understand why we want to avoid this, as config properties aren't a big deal (it'd be on by default and disable-able by the community) and this is exactly the sort of thing config values are useful for?

Didn't you say or agree with me recently that our configs are too complex :P ? It's just part of trying to reduce the amount of configs required. Try to have the default code behavior match the most use cases.

paweldomas:
Will it not be good if we get less questions on the community forum that people's configs are suddenly broken?

PR #500, opened 2020-04-30
Fix NPE while parsing response on getting conference info

herasymchuk:
This is to fix behavior described at: https://community.jitsi.org/t/jitsi-reservation-system-conflict-workflow/20548
When reservation API returns HTTP 409, Jicofo fetches info about conflicting conference and NPE occurs on parsing response from GET /conference


bgrozev:
The reservation system was moved to a prosody module.

PR #509, opened 2020-05-04
fix: Remove a participant's sources from octo when terminating.

bgrozev:


src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
```diff
@@ -2620,7 +2620,27 @@ private void reInviteParticipants(List<Participant> participants)
         {
             for (Participant participant : participants)
             {
+                BridgeSession session = participant.getBridgeSession();
```
bbaldino, line 2623: Given a meeting with bridges A, B, and C with a participant 1 on bridge A: if participant 1 moves from bridge A to bridge B, would bridge C have previously seen no change but will now see participant 1 removed as an Octo participant and then re-added as an Octo participant?

bgrozev: Yes, as far as I can see this is exactly right.

PR #510, opened 2020-05-05
fix: Fix disabled health checks.

bgrozev:


paweldomas:
We can't go with having health  checks  disabled and returning 200. People will think they are doing checks when actually they are not. Especially that it's a silent change of behavior.

PR #515, opened 2020-05-06
remove pre-spec bundle signaling

fippo:
removes the legacy
  http://estos.de/ns/bundle
version of signaling bundle. This has been superseeded by
  https://xmpp.org/extensions/xep-0338.html


fippo:
where does `BundlePacketExtension` come from? I like removing code ;-)

bgrozev:
jenkins add to whitelist

bgrozev:
https://github.com/jitsi/jitsi-xmpp-extensions/blob/master/src/main/java/org/jitsi/xmpp/extensions/jitsimeet/BundlePacketExtension.java

PR #516, opened 2020-05-06
Require bundle

bbaldino:
Non-bundle support was deprecated in JVB 2, so we no longer support clients doing anything short of max-bundle.  These changes throw an exception if a participant does not list bundle among its advertised features, and assumes all clients are using bundle everywhere else (which meant removing many flags/conditionals to distinguish between bundle and non-bundle cases).

### DIFF
--- a/src/main/java/org/jitsi/impl/protocol/xmpp/colibri/ColibriConferenceImpl.java
+++ b/src/main/java/org/jitsi/impl/protocol/xmpp/colibri/ColibriConferenceImpl.java
@@ -17,6 +17,7 @@
  */
 package org.jitsi.impl.protocol.xmpp.colibri;
 
+import org.jetbrains.annotations.*;
 import org.jitsi.protocol.xmpp.colibri.exception.*;
 import org.jitsi.xmpp.extensions.colibri.*;
 import org.jitsi.xmpp.extensions.jingle.*;
@@ -919,8 +920,7 @@ public class ColibriConferenceImpl
             Map<String, RtpDescriptionPacketExtension> descriptionMap,
             MediaSourceMap sources,
             MediaSourceGroupMap sourceGroups,
-            IceUdpTransportPacketExtension bundleTransport,
-            Map<String, IceUdpTransportPacketExtension> transportMap,
+            @NotNull IceUdpTransportPacketExtension bundleTransport,
             String endpointId,
             List<String> relays)
     {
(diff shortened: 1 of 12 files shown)


bbaldino:
As part of looking into the transition to new Jicofo <-> JVB signaling, I came across this and wanted to yank it out in the interest of removing as much code which can get in the way as possible :)

src/main/java/org/jitsi/jicofo/Participant.java
```diff
@@ -284,6 +275,9 @@ public void setSupportedFeatures(List<String> supportedFeatures)
     {
         this.supportedFeatures
             = Objects.requireNonNull(supportedFeatures, "supportedFeatures");
+        if (!hasBundleSupport()) {
+            throw new IllegalStateException("Participant doesn't support bundle, which is required");
```
paweldomas, line 279: is this exception handled anywhere?

bbaldino: No, the idea was to cause a failure...I originally had it as a log but thought it would be better to fail here to catch anything.  Do you think I should change it back to a log?

paweldomas: without  reading the entire PR I think 'hasBundleSupport()' should be called before starting the process of inviting a participant and  log an error or something

bbaldino: My thinking was that we need to validate this after hearing back from the client, since we want to validate the client told us it supports bundle?

paweldomas: can you just try to throw this exception there regardless and see what happens? if jicofo would catch it gracefully somewhere and log then it's fine, otherwise it should be handled gracefully by checking and logging I think

src/main/java/org/jitsi/impl/protocol/xmpp/colibri/ColibriConferenceImpl.java
```diff
@@ -919,8 +920,7 @@ public void updateChannelsInfo(
             Map<String, RtpDescriptionPacketExtension> descriptionMap,
             MediaSourceMap sources,
             MediaSourceGroupMap sourceGroups,
-            IceUdpTransportPacketExtension bundleTransport,
-            Map<String, IceUdpTransportPacketExtension> transportMap,
+            @NotNull IceUdpTransportPacketExtension bundleTransport,
```
bgrozev, line 887: The octo channel allocator passes `null` here

bbaldino: 🤦 thanks.  Guess we don't have any Octo tests for this.

src/main/java/org/jitsi/jicofo/AbstractChannelAllocator.java
```diff
@@ -172,7 +172,15 @@ private void doRun()
...
         List<ContentPacketExtension> offer;
 
-        offer = createOffer();
+        try
+        {
+            offer = createOffer();
+        }
+        catch (UnsupportedFeatureConfigurationException e)
+        {
+            logger.error("Error creating offer", e);
```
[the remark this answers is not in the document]
bbaldino: I don't think we've actually allocated channels yet, `allocateChannels` is below here.

PR #518, opened 2020-05-07
Update README.md

mimugmail:
https://community.jitsi.org/t/how-to-use-jwt-token-in-jitsi-external-api/18321/14

It took me a day or so to find out that new VirtualHost has to be at the end  and not directly after the second one.


mimugmail:
I just signed it :)

luixxiul:
It is right. You don't have to put the block at the end of the file.

bgrozev:

PR #521, opened 2020-05-08
fix: Don't stop checking a bridge when a single check fails.

bgrozev:


bgrozev:
This is an emergency PR. I'm already working on a proper fix.

PR #525, opened 2020-05-13
ref: remove bridge not available error

paweldomas:
Removes bridge not available error added to Jicofo presence. This error is obsolete. Jicofo will start the session when a bridge becomes available.


paweldomas:
ups I deleted that by mistake

paweldomas:
not sure even how

PR #530, opened 2020-05-15
Cleanup

bgrozev:


src/main/java/org/jitsi/jicofo/xmpp/FocusComponent.java
```diff
@@ -258,7 +253,7 @@ public IQ handleIQSetImpl(IQ iq)
                 org.jivesoftware.smack.packet.IQ response
                     = handleConferenceIq((ConferenceIq) smackIq);
 
-                return response != null ? IQUtils.convert(response) : null;
+                return IQUtils.convert(response);
```
paweldomas, line 256: how about the `null` check? IQUtils.convert doesn't seem to be safe in that regard.

src/main/java/org/jitsi/jicofo/Main.java
```diff
@@ -124,6 +127,9 @@ public static FocusComponent getFocusXmppComponent()
     public static void main(String[] args)
         throws ParseException
     {
+        Thread.setDefaultUncaughtExceptionHandler((t, e) ->
```
bbaldino, line 130: What prompted this?

src/main/java/org/jitsi/impl/protocol/xmpp/ChatMemberImpl.java
```diff
@@ -18,10 +18,10 @@
 package org.jitsi.impl.protocol.xmpp;
 
 import net.java.sip.communicator.impl.protocol.jabber.*;
+import org.jitsi.utils.logging.*;
```
paweldomas, line 21: can you check with this change that the XMPP packets logging works as expected? see a comment in logging.properties on how to enable or ping me if you have problems

src/main/java/org/jitsi/jicofo/Main.java
paweldomas: that's what the UtilActivator was doing - logging unhandled exceptions

src/main/java/org/jitsi/jicofo/xmpp/FocusComponent.java
bgrozev: `handleConferenceIq` never returns null (it returns an error IQ if something goes wrong). That was detected by my IDE, and I marked it with `@NotNull`

src/main/java/org/jitsi/jicofo/Main.java
paweldomas: but... I'm not 100% sure if that was only it(mentioning that because Boris reacted to my comment). anyway lets try to get rid of it

bgrozev: I wasn't able to follow all the usages inside jitsi, but if something was broken we would have seen an exception and there were none in my testing and the PR tests.

src/main/java/org/jitsi/impl/protocol/xmpp/ChatMemberImpl.java
bgrozev: Confirmed, XMPP logging works as before

PR #534, opened 2020-05-18
track how long it takes to allocate channels on the jvb

bbaldino:
We want to track how long it takes for the allocate request to complete, as we're looking at a new transport for this path and want to have some data to compare.

@bgrozev I know we talked about doing this for health checks as well, but we have some logic there that [retries](https://github.com/jitsi/jicofo/blob/master/src/main/java/org/jitsi/jicofo/bridge/JvbDoctor.java#L453-L483) and I wasn't sure what we want to do there, so thought I'd at least start with this bit.

### DIFF
--- a/src/main/java/org/jitsi/impl/protocol/xmpp/colibri/ColibriConferenceImpl.java
+++ b/src/main/java/org/jitsi/impl/protocol/xmpp/colibri/ColibriConferenceImpl.java
@@ -970,11 +970,8 @@ public class ColibriConferenceImpl
          * An average of the time it takes to make allocate channel requests
          * to JVB.
          */
-        private final RateStatistics allocateChannelsReqTimes =
-            new RateStatistics((int)Duration.ofMinutes(1).toMillis());
-
-        private final RateStatistics healthCheckReqTimes =
-            new RateStatistics((int)Duration.ofMinutes(1).toMillis());
+        private final MovingAverage<Long> allocateChannelsReqTimes =
+            new MovingAverage<>(Duration.ofMinutes(1));
 
         /**
          * Notify the stats object how long an allocate channels request took
@@ -982,14 +979,14 @@ public class ColibriConferenceImpl
          * @param nanos the time, in nanoseconds
          */
         void allocateChannelsRequestTook(long nanos) {
-            allocateChannelsReqTimes.update((int)nanos, clock.millis());
+            allocateChannelsReqTimes.add(nanos);
         }
 
         @SuppressWarnings("unchecked")
         public JSONObject toJson()
         {
             JSONObject json = new JSONObject();
-            json.put("avg_allocate_channels_req_time_nanos", allocateChannelsReqTimes.getRate());
+            json.put("avg_allocate_channels_req_time_nanos", allocateChannelsReqTimes.get());
 
             return json;
         }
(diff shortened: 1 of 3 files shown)


src/main/java/org/jitsi/impl/protocol/xmpp/colibri/ColibriConferenceImpl.java
```diff
@@ -943,4 +952,46 @@ public void release()
...
+        }
+
+        /**
+         * An average of the time it takes to make allocate channel requests
+         * to JVB.
+         */
+        private final RateStatistics allocateChannelsReqTimes =
+            new RateStatistics((int)Duration.ofMinutes(1).toMillis());
+
+        private final RateStatistics healthCheckReqTimes =
```
bgrozev, line 976: This is not used anywhere?

src/main/java/org/jitsi/impl/protocol/xmpp/colibri/ColibriConferenceImpl.java
```diff
@@ -943,4 +952,46 @@ public void release()
...
+            new RateStatistics((int)Duration.ofMinutes(1).toMillis());
+
+        /**
+         * Notify the stats object how long an allocate channels request took
+         * to execute
+         * @param nanos the time, in nanoseconds
+         */
+        void allocateChannelsRequestTook(long nanos) {
+            allocateChannelsReqTimes.update((int)nanos, clock.millis());
```
bgrozev, line 985: An  `int` can only hold about 4 seconds expressed in nanos. Do we care about nanosecond precision?

bgrozev: `RateStatistics` uses longs for everything  but the `update` API 🤷

src/main/java/org/jitsi/impl/protocol/xmpp/colibri/ColibriConferenceImpl.java
```diff
@@ -943,4 +952,46 @@ public void release()
...
+        }
+
+        @SuppressWarnings("unchecked")
+        public JSONObject toJson()
+        {
+            JSONObject json = new JSONObject();
+            json.put("avg_allocate_channels_req_time_nanos", allocateChannelsReqTimes.getRate());
```
bgrozev, line 992: Hm, actually, `RateStatistics` doesn't do what we need here. This just returns the accumulated values (multiplied with a constant scale). We want the accumulated values divided by number of events.

src/main/java/org/jitsi/impl/protocol/xmpp/colibri/ColibriConferenceImpl.java
bbaldino: Yeah, I had started this but wasn't sure how/if we wanted to measure it because of the retry stuff...I can either remove it or we can come up with an idea of how we want to do the health check tracking (though with something other than `RateStatistics`, as you pointed out).

src/main/java/org/jitsi/impl/protocol/xmpp/colibri/ColibriConferenceImpl.java
bbaldino: 🤦 Yeah, this will need some more thought on what we want to use.

PR #536, opened 2020-05-20
fix: Skips extra presences on join.

damencho:


src/main/java/org/jitsi/jicofo/VersionBroadcaster.java
```diff
@@ -34,14 +34,13 @@
...
 /**
- * The class listens for "focus joined room" and "conference created" events
- * and adds the info about all conference components versions to Jicofo's MUC
+ * Providers info about all conference components versions to Jicofo's MUC
  * presence.
  *
  * @author Pawel Domas
+ * @author Damian Minkov
  */
 public class VersionBroadcaster
```
paweldomas, line 43: Maybe rename the class to `Versions` as it's no longer broadcasting anything?

src/main/java/org/jitsi/jicofo/VersionBroadcaster.java
```diff
@@ -34,14 +34,13 @@
 import java.util.stream.*;
 
 /**
- * The class listens for "focus joined room" and "conference created" events
- * and adds the info about all conference components versions to Jicofo's MUC
+ * Providers info about all conference components versions to Jicofo's MUC
```
paweldomas, line 37: typo in "Providers"

PR #538, opened 2020-05-27
Integrate JVB API

bbaldino:
This relies on the work in https://github.com/jitsi/jitsi-videobridge/pull/1268 and integrates the JVB API with Jicofo.

In this first version, Jicofo looks for values present in JVB's presence that advertise support for the new JVB API.  If those are present, all requests are sent via the JVB API, if not then it falls back to the existing XMPP connection.

Still considering this a WIP, as we may want some other tweaks before merging this (even if we didn't enable it in JVB yet).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -198,6 +198,11 @@
       <artifactId>jersey-media-json-jackson</artifactId>
       <version>${jersey.version}</version>
     </dependency>
+    <dependency>
+      <groupId>${project.groupId}</groupId>
+      <artifactId>jvb-api-client</artifactId>
+      <version>1.0-SNAPSHOT</version>
+    </dependency>
     <!-- jersey relies on this version of javassist or we get hk2
     errors, but other libs (Powermock, at this time) also rely on
     it but an older version, and that version is getting selected
--- a/pom.xml
+++ b/pom.xml
@@ -201,7 +201,7 @@
     <dependency>
       <groupId>${project.groupId}</groupId>
       <artifactId>jvb-api-client</artifactId>
-      <version>1.0-SNAPSHOT</version>
+      <version>2.1-227-g3ce5bf97</version>
     </dependency>
     <!-- jersey relies on this version of javassist or we get hk2
     errors, but other libs (Powermock, at this time) also rely on
--- a/src/main/java/org/jitsi/impl/protocol/xmpp/colibri/ColibriConferenceImpl.java
+++ b/src/main/java/org/jitsi/impl/protocol/xmpp/colibri/ColibriConferenceImpl.java
@@ -472,20 +472,12 @@ public class ColibriConferenceImpl
         Stanza reply;
         if (jvbApi != null)
         {
-            if (logger.isDebugEnabled())
-            {
-                logger.debug("Using JVB API for request");
-            }
             reply = jvbApi.sendIqAndGetReply(iq);
         }
         else
         {
             try
             {
-                if (logger.isDebugEnabled())
-                {
-                    logger.debug("Using XMPP connection for request");
-                }
                 reply = connection.sendPacketAndGetReply(iq);
             }
             catch (OperationFailedException ofe)
@@ -908,6 +900,10 @@ public class ColibriConferenceImpl
     public void setJvbApi(JvbApi jvbApi)
     {
         this.jvbApi = jvbApi;
+        if (logger.isDebugEnabled())
+        {
+            logger.debug("Using JVB API for requests");
+        }
     }
 
     /**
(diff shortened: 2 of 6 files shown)


damencho:
Can you fix the line length and formatting, please.

src/main/java/org/jitsi/impl/protocol/xmpp/colibri/ColibriConferenceImpl.java
```diff
@@ -445,17 +452,73 @@ protected Stanza sendAllocRequest(String endpointId,
...
+    protected Stanza sendAndGetReply(IQ iq)
+        throws ColibriException
+    {
+        Stanza reply;
+        if (jvbApi != null)
+        {
+            if (logger.isDebugEnabled())
+            {
+                logger.debug("Using JVB API for request");
+            }
```
JonathanLennox, line 478: I don't think you'll want this log on every request - is there a way to log it once?

src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
```diff
@@ -734,6 +737,62 @@ private ColibriConference createNewColibriConference(Jid bridgeJid)
...
+                    {
+                        conf.setJvbApi(
+                            new org.jitsi.videobridge.api.client.v1.JvbApi(
+                                bridge.getJvbApiUrl(),
+                                bridge.getJvbApiPort()
+                            )
+                        );
+                        break;
+                    }
+                    default: throw new RuntimeException("Found common JVB " +
```
JonathanLennox, line 785: If the api dependency is bumped such that it implements a new version, but this code isn't updated, won't this fail even if both client and server still support v1?

bbaldino: `maxApiVersion` could probably be named better: it's the calculated `max common supported API version` between the JVB and Jicofo (see line 763).

JonathanLennox: Yeah, but it's the max version supported by the JVB API module being used by this code, which isn't necessarily the max version supported by the code locally here in jicofo, isn't it?

(I.e., consider that the JVB API is updated to support v2 and the version is bumped in the jicofo pom, but the jicofo code proper isn't updated.)

bbaldino: There are 2 pieces:
The version(s) supported by the `jvb-api-client` version that Jicofo is using, and the versions supported by the `jvb-api-server` (in this case, the JVB itself).  The 'server' versions are advertised by the JVB in its presence.  When JVB supports v2, it will advertise support for both v1 and v2, so we'd still settle on v1 here.

bbaldino: Talked offline and I see the case jonathan is referring to...pushed an update to take into account which API versions Jicofo has _implemented_ support for as well.

src/main/java/org/jitsi/impl/protocol/xmpp/colibri/ColibriConferenceImpl.java
bbaldino: Well, it is on debug :) But I could log it in `setJvbApi` instead?

JonathanLennox: I think it'd make sense - if you have debug logs on because you're debugging something, it'll get very noisy.

src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
JonathanLennox: Did you want to actually add the code which uses the intersect?

bbaldino: It's in there...line 773

pom.xml
```diff
@@ -198,6 +198,11 @@
       <artifactId>jersey-media-json-jackson</artifactId>
       <version>${jersey.version}</version>
     </dependency>
+    <dependency>
+      <groupId>${project.groupId}</groupId>
+      <artifactId>jvb-api-client</artifactId>
+      <version>2.1-227-g3ce5bf97</version>
```
JonathanLennox, line 204: Do we want to create a jvb.version property?  Having the jvb-api-client and jitsi-videobridge versions out of sync seems like a recipe for weird errors (even though the latter is just for testing).

bbaldino: Hmm, the flip side of that is that it makes it easier for someone to upgrade the jvb api client without realizing they're doing so.  I think that since the JVB is just used for testing (and has no interaction with the API code right now), it may feel better that they're versions are independent?

JonathanLennox: The issue is that with how maven dependencies work, I think the jvb used for testing will be run with the version of jvb-api-common pulled in by jvb-api-client, not the version the jvb itself was built with.

bbaldino: True, but I don't know that Jicof even starts the OSGi bundles when for the mock bridge, does it?

JonathanLennox: No idea, but it sounds likely that it doesn't.

If you think it's okay, then ok.

bbaldino: Ok, will keep an eye on it.

PR #539, opened 2020-05-27
Refactor Jvb health checks (cleanup only).

bgrozev:


src/main/java/org/jitsi/jicofo/bridge/BridgeSelector.java
```diff
@@ -382,13 +382,6 @@ private void notifyBridgeDown(Bridge bridge)
         eventAdmin.postEvent(BridgeEvent.createBridgeDown(bridge.getJid()));
     }
 
-    private void notifyBridgeOffline(Bridge bridge)
```
bbaldino, line 383: Why don't we need this event anymore?

bgrozev: Because the only place that was using it is `JvbDoctor`, which now gets notified about it via a direct API call.

src/main/java/org/jitsi/jicofo/bridge/BridgeSelector.java
```diff
@@ -233,6 +233,7 @@ synchronized public Bridge addJvbAddress(
         bridges.put(bridgeJid, newBridge);
 
         notifyBridgeUp(newBridge);
+        jvbDoctor.addBridge(newBridge.getJid());
```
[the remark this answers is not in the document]
bgrozev: I don't understand what you mean. Instead of the event we use a method call to notify JvbDoctor of new bridges.

PR #542, opened 2020-06-01
update the relays on the bridge in BridgeSession#setRelays

bbaldino:
We noticed an issue in the following scenario:

1. Bridges A and B have participants in the same conference
1. All the participants in that conference on bridge A leave
1. Bridge B is told that the remote participants are gone, but isn't told that relay (bridge A) is no longer in the call, so it continues forwarding packets there

Part of the reason the above flow fails, is that in `BridgeSession#removeSourcesFromOcto`, the relays we include in the colibri come from the `OctoParticipant`, but the 'removed' relay has not yet been removed from the participant at that point.

This PR adds a call to `updateColibriOctoChannels` to `BridgeSession#setRelays`.  This is where the relays in the `OctoParticipant` are updated, so doing an update at this point does include the proper set of relays.

### DIFF
--- a/src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
+++ b/src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
@@ -3083,6 +3083,7 @@ public class JitsiMeetConferenceImpl
 
             OctoParticipant octoParticipant = getOctoParticipant(remoteRelays);
             octoParticipant.setRelays(remoteRelays);
+            updateColibriOctoChannels(octoParticipant);
         }
 
         /**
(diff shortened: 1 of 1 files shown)


src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
```diff
@@ -3083,6 +3083,7 @@ private void setRelays(List<String> allRelays)
 
             OctoParticipant octoParticipant = getOctoParticipant(remoteRelays);
             octoParticipant.setRelays(remoteRelays);
+            updateColibriOctoChannels(octoParticipant);
```
bgrozev, line 3086: I am not sure that it is safe to call `updateColibriOctoChannels` without checking whether the "session" has been established (i.e. jicofo has learnt the IDs of the colibri channels). The existing code always does that check.

We already have code which will [update the relay IDs if they changed before the session was established](https://github.com/jitsi/jicofo/blob/master/src/main/java/org/jitsi/jicofo/OctoChannelAllocator.java#L165), so I think you can just wrap the code in `if (octoParticipant.isSessionEstablished())` and do nothing if the session is not established.

bbaldino: Cool, I can add that check

PR #544, opened 2020-06-02
debian: don't depend on a specific JRE version

saghul:
This makes it possible to use AdoptOpenJDK builds. This is useful for
running Java 8 on Debian Buster, for example.

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -3,12 +3,12 @@ Section: net
 Priority: extra
 Maintainer: Jitsi Team <dev@jitsi.org>
 Uploaders: Emil Ivov <emcho@jitsi.org>, Damian Minkov <damencho@jitsi.org>, Pawel Domas <pawel.domas@jitsi.org>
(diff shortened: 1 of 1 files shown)


debian/control
```diff
@@ -3,12 +3,12 @@ Section: net
 Priority: extra
 Maintainer: Jitsi Team <dev@jitsi.org>
 Uploaders: Emil Ivov <emcho@jitsi.org>, Damian Minkov <damencho@jitsi.org>, Pawel Domas <pawel.domas@jitsi.org>
-Build-Depends: debhelper, openjdk-8-jre-headless | openjdk-11-jre-headless, maven
+Build-Depends: debhelper, java8-runtime-headless | java8-runtime | java11-runtime-headless | java11-runtime, maven
```
gpolitis, line 6: Doesn't java8-runtime imply `java8-runtime-headless` (so `java8-runtime-headless | java8-runtime` reduces to `java8-runtime-headless`)

saghul: AdoptOpenJDK doesn't provide headless variants.

PR #549, opened 2020-06-12
Do not use bridges with mismatching octo versions.

bgrozev:


src/main/java/org/jitsi/jicofo/bridge/Bridge.java
```diff
@@ -123,6 +120,11 @@ static void setAvgParticipantPacketRatePps(int avgParticipantPacketRatePps)
      */
     private String version = null;
 
+    /**
+     * The version of Octo that this bridge supports.
+     */
+    private int octoVersion = 0;
```
bbaldino, line 126: Should this be `-1` since `0` is a valid version?

bgrozev: I wanted to default to something reasonable.

bbaldino: Ok, I see now.  Can you either update this comment to include "This is set to the default version (0) which will be assumed if the bridge does not explicitly advertise a version). Or, do something like define `DEFAULT_BRIDGE_OCTO_VERSION = 0` and change the block below where we parse the version to be:

```
Integer octoVersion = stats.getValueAsInt("octo_version");
if (octoVersion != null)
{
    this.octoVersion = octoVersion;
}
else
{
    this.octoVersion = DEFAULT_BRIDGE_OCTO_VERSION;
}
```

PR #553, opened 2020-06-16
Use 32bit integers for GID. Warn when Octo is enabled without a jicofo ID being set.

bgrozev:


bgrozev:
Wait until jitsi/jitsi-videobridge#1311 is in

src/main/java/org/jitsi/jicofo/FocusManager.java
```diff
@@ -258,6 +264,20 @@ public void start()
...
                 config.getString(XMPP_DOMAIN_PNAME));
 
+        int jicofoId = config.getInt(JICOFO_SHORT_ID_PNAME, -1);
+        if (jicofoId < 1 || jicofoId > 0xffff)
+        {
+            logger.warn(
+                "Jicofo ID is not set. Configure a valid value [1-65535] by "
+                + "setting " + JICOFO_SHORT_ID_PNAME + ". Future versions "
+                + "will require this for Octo.");
+            this.jicofoId = 0;
```
JonathanLennox, line 274: Is this gonna be spitting out warnings on default-install jitsi-meet Debian instances?  Is there any way to detect the case where the system only has a single Jicofo, so you don't have to worry about this?

src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
```diff
@@ -3246,8 +3256,8 @@ public String toString()
     {
         return
             String.format(
-                    "JitsiMeetConferenceImpl[id=%s, name=%s]",
-                    id,
+                    "JitsiMeetConferenceImpl[gid=%s, name=%s]",
```
JonathanLennox, line 3259: Should the format be %d here now? I know Integer.toString() works, but semantically it seems odd

JonathanLennox:
An error in the description: you say "Use 32bit integers", but actually you're using 16 bit integers.

src/main/java/org/jitsi/jicofo/FocusManager.java
```diff
@@ -258,6 +264,20 @@ public void start()
         DomainBareJid xmppDomain = JidCreate.domainBareFrom(
                 config.getString(XMPP_DOMAIN_PNAME));
 
+        int jicofoId = config.getInt(JICOFO_SHORT_ID_PNAME, -1);
+        if (jicofoId < 1 || jicofoId > 0xffff)
```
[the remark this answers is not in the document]
bgrozev: No longer used in  health checks

src/main/java/org/jitsi/jicofo/FocusManager.java
bgrozev: It will log a single warning at startup, and any time a conference uses Octo.

I can't think of a way to detect when a system has more than one jicofo. We could provision debian/docker with a valid default valid to  suppress the log, but that defeats the purpose of the log -- if you're building a system with Octo and multiple jicofos you will not see the warning.

src/main/java/org/jitsi/jicofo/FocusManager.java
```diff
@@ -474,26 +494,29 @@ public boolean conferenceRequest(
...
                     focusUserName,
                     protocolProviderHandler,
                     jvbProtocolProvider,
                     this, config, globalConfig, logLevel,
                     id, includeInStatistics);
 
-        conferences.put(room, conference);
-        conferenceIds.add(id);
+            conferences.put(room, conference);
+            conferenceGids.add(id);
```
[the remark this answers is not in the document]
bgrozev: I actually had that change and reverted it. The whole `createConference` method always runs with the lock acquired, so it makes no difference to the correctness. And having `generateConferenceId` put it in the map feels wrong -- it's the conference-aware code that is responsible for removing it later on.

src/main/java/org/jitsi/jicofo/FocusManager.java
```diff
@@ -536,23 +559,17 @@ private JitsiMeetConferenceImpl createConference(
...
-        String id;
+        int id;
 
         synchronized (conferencesSyncRoot)
         {
             do
             {
-                id = jicofoShortId +
-                        String.format("%04x", RANDOM.nextInt(0x1_0000));
+                id = (jicofoId << 16) | RANDOM.nextInt(0x1_0000);
```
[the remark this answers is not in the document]
bgrozev: We don't exclude 0xffff now

src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
```diff
@@ -322,18 +324,19 @@ public JitsiMeetConferenceImpl(EntityBareJid            roomName,
...
+    public JitsiMeetConferenceImpl(
+            EntityBareJid roomName,
+            Resourcepart focusUserName,
+            ProtocolProviderHandler protocolProviderHandler,
+            ProtocolProviderHandler jvbXmppConnection,
+            ConferenceListener listener,
+            JitsiMeetConfig config,
+            JitsiMeetGlobalConfig globalConfig,
+            Level logLevel,
+            int gid)
```
bbaldino, line 336: `long` for gid here

PR #555, opened 2020-06-18
Detects correct muc service based on prefix "conference".

damencho:


src/main/java/org/jitsi/jicofo/FocusManager.java
```diff
@@ -289,7 +300,9 @@ public void start()
             = new JitsiMeetServices(
                     protocolProviderHandler,
                     jvbProtocolProvider,
-                    focusUserDomain);
+                    focusUserDomain,
+                    JidCreate.domainBareFrom(
+                        conferenceMucPrefix + "." + xmppDomainConfig));
```
bgrozev, line 305: If we're matching the complete domain, why not get rid of the discovery logic? Just pass the configuration to `Health`.

paweldomas: +1

damencho: Ok, make sense :)

damencho: How about now? 
Now this PR removes more than it adds :)

PR #556, opened 2020-06-18
JingleOfferFactory clean-up.

bgrozev:


src/main/java/org/jitsi/jicofo/OctoChannelAllocator.java
```diff
@@ -81,30 +81,18 @@ public OctoChannelAllocator(
 
         boolean useIce = false;
         boolean useDtls = false;
-        boolean useRtx = false; // config.isRtxEnabled();
-        boolean enableRemb = false; // config.isRembEnabled();
-        boolean enableTcc = false; // config.isTccEnabled();
-        boolean useSctp = false; // config.openSctp() == null || config.openSctp();
```
bbaldino, line 87: Can we get away with removing the code that supports this?  Previously we said we didn't want to remove it from the bridge because of people's deployments.

src/main/java/org/jitsi/jicofo/OctoChannelAllocator.java
```diff
@@ -81,30 +81,18 @@ public OctoChannelAllocator(
 
         boolean useIce = false;
```
bbaldino, line 82: Is this even optional at this point?

bgrozev: Yes, as you see Octo channels don't have ICE :)

src/main/java/org/jitsi/jicofo/OctoChannelAllocator.java
bgrozev: We haven't migrated to websockets everywhere AFAIK

src/main/java/org/jitsi/jicofo/OctoChannelAllocator.java
bbaldino: Ah

src/main/java/org/jitsi/jicofo/OctoChannelAllocator.java
bbaldino: Then how can we disable support for using SCTP?  (Or perhaps I've misunderstood the implication of removing this flag--I assumed it meant we never would do sctp)

JonathanLennox: This is the `OctoChannelAllocator`.  We still conditionally add SCTP for the `ParticipantChannelAllocator`.

bbaldino: Ah! Missed that.

PR #557, opened 2020-06-19
fix(JvbDoctor): connection null on bad timing

paweldomas:
Use getter to fetch the current XMPP connection. This fixes
an issue where the connection is always null if it wasn't
available on bundle startup.

Also reduces logging level on discovery error to warning.

### DIFF
--- a/src/main/java/org/jitsi/impl/protocol/xmpp/OpSetSimpleCapsImpl.java
+++ b/src/main/java/org/jitsi/impl/protocol/xmpp/OpSetSimpleCapsImpl.java
@@ -140,7 +140,7 @@ public class OpSetSimpleCapsImpl
         }
         catch (Exception e)
         {
-            logger.error(
+            logger.warn(
                     String.format(
                             "Failed to discover features for %s: %s", node, e.getMessage()));
         }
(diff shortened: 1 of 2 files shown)


bkempe:
@paweldomas Thanks for the quick fix! BTW, does this indicate a configuration problem?
```
Failed to discover features for focus.ourdomain.com: XMPP error reply received from focus.ourdomain.com: XMPPError: service-unavailable - wait
```

paweldomas:
> @paweldomas Thanks for the quick fix! BTW, does this indicate a configuration problem?
> 
> ```
> Failed to discover features for focus.ourdomain.com: XMPP error reply received from focus.ourdomain.com: XMPPError: service-unavailable - wait
> ```

No, it doesn't. There might be some components that always return this error even when the config is okay and I've seen users complaining that printing error in such case is confusing.

paweldomas:
> ```
> Failed to discover features for focus.ourdomain.com: XMPP error reply received from focus.ourdomain.com: XMPPError: service-unavailable - wait
> ```

So it depends. I suppose this might be logged during the initial startup when Jicofo's user connection has connected already, before the component one. But on the other hand it might show that the focus component connection is not working, but then no conference would work on web. I would not rely on this message in general.

PR #562, opened 2020-07-07
feat: add support for session-terminate

paweldomas:
Implements 'session-terminate' which allows the clients to end the session. If there's the restart attribute set to true in the <bridge-session> extension, then Jicofo will start a new session after tearing down the existing one.

Adds rate limiting on the restart requests to prevent abuse and/or client bugs. There must be at least 10 second gap between the requests and no more than 3 requests are allowed in the last minute.

Depends on https://github.com/jitsi/jitsi-xmpp-extensions/pull/30

### DIFF
--- a/src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
+++ b/src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
@@ -1286,7 +1286,11 @@ public class JitsiMeetConferenceImpl
                 = findParticipantForChatMember(chatRoomMember);
             if (leftParticipant != null)
             {
-                terminateParticipant(leftParticipant, Reason.GONE, null);
+                terminateParticipant(
+                        leftParticipant,
+                        Reason.GONE,
+                        null,
+                        /* no need to send session-terminate - gone */ false);
             }
             else
             {
@@ -1308,8 +1312,15 @@ public class JitsiMeetConferenceImpl
 
     private void terminateParticipant(Participant    participant,
                                       Reason         reason,
-                                      String         message)
+                                      String         message,
+                                      boolean        sendSessionTerminate)
     {
+        logger.info(String.format(
+                "Terminating %s, reason: %s, send st: %s",
+                participant,
+                reason,
+                sendSessionTerminate));
+
         BridgeSession bridgeSession;
         synchronized (participantLock)
         {
@@ -1317,9 +1328,8 @@ public class JitsiMeetConferenceImpl
             if (participant.isSessionEstablished())
             {
                 JingleSession jingleSession = participant.getJingleSession();
-                logger.info("Terminating: " + contactAddress);
 
-                jingle.terminateSession(jingleSession, reason, message, reason != null);
+                jingle.terminateSession(jingleSession, reason, message, sendSessionTerminate);
 
                 removeSources(
                     jingleSession,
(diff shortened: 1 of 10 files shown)


src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
```diff
@@ -1543,6 +1545,72 @@ public XMPPError onSessionInfo(JingleSession session, JingleIQ iq)
...
+    public XMPPError onSessionTerminate(JingleSession session, JingleIQ iq)
+    {
+        Participant participant = findParticipantForJingleSession(session);
+
+        // FIXME: (duplicate) there's very similar logic in onSessionAccept/onSessionInfo
+        if (participant == null)
+        {
+            String errorMsg = "No session for " + session.getAddress();
+
+            logger.warn("onSessionInfo: " + errorMsg);
```
bgrozev, line 1563: Log "onSessionTerminate" or something more useful? 

"No session" is also not accurate -- there's a Jingle session, but no participant associated with it.

src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
```diff
@@ -1543,6 +1545,72 @@ public XMPPError onSessionInfo(JingleSession session, JingleIQ iq)
...
+            return XMPPError.from(XMPPError.Condition.item_not_found, "invalid bridge session ID").build();
+        }
+
+        logger.info(String.format(
+                "Received session-terminate from %s, session: %s, restart: %s",
+                participant,
+                bridgeSession,
+                restartRequested));
+
+        terminateParticipant(participant, null, null);
```
bgrozev, line 1594: Would setting the reason and message to something like "restarting" make sense? It might make debugging easier.

bgrozev: Or rather "requested from the client, restartRequested=X"

src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
```diff
@@ -1543,6 +1545,72 @@ public XMPPError onSessionInfo(JingleSession session, JingleIQ iq)
...
+
+        if (restartRequested)
+        {
+            if (participant.incrementAndCheckRestartRequests())
+            {
+                participants.add(participant);
```
bgrozev, line 1600: I think this line and the line above require `participantLock`

src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
bgrozev: Oh, I see `reason == null` is used as a condition to prevent sending a session-terminate of our own. How about adding a separate parameter?

src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
paweldomas: thanks, that's a good catch

src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
paweldomas: yes, I'll add a separate parameter

src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
paweldomas: yes, thanks

PR #566, opened 2020-08-03
Use metaconfig, add SCTP configuraion.

bgrozev:


pom.xml
```diff
@@ -281,8 +281,25 @@
       <version>${jersey.version}</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>io.kotlintest</groupId>
+      <artifactId>kotlintest-runner-junit5</artifactId>
+      <version>3.4.2</version>
```
bbaldino, line 287: `Kotlintest` changed names to `kotest` and has a newer version, can we use that here?  I plan on upgrading other places soon.

src/main/java/org/jitsi/jicofo/bridge/BridgeSelectionStrategy.java
```diff
@@ -63,9 +62,9 @@
     private int totalSplitDueToLoad;
 
     /**
-     * Maximum participants per bridge in one conference.
+     * Maximum participants per bridge in one conference, or {@code -1} for no maximum.
      */
-    private int maxParticipantsPerBridge = Integer.MAX_VALUE;
+    private int maxParticipantsPerBridge = BridgeConfig.config.maxBridgeParticipants();
```
bbaldino, line 67: Any pattern to where you  do `BridgeConfig.config...` vs statically importing `BridgeConfig.config` and just do `config...` ?

src/main/java/org/jitsi/jicofo/jigasi/JigasiDetector.java
```diff
@@ -50,7 +50,7 @@
     /**
      * The local region of the this jicofo instance.
      */
-    private final String localRegion;
+    private final String localRegion = JicofoConfig.config.localRegion();
```
bbaldino, line 53: Do you think we should get rid of the member and use `JicofoConfig.config.localRegion()` in each place?

src/main/resources/reference.conf
```diff
@@ -0,0 +1,4 @@
+jicofo {
+  # The region in which the machine is running.
```
bbaldino, line 11: I was thinking about a convention where comments in HOCON files use `//` and commented-out properties (for the purpose of examples or whatever) use `#`, to help differentiate them.  Thoughts?

bbaldino: (I did this in Jibri and found that it helped readability, in my opinion)

src/main/java/org/jitsi/jicofo/jigasi/JigasiDetector.java
bgrozev: I don't know, I have to think about it more. My first thought is that, unless the config can change at runtime, I prefer to initialize it once. Being able to change things at runtime would be nice, but it would require a larger effort. I'd leave this change for the time when we decide to do that.

bbaldino: Sure...but that's already implemented in the metaconfig layer (we only try retrieving a value once and then cache the result) if that makes you feel better.

PR #574, opened 2020-08-06
Update codec configuration, add support for opus/red.

bgrozev:


src/main/kotlin/org/jitsi/jicofo/CodecConfig.kt
```diff
@@ -0,0 +1,109 @@
...
+
+    fun enabled() = enabled && pt > 0
+
+    private val pt: Int by config {
+        "$legacyBase.${codecName}_PT".from(JitsiConfig.legacyConfig)
+        "$newBase.pt".from(JitsiConfig.newConfig)
+    }
+
+    @Throws(IllegalStateException::class)
+    fun pt(): Int = if (enabled()) pt else throw IllegalStateException("$codecName is not enabled.")
```
bbaldino: Metaconfig has you covered for these :) You can change pt's definition to:
```
private val pt: Int by config {
    onlyIf("$codecName is enabled", ::enabled) {
        "$legacyBase.${codecName}_PT".from(JitsiConfig.legacyConfig)
        "$newBase.pt".from(JitsiConfig.newConfig)        
    }
}
```

src/main/kotlin/org/jitsi/jicofo/CodecConfig.kt
```diff
@@ -0,0 +1,109 @@
...
+
+    val id: Int by config {
+        "$newBase.id".from(JitsiConfig.newConfig)
+    }
+    fun id() = id
+}
+
+class TccConfig : RtpExtensionConfig("none", "${CodecConfig.EXTENSIONS_BASE}.tcc") {
+    override val enabled: Boolean by config {
+        "$newBase.enabled".from(JitsiConfig.newConfig)
```
bbaldino: I'm confused by this...only TCC doesn't check for a legacy name?

src/main/resources/reference.conf
```diff
@@ -31,32 +31,36 @@ jicofo {
 
     // RTP header extensions
     rtp-extensions {
+      audio-level {
+        enabled=true
+        id=1
+      }
```
JonathanLennox, line 64: Turning off audio levels will break a bunch of bridge behavior -- is this just a case of "you get what you ask for"? Is there a use case for turning them off, other than symmetry among the extensions?

src/main/resources/reference.conf
```diff
@@ -31,32 +31,36 @@ jicofo {
 
     // RTP header extensions
     rtp-extensions {
+      audio-level {
```
JonathanLennox, line 61: You've got the extension a different order in reference.conf than you do in CodecConfig.kt and CodecConfigTest.kt - is there a reason?

src/main/java/org/jitsi/jicofo/codec/JingleOfferFactory.java
```diff
@@ -434,51 +434,75 @@ private static void addAudioToContent(
...
-        opus.setChannels(2);
+        if (config.opus.enabled())
+        {
+            // a=rtpmap:111 opus/48000/2
+            PayloadTypePacketExtension opus = addPayloadTypeExtension(rtpDesc, config.opus.pt(), Constants.OPUS, 48000);
+            // Opus is always signaled with 2 channels, regardless of 'stereo'
+            opus.setChannels(2);
+
+            // Though RED has a payload type of its own and can be used to encode multiple other payload types, we need
+            // it to be advertised with the same clock rate as opus, so its defined here.
```
JonathanLennox, line 445: Grammar, "it's"

src/main/java/org/jitsi/jicofo/codec/JingleOfferFactory.java
```diff
@@ -434,51 +434,75 @@ private static void addAudioToContent(
...
+            if (config.opus.red.enabled())
+            {
+                PayloadTypePacketExtension red = addPayloadTypeExtension(rtpDesc, config.opus.red.pt(), "red", 48000);
+                red.setChannels(2);
+            }
```
JonathanLennox, line 450: Do you want to set the fmtp for the red packet? (Note that the Jingle syntax for the red fmtp is poorly defined...)

src/main/kotlin/org/jitsi/jicofo/codec/CodecConfig.kt
```diff
@@ -77,11 +110,20 @@ private class RtpExtensionConfigWithLegacy(
...
     @JvmField
-    val vp8 = VideoCodecConfig(LEGACY_BASE, "$BASE.vp8", "VP8")
+    val vp8: RtxCodecConfig = RtxCodecConfigWithLegacy(LEGACY_BASE, "$VIDEO_BASE.vp8", "VP8")
+    @JvmField
+    val vp9: RtxCodecConfig = RtxCodecConfigWithLegacy(LEGACY_BASE, "$VIDEO_BASE.vp9", "VP9")
+    @JvmField
+    val h264: RtxCodecConfig = RtxCodecConfigWithLegacy(LEGACY_BASE, "$VIDEO_BASE.h264", "H264")
+
+    @JvmField
+    val icas16 = CodecConfig("$AUDIO_BASE.icas-16000", "icas-16000")
```
JonathanLennox, line 120: Typo: "isac" (and below)

src/main/java/org/jitsi/jicofo/codec/JingleOfferFactory.java
```diff
@@ -434,51 +434,75 @@ private static void addAudioToContent(
...
-        // a=rtpmap:104 ISAC/32000
-        addPayloadTypeExtension(rtpDesc, 104, "ISAC", 32000);
-
-        // rtpmap:126 telephone-event/8000
-        addPayloadTypeExtension(rtpDesc, 126, Constants.TELEPHONE_EVENT, 8000);
+        if (config.telephoneEvent.enabled())
+        {
+            // rtpmap:126 telephone-event/8000
+            addPayloadTypeExtension(rtpDesc, config.telephoneEvent.pt(), Constants.TELEPHONE_EVENT, 8000);
+        }
```
JonathanLennox, line 505: Any desire to add support for any other audio codecs?  Or would that be a separate PR?

src/main/kotlin/org/jitsi/jicofo/CodecConfig.kt
```diff
@@ -0,0 +1,109 @@
...
+    open val enabled: Boolean by config {
+        legacyEnabledName.from(JitsiConfig.legacyConfig)
+        "$newBase.enabled".from(JitsiConfig.newConfig)
+    }
+    fun enabled() = enabled
+
+    val id: Int by config {
+        "$newBase.id".from(JitsiConfig.newConfig)
+    }
+    fun id() = id
```
bbaldino: I see we change this to a String, I wonder is it even worth holding as an `Int`? Do we use it as a number anywhere?

src/main/resources/reference.conf
```diff
@@ -8,6 +8,58 @@ jicofo {
...
+    h264 {
+      enabled=true
+      pt=107
+      // Payload type for the associated RTX stream. Set to -1 to disable RTX.
+      rtx-pt=99
+    }
+
+    // RTP header extensions
+    rtp-extensions {
+      tof {
```
bbaldino, line 65: Do you think it's even worth adding to new config?

src/test/kotlin/org/jitsi/jicofo/CodecConfigTest.kt
```diff
@@ -0,0 +1,242 @@
...
+                }
+                withNewConfig("jicofo { codec { vp8 { rtx-pt=112 } } }") {
+                    config.vp8.enabled() shouldBe true
+                    config.vp8.pt() shouldBe 100
+                    config.vp8.rtxEnabled() shouldBe true
+                    config.vp8.rtxPt() shouldBe 112
+                }
+            }
+            context("Disabling/enabling extensions") {
+                withNewConfig("jicofo { codec { rtp-extensions { framemarking { enabled=false } } } }") {
```
bbaldino, line 197: You can use `.` syntax for these to make these more readable, e.g.:
```
withNewConfig("jicofo.codec.rtp-extensions.framemarking.enabled=false")
```

src/main/kotlin/org/jitsi/jicofo/CodecConfig.kt
bgrozev: There was no legacy config for TCC.

src/main/kotlin/org/jitsi/jicofo/CodecConfig.kt
bgrozev: It's supposed to be number, but the XML extensions (wrongly IMO) expect a string.

src/main/resources/reference.conf
bgrozev: I ordered these by ID, the order in the code was arbitrary

src/main/kotlin/org/jitsi/jicofo/CodecConfig.kt
JonathanLennox: Yeah, if you put something other than a number then the browser's SDP parser will get confused, so better to enforce the restriction here.

src/main/java/org/jitsi/jicofo/codec/JingleOfferFactory.java
bgrozev: I don't think it is necessary, and it's currently not being used. Chrome does not set it in it's offer or answer (and strips it if it's set). It also exposes a bug in jitsi-videobridge (to make the XML format as undestood by lib-jitsi-meet work we need to use name=null, value=111, and jvb saves the parameters mapped by name not allowing null keys). So I'd rather not set it.

src/main/resources/reference.conf
bgrozev: I think "you get what you ask for" describes it well.

src/main/java/org/jitsi/jicofo/codec/JingleOfferFactory.java
bgrozev: That would be separate

src/main/resources/reference.conf
bgrozev: I just added them all, I don't think it hurts

src/main/kotlin/org/jitsi/jicofo/CodecConfig.kt
bgrozev: Not quite, because of the way legacy config disables the codec. I want `pt()` to throw an exception if the `pt` property is set to -1 (which causes the `enabled()` function to return false, while the `enabled` property might be `true`.

src/main/java/org/jitsi/jicofo/codec/JingleOfferFactory.java
bgrozev: Fixed

src/main/kotlin/org/jitsi/jicofo/codec/CodecConfig.kt
bgrozev: Fixed 🤦

src/main/resources/reference.conf
JonathanLennox: It's just that having them in a different order in the two places makes it a bit difficult to visually confirm that everything's covered. Not a big deal though.

src/main/kotlin/org/jitsi/jicofo/CodecConfig.kt
bbaldino: Yeah I meant to refer to use the function `enabled()` in that check, not the property--I hadn't noticed the name would be ambiguous.

bgrozev: The thing is `enabled()` uses the `pt` property: `fun enabled() = enabled && pt > 0`

bbaldino: I think I got lost here, but whichever one you want to use you can use.  I don't know if `::enabled` will resolve to the property or the function, but if it resolves to the property you can do this to use the function:
```
private val pt: Int by config {
    onlyIf("$codecName is enabled", { enabled() }) {
        "$legacyBase.${codecName}_PT".from(JitsiConfig.legacyConfig)
        "$newBase.pt".from(JitsiConfig.newConfig)        
    }
}
```

bgrozev: The problem is that the property is enabled if `enabled()` returns true (so presumably accessing the property it calls the function at some point), but the call to `enabled()` accesses the property.

bbaldino: Ah ha, I see now.  👍

src/main/kotlin/org/jitsi/jicofo/codec/CodecConfig.kt
```diff
@@ -0,0 +1,159 @@
...
+import org.jitsi.config.JitsiConfig
+import org.jitsi.metaconfig.config
+
+open class CodecConfig(protected val base: String, protected val name: String) {
+    protected open val enabled: Boolean by config {
+        "$base.enabled".from(JitsiConfig.newConfig)
+    }
+
+    protected open val pt: Int by config {
+        "$base.pt".from(JitsiConfig.newConfig)
```
bbaldino, line 30: Using `$base` for these properties would make it difficult to grep through the code to find the usage of a property.  I almost made this comment in another PR where you used a `base` prefix, but the key itself was unique enough I figured it would be ok.  Here, though, `pt`, `enabled`, etc. aren't very unique.

PR #579, opened 2020-08-17
change jicofo tests to instantiate videobridge directly (not via bund…

bbaldino:
…le activator)

This adapts to the changes in https://github.com/jitsi/jitsi-videobridge/pull/1388, where the bridge instantiation is moved to a singleton and shimmed into OSGi.  This means that the instance survives across OSGi starts/stops, which isn't an issue in the bridge where we only start/stop things once (at startup/shutdown), but for tests which rely on re-instantiating videobridge it doesn't work.  The mock videobridge in Jicofo no longer needs to use OSGi to create the `Videobridge` instance, it can do so directly (like JVB now does)

### DIFF
--- a/src/test/java/mock/jvb/MockVideobridge.java
+++ b/src/test/java/mock/jvb/MockVideobridge.java
@@ -57,8 +57,6 @@ public class MockVideobridge
 
     private boolean returnServerError = false;
 
-    private VideobridgeBundleActivator jvbActivator;
-
     private ColibriConferenceIqHandler confIqGetHandler
             = new ColibriConferenceIqHandler(IQ.Type.get);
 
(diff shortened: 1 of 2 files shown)


bbaldino:
jenkins, test this please

src/test/java/mock/jvb/MockVideobridge.java
```diff
@@ -207,7 +200,11 @@ public IQ handleIQRequest(IQ iqRequest)
 
     public int getEndpointCount()
     {
+        System.out.println("Bridge has " + bridge.getConferences().size() + " conferences");
```
paweldomas, line 203: why those prints here? it looks like some debugging left by mistake

bbaldino: Oops, you're right. Thanks.

bbaldino:
jenkins, test this please

PR #585, opened 2020-09-07
feat: Passes room bare jid to conference IQ.

damencho:
Depends on https://github.com/jitsi/jitsi-xmpp-extensions/pull/31

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -123,7 +123,7 @@
     <dependency>
       <groupId>${project.groupId}</groupId>
       <artifactId>jitsi-xmpp-extensions</artifactId>
-      <version>1.0-14-gba982f0</version>
+      <version>1.0-15-g813e4c4</version>
     </dependency>
     <dependency>
       <groupId>org.slf4j</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -123,7 +123,7 @@
     <dependency>
       <groupId>${project.groupId}</groupId>
       <artifactId>jitsi-xmpp-extensions</artifactId>
-      <version>1.0-13-ga40f9c3</version>
+      <version>1.0-14-gba982f0</version>
     </dependency>
     <dependency>
       <groupId>org.slf4j</groupId>
--- a/src/main/java/org/jitsi/impl/protocol/xmpp/colibri/ColibriConferenceImpl.java
+++ b/src/main/java/org/jitsi/impl/protocol/xmpp/colibri/ColibriConferenceImpl.java
@@ -109,7 +109,7 @@ public class ColibriConferenceImpl
     /**
      * Flag used to figure out if Colibri conference has been
      * allocated during last
-     * {@link #createColibriChannels(boolean, String, String, boolean, List)}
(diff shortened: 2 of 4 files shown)


ibauersachs:
Why is this necessary? The room name is (or used to be...) part of the JID and cannot (could not) be a simple string.

damencho:
This change is needed because of the multi tenant option of jitsi-meet. And this is the name for the conference ... it's not a jid, and is used only for logging and stats (callstats and siteId).
The conferences can be without tenant `roomName` or with a tenant `someCompany/roomName`. Converted to jid, this would be `roomName@conference.someCompany.jitsiexample.com`. And internally in prosody this is converted and handled as jid `[someCompany]roomName@conference.jitsiexample.com`.
This change is to be able to send that siteId to callstats.
For example before this change if you have a meeting https://meet.jit.si/ingo/roomName and another one https://meet.jit.si/damencho/roomName and if you filter logs by roomName you will get logs for both conferences with no mention of the tenant, which will be improved by this change.

ibauersachs:
It's been a while, but this sounds really weird... wouldn't nginx or whatever is actually creating the rooms in Prosody do the translation/mapping from orgName/confname into confname@orgname.meet.example?

And the change on line 731 might not be what you want if I understand your explanation correctly.

damencho:
> It's been a while, but this sounds really weird... wouldn't nginx or whatever is actually creating the rooms in Prosody do the translation/mapping from orgName/confname into [confname@orgname.meet.example](mailto:confname@orgname.meet.example)?



> It's been a while, but this sounds really weird... wouldn't nginx or whatever is actually creating the rooms in Prosody do the translation/mapping from orgName/confname into [confname@orgname.meet.example](mailto:confname@orgname.meet.example)?

Well yeah its a dynamic config.js based on the URL that is served to the client then a prosody module do the conversion to the existing domain part of the jid.



> And the change on line 731 might not be what you want if I understand your explanation correctly.
Ohm yes, I got confused, yeah we are sending the whole jid ... and leave it to be parsed only in one place https://github.com/jitsi/jitsi-stats/pull/10/files#diff-76beb6d96152820b241eb7dd1f5eb65bR116 which is common for jigasi and jvb ...

ibauersachs:
The changes in jitsi-stats make a lot more sense to me than what is done here. As for the string-fiddling with the domain name there, you could consider using `Name` from dnsjava.

damencho:
Well, this will bring dnsjava to a lot of places as we need to make it a dependency of jitsi-xmpp-extensions where this is just a name ... I will update javadoc of jitsi-xmpp-extensions explaining what is the name and how it is used.

damencho:
And it already says "World readable name for the conference." ...

bgrozev:
> The changes in jitsi-stats make a lot more sense to me than what is done here.

The other option we had was to do the interpretation of "domain", "tenant" and "conference name" in jicofo, then pass it as 3 parameters to the bridge, only for it to re-construct a single string from them again. This also requires the same parsing code in jigasi.

bgrozev (approved):

bgrozev (approved):

PR #592, opened 2020-09-14
Use the new JVB stress level stat

bbaldino:
Depends on https://github.com/jitsi/jitsi-videobridge/pull/1440

### DIFF
--- a/src/main/java/org/jitsi/jicofo/bridge/Bridge.java
+++ b/src/main/java/org/jitsi/jicofo/bridge/Bridge.java
@@ -45,10 +45,10 @@ public class Bridge
         = new ColibriStatsExtension();
 
     /**
-     * We assume that each recently added participant contributes this much
-     * to the bridge's packet rate.
+     * We assume that each recently added participant will contribute this much
+     * to the bridge's load
      */
-    private static final int AVG_PARTICIPANT_PACKET_RATE_PPS = config.averageParticipantPacketRatePps();
+    private static final Double AVG_PARTICIPANT_LOAD = config.averageParticipantLoad();
 
     /**
      * We assume this is the maximum packet rate that a bridge can handle.
@@ -77,9 +77,9 @@ public class Bridge
     private final RateStatistics newEndpointsRate = new RateStatistics(10000);
 
     /**
-     * The last reported packet rate in packets per second.
+     * The last report stress level
      */
-    private int lastReportedPacketRatePps = 0;
+    private double lastReportedStressLevel = 0.0;
 
     /**
      * Holds bridge version (if known - not all bridge version are capable of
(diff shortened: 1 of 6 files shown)


JonathanLennox (approved):

src/main/resources/reference.conf
```diff
@@ -7,6 +7,8 @@ jicofo {
     max-bridge-packet-rate = 50000
     // The assumed average packet rate per participant.
     average-participant-packet-rate-pps = 500
+    // The assumed average load amount per participant
+    average-participant-load = 0.01
```
JonathanLennox, line 11: This should probably be called "stress" rather than "load"? And is 0.01 too high? (100 participants per bridge feels low).

src/main/java/org/jitsi/jicofo/bridge/Bridge.java
```diff
@@ -72,6 +78,13 @@
      */
     private int lastReportedPacketRatePps = 0;
 
+    /**
+     * The last report stress level
+     */
+    private double lastReportedStressLevel = 0.0;
+
+    private boolean usePacketRateStat = false;
```
JonathanLennox, line 86: I'd name this `usePacketRateStatForStress`, and comment that it's for supporting older bridge versions.

src/main/java/org/jitsi/jicofo/bridge/Bridge.java
```diff
@@ -281,6 +308,22 @@ public String toString()
...
+        {
+            return getStressFromPacketRate();
+        }
+        // While a stress of 1 indicates a bridge is fully loaded, we allow
+        // larger values to keep sorting correctly.
+        return (lastReportedStressLevel + Math.max(0, getRecentlyAddedEndpointCount()) * AVG_PARTICIPANT_LOAD);
+    }
+
     /**
      * Returns the "stress" of the bridge. The stress is computed based on the
```
JonathanLennox, line 328: This comment should be updated to make clear that it's the backward-compatibility mode

src/main/resources/reference.conf
bbaldino: > This should probably be called "stress" rather than "load"?

Sure

> And is 0.01 too high? (100 participants per bridge feels low).

Yeah this was a bit of a placeholder to discuss.

src/main/java/org/jitsi/jicofo/bridge/Bridge.java
bgrozev: It's also never set to `true`. Do you want to initialize it to `true`, or set to `true` every time we receive stats without the `stress` value?

bbaldino: I cleaned it up a bit, take a look now.

src/main/java/org/jitsi/jicofo/bridge/Bridge.java
```diff
@@ -54,7 +54,7 @@
      * We assume that each recently added participant will contribute this much
      * to the bridge's load
      */
-    private static final Double AVG_PARTICIPANT_LOAD = config.averageParticipantLoad();
+    private static final Double AVG_PARTICIPANT_STRESS = config.averageParticipantStress();
```
bgrozev, line 57: Do we need the constant in addition to the value in config?

bbaldino: Good point, we probably don't.  I can get rid of it.

bgrozev: Asks the person who did the same on the line above...

src/test/java/org/jitsi/jicofo/bridge/BridgeSelectorTest.java
```diff
@@ -125,14 +125,14 @@ public void selectorTest()
         jvb3.setIsOperational(true);
 
         // Jvb 1 and 3 are occupied by some conferences, 2 is free
```
JonathanLennox, line 127: This comment should say "Jvb 1 and 2 are occupied", no?

bgrozev (approved):

PR #599, opened 2020-10-01
Handle multi focus allocator listerner

### LINKED ISSUE #552: should setFocusAllocationListener set event list?
Hello,

I need to catch when room is destroyed and I want to use FocusAllocationListener interface. However, when i investigate the code, i release that event object is not a list.
```
 private FocusAllocationListener focusAllocListener;
```
However, there are already two place register itself. It seems that both of org.jitsi.impl.reservation.rest.Activator and org.jitsi.jicofo.auth.AbstractAuthAuthority are using
```
 focusManager.setFocusAllocationListener(this);
```

and

```

public void setFocusAllocationListener(FocusAllocationListener l)
    {
        this.focusAllocListener = l;
    }
```

What do you thing about that? I thing focusAllocListener must be a list to trigger listeners.

L3o-pold:
Fix #552

### DIFF
--- a/src/main/java/org/jitsi/jicofo/FocusManager.java
+++ b/src/main/java/org/jitsi/jicofo/FocusManager.java
@@ -196,12 +196,10 @@ public class FocusManager
      */
     private final Object conferencesSyncRoot = new Object();
 
-    // Convert to list when needed
     /**
-     * FIXME: remove eventually if not used anymore
      * The list of {@link FocusAllocationListener}.
      */
-    private FocusAllocationListener focusAllocListener;
+    private final List<FocusAllocationListener> focusAllocListeners = new ArrayList<>();
 
     /**
      * XMPP protocol provider handler used by the focus.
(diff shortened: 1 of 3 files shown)


src/main/java/org/jitsi/jicofo/FocusManager.java
```diff
@@ -686,11 +686,16 @@ private int getNonHealthCheckConferenceCount()
     /**
      * Sets the listener that will be notified about conference focus
      * allocation/disposal.
+     * @param className the listener classname.
      * @param l the listener instance to be registered.
      */
-    public void setFocusAllocationListener(FocusAllocationListener l)
+    public void setFocusAllocationListener(String className, FocusAllocationListener l)
```
damencho, line 692: Is this className needed? Why don't you change the method to addFocusAllocationListener and drop the className param?

engesin: instead of hashmap, you may use list. Every class can add itself or delete itself. like,

        chatRoom.addMemberPresenceListener(this); X  chatRoom.removeMemberPresenceListener(this);

L3o-pold: I replaced hashmap with list so everyone should be happy

src/main/java/org/jitsi/jicofo/FocusManager.java
```diff
@@ -196,12 +196,10 @@
      */
     private final Object conferencesSyncRoot = new Object();
 
-    // Convert to list when needed
     /**
-     * FIXME: remove eventually if not used anymore
      * The list of {@link FocusAllocationListener}.
      */
-    private FocusAllocationListener focusAllocListener;
+    private final Vector<FocusAllocationListener> focusAllocListeners = new Vector<>();
```
damencho, line 202: Please, can you change this to be Arraylist, thank you?

damencho (changes requested):

damencho:
Jenkins, test this please.

PR #603, opened 2020-10-08
Handle Jibri errors on initial request

bbaldino:
If Jicofo encounters a Jibri error on the initial request, it throws and fails immediately because the assumption was that anything in the instances list should be present.  We've recently run into errors where a Jibri in a strange state is at the front of the queue and fails repeatedly: this means Jicofo fails each time but, for the next request, tries the same Jibri again.  Because we don't know how permanent the error state is for this Jibri, this PR adds an API to denote that a JibriSession encountered a transient error with the Jibri, which moves it to the back of the line.  It also retries in this case, similar to the retries we perform when getting an error response back from Jibri.

### DIFF
--- a/src/main/java/org/jitsi/jicofo/recording/jibri/JibriSession.java
+++ b/src/main/java/org/jitsi/jicofo/recording/jibri/JibriSession.java
@@ -373,8 +373,14 @@ public class JibriSession
         {
             logger.error("Failed to send start Jibri IQ: " + e, e);
             jibriDetector.memberHadTransientError(jibriJid);
-
-            throw new StartException(StartException.INTERNAL_SERVER_ERROR);
+            if (!maxRetriesExceeded())
+            {
+                retryRequestWithAnotherJibri();
+            }
+            else
+            {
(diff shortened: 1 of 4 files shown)


src/main/java/org/jitsi/jicofo/recording/jibri/JibriSession.java
```diff
@@ -372,8 +372,15 @@ private void startInternal()
         catch (Exception e)
         {
             logger.error("Failed to send start Jibri IQ: " + e, e);
-
-            throw new StartException(StartException.INTERNAL_SERVER_ERROR);
+            jibriDetector.memberHadTransientError(jibriJid);
+            if (!maxRetriesExceeded())
+            {
+                retryRequestWithAnotherJibri();
```
bbaldino, line 378: Note that this is effectively a recursive call, as this calls `start` which calls `startInternal`.  We did not recursively call `start` before because `retryRequestWithAnotherJibri` was called in the context of handling a Jibri response.  I don't think this should be a problem (I'll be testing to verify that) but thought it was worth pointing out.

damencho (approved):

PR #607, opened 2020-10-21
a test that checks the default value of minParticipants

gluckzhang:
Hello,

I'd like to contribute in order to improve the test quality and effectiveness of Jicofo. During mutation testing, I found that the behavior of [`org.jitsi.jicofo.JitsiMeetConfig.getMinParticipants()`](https://github.com/jitsi/jicofo/blob/master/src/main/java/org/jitsi/jicofo/JitsiMeetConfig.java#L287) is not specified. If the method returns an arbitrary value that is <= 2, no test fails.

Here is a test case that specifies it with a new assertion.

```java
@Test
public void testGetMinParticipants() {
    Map<String, String> properties = new HashMap<String, String> ();
    JitsiMeetConfig jitsiMeetConfig = new JitsiMeetConfig(properties);
    Assert.assertEquals(2, jitsiMeetConfig.getMinParticipants());
}
```

What do you think?

Thanks!


monperrus:
@ibauersachs CLA and CI checks are all green :)

monperrus:
@bgrozev WDYT of this new test that better specifies `getMinParticipants` (was not tested before)?

bgrozev:
Hey, @gluckzhang thanks for taking the time to contribute to our project and sorry for the slow response! After reviewing the use of this option, we decided to [remove it](https://github.com/jitsi/jicofo/pull/638) and make it configurable strictly on jicofo's side (this is part of a wider effort to remove some of the client-side configuration). As a result, the method does not exist anymore (once the PR is merged). 

If you think tests are necessary for the new code (in `ConferenceConfig.kt`), you are welcome to add them (e.g. something similar to `JicofoConfigTest.kt`). I promise a prompt review!

gluckzhang:
Hej @bgrozev , thanks for your reply! Based on the PR you mentioned, I think I can close this PR for now. If I find another similar case (the method is covered by test cases, but it is not well tested according to the mutation testing), I will make a new PR to discuss with you :)

PR #613, opened 2020-10-29
Add Content-Type to Shibboleth handler response

maxbes:
Currently, the Shibboleth handler (/login) does not set a Content-Type:

image

This can cause issues with certain reverse proxies, certain default Content-Type configurations, and `X-Content-Type-Options: nosniff` that may cause the HTML source code to appear in the browser window instead of being rendered

This PR adds the missing Content-Type header


maxbes:
done

> Hi, thanks for your contribution!
> If you haven't already done so, could you please make sure you sign our CLA (https://jitsi.org/icla for individuals and https://jitsi.org/ccla for corporations)? We would unfortunately be unable to merge your patch unless we have that piece :(.

done

PR #614, opened 2020-10-30
fix: Remove non-operational bridges and add stats.

bgrozev:


src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
```diff
@@ -1575,6 +1575,7 @@ public XMPPError onSessionInfo(JingleSession session, JingleIQ iq)
                     address,
                     bridgeSessionId));
         }
+        listener.participantIceFailed();
```
bbaldino, line 1578: Should we update this even in the "invalid session" case? (I'm not sure what 'invalid session' implies, exactly)

bbaldino (dismissed):
One small question, but lgtm.

src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
bgrozev: I don't know what the "invalid session" case is either. I wanted to keep the semantics closer to "how many times a client reported ICE failed", so this made sense.

src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
```diff
@@ -870,14 +884,25 @@ private void inviteParticipant(
             boolean[] startMuted)
     {
         BridgeSession bridgeSession;
+
+        // Some of the bridge in the conference may have become non-operational. Inviting a new participant to the
```
JonathanLennox, line 888: Grammar: "some of the bridges"

src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
```diff
@@ -846,8 +846,22 @@ private Bridge selectBridge(Participant participant)
...
+    /**
+     * Removes all non-operational bridges from the conference and re-invites their participants.
+     */
+    private void removeNonOperationalBridges()
+    {
+        Set<Jid> nonOperationalBridges = bridges.stream()
+                .filter(session -> session.hasFailed || !session.bridge.isOperational())
+                .map(session -> session.bridge.getJid())
+                .collect(Collectors.toSet());
+        if (nonOperationalBridges.size() > 0)
```
JonathanLennox, line 861: Suggest `if (!nonOperationalBridges.empty())`.

src/main/java/org/jitsi/jicofo/stats/Statistics.java
```diff
@@ -18,4 +18,24 @@
      * Jicofo since it was started
      */
     public final AtomicInteger totalParticipants = new AtomicInteger(0);
+
+    /**
+     * The number of participants that were moved away from a failed bridge.
+     */
+    public final AtomicInteger totalParticipantsMoved = new AtomicInteger();
```
JonathanLennox, line 25: Suggest these use `LongAdder`?  But probably all the stats should be updated at once.

bgrozev: I don't think any of these are updated often enough for performance to be a concern.

JonathanLennox (approved):

PR #622, opened 2020-11-14
Jibri send 'stopRequst' with session ID.

jiyeyuran:
I'm working on Jibri to make it supporting recording multi rooms parallelly. The jibri need the session ID to decide which room recording should stop.


jiyeyuran:
Signed

bbaldino:
I'm not sure we'd merge all needed changes to support this feature, but I think this one is reasonable.

bbaldino (approved):

PR #623, opened 2020-11-16
Clean up and port configuration (part 1)

bgrozev:


src/main/java/org/jitsi/jicofo/FocusManager.java
```diff
@@ -152,14 +123,42 @@
...
+        {
+            ServiceConnectionConfig config = XmppConfig.xmppConfig.getServiceConnectionConfig();
 
-    /**
-     * The username used by the focus to login.
-     */
-    private Resourcepart focusUserName;
+            if (!config.enabled())
+            {
+                logger.info("Service XMPP connection noot enabled.");
```
bbaldino, line 138: nit: typo

src/main/java/org/jitsi/jicofo/FocusManager.java
```diff
@@ -284,44 +279,33 @@ public void start()
...
-        String focusUserPassword = config.getString(FOCUS_USER_PASSWORD_PNAME);
-        String xmppServerPort = config.getString(XMPP_PORT_PNAME);
+        ClientConnectionConfig clientConnectionConfig = xmppConfig.getClientConnectionConfig();
 
         // We default to "conference" prefix for the muc component
-        String conferenceMucPrefix
-            = config.getString(XMPP_MUC_COMPONENT_PREFIX_PNAME, "conference");
-        conferenceMucService = JidCreate.domainBareFrom(
-            conferenceMucPrefix + "." + xmppDomainConfig);
+        String conferenceMucPrefix = config.getString(XMPP_MUC_COMPONENT_PREFIX_PNAME, "conference");
```
bbaldino, line 285: Why is this property still in the old config?  Is it not part of `ClientConnectionConfig`?

src/main/kotlin/org/jitsi/jicofo/bridge/BridgeConfig.kt
```diff
@@ -101,6 +102,12 @@ class BridgeConfig {
             .transformedBy { Duration.ofMillis(it.toMillis() / 2) }
     }
 
+    val brewery: String? by optionalconfig {
```
bbaldino, line 105: nit: this is the brewery jid (as a string), right?  `breweryJid` or `breweryName`, maybe (since it isn't actually a `Jid`)?

src/main/kotlin/org/jitsi/jicofo/xmpp/XmppConfig.kt
```diff
@@ -0,0 +1,156 @@
...
+        @JvmField
+        val xmppConfig = XmppConfig()
+    }
+}
+
+class ServiceConnectionConfig {
+    private val enabled: Boolean by config {
+        "jicofo.xmpp.service.enabled".from(newConfig)
+    }
+    fun enabled() = legacyHostname != null || enabled
```
bbaldino, line 44: nit: I don't think it really matters, but I _think_ we could do:
```
val enabled: Boolean by config {
    "org.jitsi.jicofo.BRIDGE_MUC_XMPP_HOST".from(legacyConfig).convertFrom<String> { it != null }
    "jicofo.xmpp.service.enabled".from(newConfig)
}
```

and then get rid of the need for the separate `enabled()` and `legacyHost` fields.  I don't care much if we change it, though.

src/main/kotlin/org/jitsi/jicofo/xmpp/XmppConfig.kt
```diff
@@ -0,0 +1,156 @@
...
+
+    private val legacyHostname: String? by optionalconfig {
+        "org.jitsi.jicofo.BRIDGE_MUC_XMPP_HOST".from(legacyConfig)
+    }
+
+    val hostname: String by config {
+        "org.jitsi.jicofo.BRIDGE_MUC_XMPP_HOST".from(legacyConfig)
```
bbaldino, line 51: If we _do_ keep `legacyHostName`, I think you could re-use it here by doing:
```
val hostname: String by config {
    "legacy hostname" { legacyHostname }
    "jicofo.xmpp.service.hostname".from(newConfig)
}
```

src/main/java/org/jitsi/jicofo/FocusManager.java
bgrozev: This is fixed in a later commit

bgrozev: Which will now be part of another PR

src/main/java/org/jitsi/jicofo/FocusManager.java
bgrozev: Because I split the changes in multiple PRs.

src/main/kotlin/org/jitsi/jicofo/bridge/BridgeConfig.kt
bgrozev: Yes, that's the brewery JID as a string. Do you care about the name of the `val`, or the name in the config file?

bbaldino: The name of the `val` is what caught my eye, but I suppose both.

src/main/kotlin/org/jitsi/jicofo/bridge/BridgeConfig.kt
```diff
@@ -102,11 +104,15 @@ class BridgeConfig {
...
     }
 
-    val brewery: String? by optionalconfig {
-        "org.jitsi.jicofo.BRIDGE_MUC".from(JitsiConfig.legacyConfig)
-        "$BASE.brewery".from(JitsiConfig.newConfig)
+    val breweryJid: Jid? by optionalconfig {
+        "org.jitsi.jicofo.BRIDGE_MUC".from(JitsiConfig.legacyConfig).convertFrom<String> {
+            JidCreate.bareFrom(it)
+        }
+        "$BASE.brewery".from(JitsiConfig.newConfig).convertFrom<String> {
```
bbaldino, line 111: forgot to update the id here to `brewery-jid`

bbaldino (approved):

PR #624, opened 2020-11-16
Clean up and port config (part 2)

bgrozev:


src/main/resources/reference.conf
```diff
@@ -167,4 +167,11 @@ jicofo {
       reply-timeout = 15 seconds
     }
   }
+
+  octo {
+    // An identifier of the Jicofo instance, used for the purpose of generating conference IDs unique across a set of
+    // Jicofo instances.
+    // Valid values are [1, 65535]. The value "0" is reserved for the case where an ID is not explicitly configured.
+    id = -1
```
bbaldino, line 175: Does `-1` make sense for the default?  What is the meaning of the `-1` value? (i.e. what will happen with this default?)

bgrozev: Hmm. Actually anything <1 is treated the same: we will log a warning on startup, and every time Octo is enabled for a conference.

bbaldino: Ok, should we make it optional and just have that behavior occur when it isn't set?  Or maybe use `0` for the default, since its behavior is documented there.

bbaldino (approved):

PR #626, opened 2020-11-16
Refactor FocusComponent and Health (non-OSGi).'

bgrozev:


src/main/java/org/jitsi/jicofo/Main.java
```diff
@@ -38,54 +41,133 @@
...
+        {
+            WaitableBundleActivator.waitUntilStarted();
+        }
+        catch (Exception e)
+        {
+            logger.error("Failed to start all OSGi bundles, exiting.");
+            OSGi.stop(activator);
+            return;
+        }
+        logger.warn("OSGi services started.");
```
bbaldino, line 88: Log on info?

src/main/java/org/jitsi/jicofo/Main.java
```diff
@@ -38,54 +41,133 @@
...
+            public void stop(BundleContext bundleContext)
+                    throws Exception
+            {
+                // We're doing nothing
+            }
+        };
+
+
+        // Start OSGi
+        logger.warn("Starting Osgi");
```
bbaldino, line 160: info?

src/main/java/org/jitsi/jicofo/xmpp/FocusComponent.java
```diff
@@ -80,67 +83,60 @@
...
-    /**
-     * Releases resources used by this instance.
-     */
-    public void dispose()
+    public void connect()
     {
-        OSGi.stop(this);
+        if (!isPingTaskStarted())
+        {
+            startPingTask();
```
bbaldino, line 124: Does this need to be atomic with the `isPingTaskStarted()` check?

bgrozev: In theory -- yes. In practice -- no. That's the only place any of that code is used, it [has always been](https://github.com/jitsi/jicofo/blob/master/src/main/java/org/jitsi/jicofo/xmpp/FocusComponent.java#L118) like this, and I'm hoping to remove all that code soon, so I'd rather not add to it.

bbaldino: Cool, just checking.

bbaldino (approved):

PR #630, opened 2020-11-19
Port the authentication config, simplify component config.

bgrozev:


src/main/java/org/jitsi/jicofo/Main.java
```diff
@@ -199,34 +199,29 @@ private static void setSystemProperties(String[] args)
...
         {
             secret = System.getenv("JICOFO_SECRET");
         }
-        if (secret != null)
-        {
-            System.setProperty(XmppComponentConfig.secretPropertyName, secret);
-        }
+
+        XmppComponentConfig.config = new XmppComponentConfig(
+                host == null ? "" : host,
```
bbaldino, line 219: I don't think I'm following this...how does `host` get pulled from config?  I see we read it from the command line here and pass that if it's non-null.

bgrozev: Yes. Currently the only way to configure this value is with command line arguments. Since the aim is to soon remove the component code altogether, I chose not to do "new config" properties for it.

bbaldino: Ah! Now I see.  Ok.

bbaldino (approved):

PR #559, opened 2020-06-24
Add access token config for reservation REST API

L3o-pold:

PR #631, opened 2020-11-19
ref: Remove the reservation system.

bgrozev:


paweldomas (approved):

paweldomas:
I can't remember if there's anything related in jitsi-meet or LJM...

bgrozev:
There's handling of `reservation-error` in `moderator.js`, but I'm not sure it's only used with jicofo's reservation system. @damencho mentioned there are prosody modules that implement reservation. Do they use the same signaling?

paweldomas:
I don't know about any prosody modules for reservation, maybe there's something in the community

damencho:
Yeah, I mentioned that people are handling this in prosody and this is from the community, nothing official

bgrozev:
> Yeah, I mentioned that people are handling this in prosody and this is from the community, nothing official

Do they use the same signaling? I.e. do we need to keep the code in the client?

damencho:
Don't think so, it is based on tokens 🔥 it

freyes-lef:
Excuse me ... i´m integrating a reservation system 
to jicofo over kubernetes  ... ¿Why remove the reservation system? ¿have problems? 

Thanks in advance

bgrozev:
We removed it because it is not in active use in our project (or others we know about), and the implementation rots over time. Can you tell us more about your use-case? Can you use prosoddy modules to accomplish the same?

qwertiko:
Hi,

we too are using this reservation system - it checks via https against another db if there are current planed reservations, a matching 'private room' and limits the number of concurrent meetings per user as well as enforcing time limits on other users.

This module is kinda crucial to the service we provide currently.

qwertiko:
Prosody can be used for authentication - we are doing that too, but that is not related to the reservation handling.
I'll happily stand corrected if you point me to a prosody plugin that will accomplish similar regarding reservations and blocking the creation of new rooms.

damencho:
All those functionality can be implemented as a prosody module. There is no such module at the moment but we will be happy to help doing one if you are willing to work on contributing one.

qwertiko:
We would be happy to work on one - with a timeline. By that I mean the possibility of reverting the removal of this code until we have a working alternative in prosody. Your help in this will be greatly appreciated. Of course the result will be published for all others to obtain and use, too.

L3o-pold:
I agree, we still need reservation system in jicofo until a prosody module is release (if possible with authentication support as https://github.com/jitsi/jicofo/pull/559).

paweldomas:
@bgrozev maybe reservation can be kept with no guarantees and up for the community to maintain it? that's how it's been so far, when something broke people would submit patches to fix

paweldomas:
@damencho doing things as prosody module is not ideal because you are stuck with Prosody and there are other servers out there

freyes-lef:
> We removed it because it is not in active use in our project (or others we know about), and the implementation rots over time. Can you tell us more about your use-case? Can you use prosoddy modules to accomplish the same?

Thank you for you reply

our needs are very simple: we have a jitsi kubernetes deployed in aws (elastic machines)
, but we have to integrate it with our applications (in various technologies:  .net, react, php ...).

The integration requires (in a simplified way):
1. a user can create the room giving it a name and optionally a password
2. Obtain statistics from that room later (by name) to integrate the info in ours apps
3. be able to delete the room or, at least, know the deletion

Through the "restapi" we have not found a way to match the identifier with the room names created, so we have done the following:
1. create rooms to start the flow with direct calls to "jitsi-meet? Room=xxxx"
2. collect name data through "jicofo / reservation" and obtain identifiers
3. request statistics through your "rest-api" with the identifiers

In preparation we have the issues of controlling the users who are connecting, although it is not essential

stefanplk3:
+1 for please do not remove without providing a working alternative. We are as well heavily using the reservation system (timeslots, limit concurrent reservations etc.)

damencho:
This change had been reverted. https://github.com/jitsi/jicofo/pull/648

damencho:
@stefanplk3 @freyes-lef @L3o-pold @qwertiko 
Please can you test this with your setup: https://github.com/jitsi/jitsi-meet/pull/8386
I will be looking at the PR and testing it in the following weeks, so we can proceed with it.
This is a chance to help improve it and to accommodate your needs and also we can work on the UI part so it looks ok (maybe in follow up PR), any help testing and comment in the PR will be more than welcome, thank you all.

qwertiko:
sorry for the late feedback. we are looking into the prosody plugin today.

qwertiko:
directly after a restart of prosody it seems to work, however the following tries result in the reservation system not being triggered at all (no entries in the prosody log regarding reservations).

We are using a prosody plugin for authentication and only when we are asked to authenticate does is the prosody reservation plugin triggered. however, the authentication is kept alive so in the following tests we are not prompted for authentication. in exaclty these cases we see nothing in the logs regarding the reservation plugin.

any ideas?

damencho:
You mean that secure domain is configured with the reservation system?

qwertiko:
correct

qwertiko:
when we set DISABLE_AUTOLOGIN for jicofo it works :+1:

damencho:
But when using the jicofo reservation, the reservation system was contacted even when auth is stored?
@shawnchin were you using authentication in some of your tests? 

I wonder how is this possible ... @shawnchin does that mean that jicofo is not leaving the rooms when authentication is stored (this will prevent destroying the room in prosody) ... if this is the case, that is a jicofo bug ...

shawnchin:
@damencho I'm afraid I've only tested the plugin with JWT token auth since that is all we use.

Would certainly be problematic if jicofo does not leave the room; the prosody plugin was designed such that is does not interfere if room already exists.

@qwertiko do you see logs for "Dropped reservation data for destroyed room" in prosody logs after all users leave when you have auth enabled?

damencho:
@shawnchin Yep, no problem ... we will figure it out, I can be testing and that case when I come to it ... but yeah these investigations help a lot. Thanks @qwertiko

qwertiko:
we are using a type of mod_auth_http - with cached authentication reservation code was not triggered at all.

while the reservation was part of jicofo this worked fine.

qwertiko:
@shawnchin we do find the mentioned log entries, however also at times after we had set DISABLE_AUTOLOGIN

stefanplk3:
Same problems here. We are using secure domain with ldap. But I have also tried secure domain with prosody user auth on a fresh jitsi installation following the Debian/Ubuntu Install Docs on jitsi.org.

The mod_reservations calls room:clear() and room:destroy() but jicofo doesn't do a thing after that. We cannot rejoin this room until jicofo is restarted.

All is well when all users leave the room. But the case of the time expiring the room is where jicofo doesn't leave and the room cannot be rejoined.

@damencho It would be awesome if you could fix this.

shawnchin:
FYI, just pushed a change to the PR to fix bug where DELETE /conference call was not issued on room expiry. 
Does not address issue with Jicofo not leaving the room though.

PR #640, opened 2020-12-01
feat: Add config to enable/disable Octo. Remove client-side config.

bgrozev:


bbaldino (approved):

rasos:
The [docs](https://github.com/jitsi/jitsi-videobridge/blob/master/doc/octo.md#jitsi-meet-configuration) still suggest to enable octo - I assume that this is not required anymore in config.js

bgrozev:
> The [docs](https://github.com/jitsi/jitsi-videobridge/blob/master/doc/octo.md#jitsi-meet-configuration) still suggest to enable octo - I assume that this is not required anymore in config.js

Correct. The doc is outdated, but I don't think it's worth updating at this stage since secure-octo is now merged.

PR #643, opened 2020-12-01
Remove conference-wide RTX/REMB/TCC/opus-red options.

bgrozev:
The RTX, REMB, TCC and opus-red per-conference propeties have been remove from the client. They have been converted to: jicofo config which allows allows or does not allow them to be used, and the clients advertising support for these features (a feature is used if it is enabled in jicofo's config AND the client supports it, with no per-conference setting).


bgrozev:
Will wait before merging this, see the discussin in jitsi/lib-jitsi-meet#1423

bbaldino (approved):

PR #644, opened 2020-12-01
ref: Move the shared document name config.

bgrozev:


src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
```diff
@@ -2532,13 +2532,13 @@ public void setStartMuted(boolean[] startMuted)
      */
     private String createSharedDocumentName()
     {
-        if (config.useRoomAsSharedDocName())
+        if (ConferenceConfig.config.getUseRandomSharedDocumentName())
         {
-            return roomName.getLocalpart().toString();
+            return UUID.randomUUID().toString().replaceAll("-", "");
```
bbaldino, line 2537: Are these flipped?  Before, if `useRoomAsSharedDocName` was true, we used `roomName.getLocalpart().toString();`

bbaldino: Oops, I just noticed the variable name is different and the semantics were inverted.

bbaldino: nit: it'd be nice for places where java is the main consumer of config params to create functions to make the call more readable (e.g. `ConferenceConfig.config.useRandomSharedDocumentName()` here)

bbaldino (approved):

PR #646, opened 2020-12-01
ref: Remove the openSctp option.

bgrozev:


bbaldino:
This is effectively removing the client's ability to set this option, is that right?  But we still have a backend configuration for it?

bgrozev:
Exactly. We have 3 layers of SCTP config:
1. The global config in jicofo
2. The "openSctp" option, which comes from the client and is conference-wide
3. Per-participant support

This PR removes the second option.

bbaldino (approved):

PR #654, opened 2020-12-10
Move jetty out of AuthBundleConfig

bgrozev:


src/main/java/org/jitsi/jicofo/ChatRoomRoleAndPresence.java
```diff
@@ -101,7 +101,8 @@ public ChatRoomRoleAndPresence(JitsiMeetConferenceImpl conference,
      */
     public void init()
     {
-        authAuthority = Main.getJicofoServices().getAuthenticationAuthority();
+        authAuthority = JicofoServices.jicofoServicesSingleton == null
```
bbaldino, line 104: Is this only true in the unit test case?  When unit testing, don't we need to stub out `JicofoServices.jicofoServicesSingleton` ?

bbaldino (approved):

src/main/java/org/jitsi/jicofo/ChatRoomRoleAndPresence.java
bgrozev: We could, but currently we don't (this is the only place that uses it). Let's see how this evolves.

PR #656, opened 2020-12-17
Handle ConferenceIq and authentication IQs coming from Smack

bgrozev:


src/main/kotlin/org/jitsi/jicofo/JicofoServices.kt
```diff
@@ -94,20 +94,24 @@ open class JicofoServices(
             }
         }
 
-        val focusJid = XmppConfig.client.username.toString() + "@" + XmppConfig.client.domain.toString()
         conferenceRequestHandler = ConferenceRequestHandler(
             focusManager = focusManager,
-            focusAuthJid = focusJid,
+            focusAuthJid = XmppConfig.client.username.toString() + "@" + XmppConfig.client.domain.toString(),
```
bbaldino: nit:  `"${XmppConfig.client.username}@${XmppConfig.client.domain}`

(I'm _pretty_ sure you don't need the `toString`s there when using that method, but not 100%)

src/main/kotlin/org/jitsi/jicofo/JicofoServices.kt
```diff
@@ -94,20 +94,24 @@ open class JicofoServices(
...
             isFocusAnonymous = StringUtils.isBlank(XmppConfig.client.password),
             authAuthority = authenticationAuthority,
             reservationSystem = reservationSystem
         )
-        focusComponent = FocusComponent(XmppComponentConfig.config, conferenceRequestHandler).apply {
-            val configService = ServiceUtils2.getService(bundleContext, ConfigurationService::class.java)
-            loadConfig(configService, "org.jitsi.jicofo")
-            authenticationAuthority?.let { setAuthAuthority(authenticationAuthority) }
+
+        focusComponent = if (XmppComponentConfig.config.enabled) {
```
bbaldino, line 101: I forget, is this use of "component" something that we're working on getting rid of?  Or is there something that still needs it?

src/main/kotlin/org/jitsi/jicofo/JicofoServices.kt
```diff
@@ -94,20 +94,24 @@ open class JicofoServices(
...
+            FocusComponent(XmppComponentConfig.config, conferenceRequestHandler).apply {
+                val configService = ServiceUtils2.getService(bundleContext, ConfigurationService::class.java)
+                loadConfig(configService, "org.jitsi.jicofo")
+                authenticationAuthority?.let { setAuthAuthority(authenticationAuthority) }
```
bbaldino: nit: I don't know if this was intentional or not (I've done this same thing before on accident), but I think it's more idiomatic to use `it` here:

```
authenticationAuthority?.let { setAuthAuthority(it) }
```

Using `authenticationAuthority` works since it's a val and it gets smart-cast, but I think it's a little more brittle to use `authenticationAuthority` rather than the explicit result of the `let`.

src/main/kotlin/org/jitsi/jicofo/JicofoServices.kt
```diff
@@ -94,20 +94,24 @@ open class JicofoServices(
...
+                connect()
```
bbaldino, line 108: Reading this I wonder if it would be better to move the `connect` call out to make it more obvious (also since the code in the apply is more about initialization and this is a bit different, so doing:

```
focusComponent = if (...) {
    ...
} else {
    null
}

focusComponent?.connect()
```

src/main/kotlin/org/jitsi/jicofo/xmpp/XmppConfig.kt
```diff
@@ -212,6 +212,9 @@ data class XmppComponentConfig(
     val port: Int = -1,
     val secret: String = ""
 ) {
+    val enabled: Boolean
```
bbaldino, line 215: nit: since `hostname` is a val, this could be: `val enabled: Boolean = hostname != ""`  There's also `isEmpty` or `isBlank` helpers for `String`, depending on what we actually care about hostname being.

src/main/kotlin/org/jitsi/jicofo/xmpp/AuthenticationIqHandler.kt
```diff
@@ -0,0 +1,59 @@
...
+import org.jitsi.xmpp.extensions.jitsimeet.LogoutIq
+import org.jivesoftware.smack.packet.IQ
+import org.jivesoftware.smack.packet.XMPPError
+
+class AuthenticationIqHandler(val authAuthority: AuthenticationAuthority) {
+    private val logger = createLogger()
+
+    fun handleLoginUrlIq(loginUrlIq: LoginUrlIq): IQ {
+        val peerFullJid = loginUrlIq.from.asEntityFullJidIfPossible()
+        val roomName = loginUrlIq.room
```
bbaldino, line 37: Looks like it might be worth writing some helpers to create error responses.  If we had them we could do something like:
```
val roomName = loginUrlIq.room ?: return createNotAcceptableErrorResponse(loginUrlIq)
```

with helpers like:

```
fun createErrorResponse(iq: IQ, condition: XMPPError.Condition): IQ = IQ.createErrorResponse(iq, XMPPError.getBuilder(condition))

fun createNotAcceptableErrorResponse(iq: IQ): IQ = createErrorResponse(iq, XMPPError.Condition.not_acceptable)
```

src/main/java/org/jitsi/jicofo/xmpp/IqHandler.java
```diff
@@ -39,12 +40,12 @@
  * @author Pawel Domas
  * @author Boris Grozev
  */
-public class MeetExtensionsHandler
+public class IqHandler
```
bbaldino, line 44: The name feels a little broad, does this handle all IQs? Or only a certain scope of them?

src/main/kotlin/org/jitsi/jicofo/JicofoServices.kt
```diff
@@ -105,6 +107,10 @@ open class JicofoServices(
 
         val authenticationIqHandler = authenticationAuthority?.let { AuthenticationIqHandler(it) }
 
+        iqHandler = IqHandler(focusManager).apply {
+            focusManager.addXmppConnectionListener { init(it) }
```
bbaldino: I know Intellij says to use a lambda in these cases since the interface only has a single method, but I find it pretty confusing in the cases where it's an interface that's being implemented.  I think writing it out in full would be a lot clearer (it's basically impossible to infer what `it` represents here, let alone that the handler there is for a specific `xmppConnectionInitialized` event).

I.e.:

```
focusManager.addXmppConnectionListener(object : XmppConnectionListener {
    override fun xmppConnectionInitialized(xmppConnection: XmppConnection) {
        init(xmppConnection)
    }
})
```

bbaldino (commented):
I don't have much context, so this ended up being mainly a codestyle review.

src/main/kotlin/org/jitsi/jicofo/JicofoServices.kt
bbaldino: From what I saw this got fixed in a later commit

src/main/kotlin/org/jitsi/jicofo/JicofoServices.kt
bgrozev: Yes, I'm working on getting rid of it. Jitsi-meet still uses the jid of the component, I think that's the last use.

src/main/java/org/jitsi/jicofo/xmpp/IqHandler.java
bgrozev: It's a subset (e.g. it doesn't include jingle), but it's pretty broad (authentication, conference request, muting, and rayo). What do you suggest as a better name?

bbaldino: Is there something that all those things (everything but jingle it sounds like?) falls under?

bbaldino (approved):

PR #661, opened 2021-01-06
Refactor Health and Version, adapt to jicoco.

bgrozev:


src/main/java/org/jitsi/jicofo/Main.java
```diff
@@ -62,17 +64,12 @@ public static void main(String[] args)
...
-            synchronized (exitSyncRoot)
-            {
-                exitSyncRoot.notifyAll();
-            }
-        }));
+        Runtime.getRuntime().addShutdownHook(new Thread(shutdownService::beginShutdown));
         logger.info("Starting OSGi services.");
-        BundleActivator activator = startOsgi(exitSyncRoot);
+        BundleActivator activator = new DummyActivator();
+        OSGiLauncher launcher = startOsgi(activator);
```
bbaldino, line 72: We have to still start OSGi for some things, right?  But the dummy activator does nothing, so what behavior is it we're relying on there?

bgrozev: At first flance `OSGiLauncher.start(BundleActivator)` requires a non-null activator, so I added that without digging further. But I see there a `start()` with no arguments that I can use.

bgrozev: Ugh, `start()` is private...

bbaldino: Ok...but what is it that happens as part of `OSGiLauncher.start` that we still need? (Just curious)

bgrozev: Starting all of our bundles, we still have a bunch:
https://github.com/jitsi/jicofo/blob/master/src/main/java/org/jitsi/jicofo/osgi/JicofoBundleConfig.java#L69

bbaldino: Oh, and those still work even though the activator is a no-op?

bgrozev: Yep 🤷

bgrozev: I think I could move `WaitableBundleActivator` from the bundle config here, but I'd rather not fiddle with it

bbaldino (dismissed):

bbaldino (approved):

PR #667, opened 2021-01-12
Use XMPP config directly, do not register Jitsi accounts

bgrozev:


src/main/java/org/jitsi/impl/protocol/xmpp/XmppProviderFactory.java
```diff
@@ -37,7 +37,7 @@
      * @param bundleContext the bundle context reference of the service
      * @param protocolName  the name of the protocol
      */
-    protected XmppProviderFactory(
+    public XmppProviderFactory(
```
bbaldino: Is this visibility change temporary (while refactoring things, maybe across other PRs), or fix?  Just wondering if we should comment on it being temporary so others don't use it while the bigger refactoring is taking place.

bbaldino (approved):

src/main/java/org/jitsi/impl/protocol/xmpp/XmppProviderFactory.java
bgrozev: It's not temporary, there's no reason this should not be public (the latest version is a simple one-method interface)

bgrozev:
I have to take a look at the tests, because the mock MUCs from different tests interfere with each other. I don't understand how that change triggered it.

bbaldino (dismissed):

bbaldino (approved):

PR #672, opened 2021-01-13
ref: Remove FocusComponent.

bgrozev:


bbaldino (dismissed):

damencho (approved):

licaon-kter:
So how does one run this now? 

There are no changes in https://github.com/jitsi/jicofo/blob/master/resources/jicofo.sh

bgrozev:
> So how does one run this now?

It runs the same way (some of the options in `jicofo.sh` are obsolete)

PR #673, opened 2021-01-15
fix: Fix leaking JitsiMeetConferenceImpl insatnces.

bgrozev:


damencho (approved):

src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
```diff
@@ -428,6 +428,9 @@ synchronized void stop()
 
         protocolProviderHandler.removeRegistrationListener(this);
 
+        BridgeSelector bridgeSelector = JicofoServices.jicofoServicesSingleton.getBridgeSelector();
+        bridgeSelector.removeHandler(bridgeSelectorEventHandler);
```
engesin, line 432: getBridgeSelector() may return null. it is better to make null check

engesin: also JicofoServices.jicofoServicesSingleton can return null. are these cases impossible to face?

bgrozev: This PR is from January, I think yoou have valid points but they have been fixed in master -- JitsiMeetConference.jicofoServices is initialized early (with a failure if it's null), and `getBridgeSelector` can no longer return null.

PR #681, opened 2021-01-21
ref: Remove obsolete options (port, secret).

bgrozev:


damencho:
There is still one JICOFO_SECRET in the init.d file.

damencho (approved):

PR #683, opened 2021-01-22
Remove OSGi, jitsi-desktop, libjitsi.

bgrozev:


bbaldino (approved):
The hard-coded strings (`sendonly`, `recvonly` and all the payload type extension URNs) are a bummer...we don't have those defined centrally anywhere other than jitsi (which I'm assuming is where they were)?

bgrozev:
> The hard-coded strings (`sendonly`, `recvonly` and all the payload type extension URNs) are a bummer...we don't have those defined centrally anywhere other than jitsi (which I'm assuming is where they were)?

They were in libjitsi. We can extract them as constants in jicofo, but honestly I don't see much value.

bgrozev:
The test failure is because of the size of the jicofo log file, which is twice the usual because the logging properties on the pr-torture machine hasn't been updated.

bbaldino:
> > The hard-coded strings (`sendonly`, `recvonly` and all the payload type extension URNs) are a bummer...we don't have those defined centrally anywhere other than jitsi (which I'm assuming is where they were)?
> 
> They were in libjitsi. We can extract them as constants in jicofo, but honestly I don't see much value.

Ah ok, I thought we were referencing them elsewhere as well.  Constants would still be nice--even if just because they'd be shorter, but I'd think we'd probably want to use them in tests as well.

PR #697, opened 2021-02-18
feat: Read average_participant_stress from bridge stats.

bgrozev:


src/main/java/org/jitsi/jicofo/bridge/Bridge.java
```diff
@@ -107,8 +108,12 @@ static void setFailureResetThreshold(long newValue)
      * working bridges go down and might eventually get elevated back to
      * {@code true}.
      */
-    private volatile boolean isOperational
-        = true /* we assume it is operational */;
+    private volatile boolean isOperational = true;
+
+    /**
+     * Start out with the configured value, update if the bridge reports a value.
```
JonathanLennox, line 114: Do you want to update the doc in reference.conf to clarify that the config'd value is just a default?

JonathanLennox (approved):

PR #701, opened 2021-03-02
 @bgrozev fix: Do not send source-add for injected SSRCs. 

bgrozev:
Do not send source-add for injected SSRCs because Octo currently requires each endpoint to have at least one associated SSRC, jicofo injects a dummy audio SSRC for endpoints that do not advertise any SSRCs of their own. This makes sure that the dummy SSRCs are kept only for Octo and not propagated to other endpoints.

### DIFF
--- a/src/main/java/org/jitsi/protocol/xmpp/AbstractOperationSetJingle.java
+++ b/src/main/java/org/jitsi/protocol/xmpp/AbstractOperationSetJingle.java
@@ -267,7 +267,7 @@ public abstract class AbstractOperationSetJingle
                         ssrcs.getSourcesForMedia("audio").stream().allMatch(SourcePacketExtension::isInjected);
         if (onlyInjected)
         {
-            logger.info("Suppressing source-remove for injected SSRC.");
+            logger.debug("Suppressing source-add for injected SSRC.");
             return;
         }
 
@@ -377,7 +377,7 @@ public abstract class AbstractOperationSetJingle
                         ssrcs.getSourcesForMedia("audio").stream().allMatch(SourcePacketExtension::isInjected);
         if (onlyInjected)
         {
-            logger.info("Suppressing source-remove for injected SSRC.");
+            logger.debug("Suppressing source-remove for injected SSRC.");
             return;
         }
 
--- a/src/main/java/org/jitsi/protocol/xmpp/AbstractOperationSetJingle.java
+++ b/src/main/java/org/jitsi/protocol/xmpp/AbstractOperationSetJingle.java
@@ -17,6 +17,7 @@
  */
 package org.jitsi.protocol.xmpp;
 
+import org.jitsi.impl.protocol.xmpp.*;
 import org.jitsi.jicofo.xmpp.*;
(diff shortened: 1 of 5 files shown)


src/main/java/org/jitsi/protocol/xmpp/AbstractOperationSetJingle.java
```diff
@@ -154,6 +159,7 @@ public boolean replaceTransport(
         }
 
         IQ reply = getConnection().sendPacketAndGetReply(jingleIQ);
+        stats.stanzaSent(jingleIQ.getAction());
```
JonathanLennox, line 162: Not that I understand jicofo, but `reply` can apparently be `null` here -- does that imply that a stanza was not in fact sent?

src/main/java/org/jitsi/protocol/xmpp/AbstractOperationSetJingle.java
```diff
@@ -262,21 +262,26 @@ public void sendAddSourceIQ(MediaSourceMap ssrcs,
...
                                 JingleSession        session)
     {
-        JingleIQ addSourceIq
-            = new JingleIQ(JingleAction.SOURCEADD, session.getSessionID());
+        boolean onlyInjected =
+                ssrcs.getSourcesForMedia("video").isEmpty() &&
+                        ssrcs.getSourcesForMedia("audio").stream().allMatch(SourcePacketExtension::isInjected);
+        if (onlyInjected)
+        {
+            logger.info("Suppressing source-remove for injected SSRC.");
```
JonathanLennox, line 270: Log message should say `"Suppressing source-add"` here presumably

JonathanLennox, line 270: Do we really want an INFO-level log for every one of these suppressions?  Isn't the whole reason we're doing this that there're way too many of them?

src/main/java/org/jitsi/protocol/xmpp/AbstractOperationSetJingle.java
bgrozev: More likely that a stanza was sent, but no response was received. I think "number of stanzas we tried to send" is a reasonable meaning for the counter.

src/main/kotlin/org/jitsi/jicofo/JicofoServices.kt
```diff
@@ -247,6 +248,7 @@ open class JicofoServices {
         jigasiDetector?.let { put("jigasi_detector", it.stats) }
         putAll(ColibriConferenceImpl.stats.toJson())
         put("threads", ManagementFactory.getThreadMXBean().threadCount)
+        put("jingle", AbstractOperationSetJingle.getStats())
```
JonathanLennox, line 251: Does this need a `toJson()` on it?

bgrozev: No, it's exposed as a `JSONObject` already.

JonathanLennox (approved):

PR #704, opened 2021-03-05
Refactor JigasiDetector and JibriDetector.

bgrozev:


src/main/java/org/jitsi/jicofo/jigasi/TranscriberManager.java
```diff
@@ -181,14 +182,15 @@ private TranscriptionStatusExtension getTranscriptionStatus(Presence p)
...
-    private void startTranscribing(Collection<String> preferredRegions)
+    private void startTranscribing(@NotNull Collection<String> preferredRegions)
     {
         if (active)
         {
             return;
         }
 
-        selectTranscriber(2, null, preferredRegions);
+        // We need a modifiable list for the "exclude" parameter.
```
bbaldino, line 192: Is this comment still the case?  `selectTranscriber` looks like it takes an immutable list.

src/main/kotlin/org/jitsi/jicofo/jigasi/JigasiDetector.kt
```diff
@@ -0,0 +1,117 @@
...
+        selectJigasi(instances, exclude, preferredRegions, config.localRegion, transcriber = true)
+
+    /**
+     * Selects a jigasi instance which supports SIP.
+     * @param exclude a list of excluded instances.
+     * @param preferredRegions a list of preferred regions.
+     * @return the JID of the selected instance, or  `null` if there are no available jigasis that satisfy the
+     * constraints (are not excluded and support SIP).
+     */
+    fun selectJigasi(exclude: List<Jid>, preferredRegions: Collection<String>): Jid? =
```
bbaldino, line 64: nit: looks like it would be nice to rename this to `selectSipJigasi` or something

src/main/kotlin/org/jitsi/jicofo/jigasi/JigasiDetector.kt
```diff
@@ -0,0 +1,117 @@
...
+            preferredRegions: Collection<String>,
+            localRegion: String?,
+            transcriber: Boolean
+        ): Jid? {
+
+            val availableInstances = instances.filter {
+                !exclude.contains(it.jid) &&
+                    !it.isInGracefulShutdown() &&
+                    if (transcriber) it.supportsTranscription() else it.supportsSip()
+            }
```
bbaldino, line 91: I don't feel strongly about this, but to me this reads a bit better with separate filter calls for each condition:
```
val availableInstances = instances
    .filter { !exclude.contains(it.jid) }
    .filter { !it.isInGracefulShutdown() }
    .filter { if (transcriber) it.supportsTranscription() else it.supportsSip() }
```

I also wondered if we should have the other functions pass a predicate instead of a boolean for the transcriber vs sip filter, so this function would become:

```
fun selectJigasi(
            instances: List<BaseBrewery<ColibriStatsExtension>.BrewInstance>,
            exclude: List<Jid>,
            preferredRegions: Collection<String>,
            localRegion: String?,
            additionalFilter: (BaseBrewery<ColibriStatsExtension>.BrewInstance) -> Boolean
        ): Jid? {

            val availableInstances = instances
                .filter { !exclude.contains(it.jid) }
                .filter { !it.isInGracefulShutdown() }
                .filter(additionalFilter)
...
```

And a calling function would be:
```
fun selectTranscriber(exclude: List<Jid>, preferredRegions: Collection<String>): Jid? =
        selectJigasi(instances, exclude, preferredRegions, config.localRegion) { it.supportsTranscription() }
```

But, again, just a thought.  I wouldn't call it a blocker.

src/main/kotlin/org/jitsi/jicofo/recording/jibri/JibriDetector.kt
```diff
@@ -0,0 +1,95 @@
...
+        get() = JSONObject().apply {
+            this["count"] = getInstanceCount(null)
+            this["available"] = getInstanceCount { it.status.isAvailable }
+        }
+
+    interface EventHandler {
+        fun instanceOffline(jid: Jid) {}
+    }
+
+    companion object {
```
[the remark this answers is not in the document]
bgrozev: Yep, I added that in a later commit.

src/main/java/org/jitsi/jicofo/jigasi/TranscriberManager.java
bgrozev: Yes. This is java, so it's not immutable. Line 251 adds to it...

What I did in this commit was replace this with a @NotNull annotation:
```
if (exclude == null)
{
    exclude = new ArrayList<>();
}
```

src/main/kotlin/org/jitsi/jicofo/jigasi/JigasiDetector.kt
bgrozev: I did the first part. I'll leave the second part for another time, because JigasiSelectorTest uses this static method in a bunch of places and it's a bit awkward to expose the predicates and use them in java.

src/main/java/org/jitsi/jicofo/jigasi/TranscriberManager.java
bbaldino: Oops.  Ok gotcha.

bbaldino (approved):

PR #710, opened 2021-03-23
Make sure exclude is not null

jerry2013:
When running with the recent refactor, I got the following in the log and was unable to dial out.

java.lang.IllegalArgumentException: Parameter specified as non-null is null: method org.jitsi.jicofo.jigasi.JigasiDetector.selectSipJigasi, parameter exclude


damencho (approved):

damencho:
Jenkins, test this please.

damencho:
Thank you.

PR #712, opened 2021-03-25
fix: Disables octo by default (sync with the bridge).

gpolitis:


damencho (approved):

damencho:
Maybe we need to fix the tests with this change: `Error:    OctoConfigTest expected:<true> but was:<false>`

damencho (approved):

PR #713, opened 2021-03-25
fix: Don't propagate sources if octo is disabled.

gpolitis:


src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
```diff
@@ -2807,12 +2812,17 @@ private void removeSourcesFromOcto(MediaSourceMap sources, MediaSourceGroupMap s
          */
         private void setRelays(List<String> allRelays)
```
bgrozev, line 2813: The only use of `setRelays` is from `updateOctoRelays`. Should we bubble up the check there?

bgrozev (dismissed):

bgrozev:
Left a minor question, but also approved. 👍

bgrozev (approved):

PR #714, opened 2021-03-25
Jibri clean up

bgrozev:


src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
```diff
@@ -2140,15 +2140,6 @@ public boolean handleMuteRequest(Jid fromJid, Jid toBeMutedJid, boolean doMute,
                     && participantChannels != null
                     && bridgeSession.colibriConference.muteParticipant(participantChannels, doMute, mediaType);
 
-        if (succeeded)
```
bbaldino, line 2143: (maybe explained later): how is it we can get rid of this muted state? were we not using it anywhere?

src/main/java/org/jitsi/impl/protocol/xmpp/XmppProviderImpl.java
```diff
@@ -108,7 +108,8 @@ public XmppProviderImpl(
 
         connection = createXmppConnection();
         connectRetry = new RetryStrategy(executor);
-        jibriIqHandler = new JibriIqHandler(connection);
+        jibriIqHandler = new JibriIqHandler();
+        connection.registerIQRequestHandler(jibriIqHandler);
```
bbaldino, line 112: This is awesome! (it working as a handler instead of requiring the connection)

src/main/kotlin/org/jitsi/jicofo/xmpp/JibriIqHandler.kt
```diff
@@ -0,0 +1,55 @@
...
+    fun removeJibri(jibri: CommonJibriStuff?) {
+        jibris.remove(jibri)
+    }
+
+    /**
+     * {@inheritDoc}
+     * Pass the request to the first [CommonJibriStuff] that wants it.
+     */
+    override fun handleIQRequest(iq: IQ): IQ = synchronized(jibris) {
+            iq as JibriIq
```
bbaldino: Do we need any potential error handling here if the iq is somehow not a JibriIq? Or is that effectively impossible?

src/main/kotlin/org/jitsi/jicofo/xmpp/JibriIqHandler.kt
```diff
@@ -0,0 +1,55 @@
...
+
+    fun removeJibri(jibri: CommonJibriStuff?) {
+        jibris.remove(jibri)
+    }
+
+    /**
+     * {@inheritDoc}
+     * Pass the request to the first [CommonJibriStuff] that wants it.
+     */
+    override fun handleIQRequest(iq: IQ): IQ = synchronized(jibris) {
```
bbaldino, line 51: Is this `synchronized(jibris)` necessary here? It's not used in `addJibri` or `removeJibri`.

src/main/java/org/jitsi/jicofo/recording/jibri/BaseJibriRecorder.java
```diff
@@ -41,7 +41,7 @@
  *
  * @author Pawel Domas
  */
-public abstract class CommonJibriStuff
+public abstract class BaseJibriRecorder
```
bbaldino: Does this name make sense given it's used for SIP gateway as well?

src/main/java/org/jitsi/jicofo/recording/jibri/JibriSession.java
```diff
@@ -179,6 +170,9 @@ static private boolean isStartingStatus(JibriIq.Status status)
      */
     private Jid terminator;
 
+    @NotNull
+    private final JibriStats stats = JibriStats.Companion.getGlobalStats();
```
bbaldino, line 176: This still has to use `.Companion` even with `JvmStatic`?

src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
bgrozev: Nope. It was only written, and I don't think it was consistent either.

src/main/java/org/jitsi/impl/protocol/xmpp/XmppProviderImpl.java
bgrozev: Yeah, because it needs to be registered to both providers in order to handle communication with jibri and the clients

src/main/kotlin/org/jitsi/jicofo/xmpp/JibriIqHandler.kt
bgrozev: It's supposed to be impossible, because it is registered with Smack only for those, and the existing code performed an unconditional cast, so I left it. We'd have to throw an exception anyway.

src/main/kotlin/org/jitsi/jicofo/xmpp/JibriIqHandler.kt
bgrozev: Yes, the `Collections.synchronizedList` docs specifically mention that it is required for iteration.

src/main/java/org/jitsi/jicofo/recording/jibri/BaseJibriRecorder.java
bgrozev: I'm open to better suggestions!

src/main/kotlin/org/jitsi/jicofo/xmpp/JibriIqHandler.kt
bbaldino: Wow, interesting...so it must do its own protection on that same lock inside of add and remove or something?  I'd say it's worth a comment here.  Or what about using `CopyOnWriteArrayList` instead?  It's weird to me that _only_ iteration requires synchronizing, I'd think it would be easy for someone to get tripped up when reading this.

src/main/kotlin/org/jitsi/jicofo/xmpp/JibriIqHandler.kt
bbaldino: Yeah...might be good to do `iq as? JibriIq ?: throw SomeException` just in case.

src/main/kotlin/org/jitsi/jicofo/xmpp/JibriIqHandler.kt
bgrozev: I just made that change, but I realized that each conference adds a handler here, even if jibri is not used, so under stress there will be a lot of copying.

bbaldino: The copy would just be on `addJibri`/`removeJibri`, which isn't related to the amount of handlers?  Or maybe I'm missing something.

bgrozev: The list is comparable in size to the number of conferences, and it will be copied any time a conferences is started/stoppe

bbaldino: Ok.  conference start/stop doesn't seem to me like a common enough operation that this would be a problem, but I won't insist on the change.  A comment then?  I just imagine it will be very tempting to think there's a synchronization issue here to someone who's reading it.

bbaldino (approved):

PR #716, opened 2021-03-26
Fixes blocking receiver thread

damencho:
I started doing it, but then I found that there are 3 more places like this that will need similar changes and those are the grantOwner, start Jibri and start transcriber.

Or we can change so that we have a pool of threads executing these tasks ... WDYT?

People were asking whether recording started taking a little bit longer, I guess the same is happening ...

### DIFF
--- a/src/main/java/org/jitsi/jicofo/xmpp/IqHandler.java
+++ b/src/main/java/org/jitsi/jicofo/xmpp/IqHandler.java
@@ -309,8 +309,6 @@ public class IqHandler
             return IQ.createErrorResponse(dialIq, XMPPError.getBuilder(XMPPError.Condition.not_allowed));
         }
 
-        final List<Jid> excludeList = exclude == null ? new ArrayList<>() : exclude;
-
         Set<String> bridgeRegions = conference.getBridges().keySet().stream()
             .map(Bridge::getRegion)
             .filter(Objects::nonNull)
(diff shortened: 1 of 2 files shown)


src/main/java/org/jitsi/jicofo/xmpp/IqHandler.java
```diff
@@ -309,6 +309,8 @@ private IQ handleRayoIQ(RayoIqProvider.DialIq dialIq, int retryCount, @NotNull L
             return IQ.createErrorResponse(dialIq, XMPPError.getBuilder(XMPPError.Condition.not_allowed));
         }
 
+        final List<Jid> excludeList = exclude == null ? new ArrayList<>() : exclude;
```
bgrozev, line 312: `exclude` is `@NotNull`

src/main/java/org/jitsi/jicofo/xmpp/IqHandler.java
```diff
@@ -334,31 +336,57 @@ private IQ handleRayoIQ(RayoIqProvider.DialIq dialIq, int retryCount, @NotNull L
...
+                    logger.error("Error sending dialIQ to "+ jigasiJid + " " + exception.getMessage());
+
+                    try
+                    {
+                        if (retryCount > 0)
+                        {
+                            excludeList.add(jigasiJid);
+
+                            // let's retry lowering the number of attempts
+                            IQ result = this.handleRayoIQ(dialIq, retryCount - 1, excludeList);
```
bgrozev, line 361: Looks like `this.handleRayoIQ` now unconditionally returns `null`

damencho: It can return few errors.

src/main/java/org/jitsi/jicofo/xmpp/IqHandler.java
damencho: Fixed it ... I was working on older version ...

damencho:
Actually, transcriber is using its own thread for executions ... and grand owner is also fine as it is communicating with prosody ...

bgrozev (approved):

PR #719, opened 2021-04-05
feat: Do not send source-remove when a participant leaves.

bgrozev:
~IMPORTANT: Before merging verify that mobile clients have been updated to remove tracks when a member leaves.~ (they have been)

### DIFF
--- a/src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
+++ b/src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
@@ -1108,7 +1108,8 @@ public class JitsiMeetConferenceImpl
                         leftParticipant,
                         Reason.GONE,
                         null,
-                        /* no need to send session-terminate - gone */ false);
+                        /* no need to send session-terminate - gone */ false,
+                        /* no need to send source-remove */ false);
             }
             else
             {
@@ -1131,7 +1132,8 @@ public class JitsiMeetConferenceImpl
             Participant participant,
             Reason reason,
             String message,
-            boolean sendSessionTerminate)
+            boolean sendSessionTerminate,
+            boolean sendSourceRemove)
     {
         logger.info(String.format(
                 "Terminating %s, reason: %s, send session-terminate: %s",
@@ -1151,7 +1153,8 @@ public class JitsiMeetConferenceImpl
                     jingleSession,
                     participant.getSourcesCopy(),
                     participant.getSourceGroupsCopy(),
-                    false /* no JVB update - will expire */);
+                    false /* no JVB update - will expire */,
+                    sendSourceRemove);
 
                 participant.setJingleSession(null);
             }
@@ -1395,7 +1398,7 @@ public class JitsiMeetConferenceImpl
 
         synchronized (participantLock)
         {
-            terminateParticipant(participant, null, null, /* do not send session-terminate */ false);
+            terminateParticipant(participant, null, null, /* do not send session-terminate */ false, true);
 
             if (restartRequested)
             {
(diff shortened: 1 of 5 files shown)


src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
```diff
@@ -1395,7 +1398,7 @@ public XMPPError onSessionTerminate(JingleSession session, JingleIQ iq)
 
         synchronized (participantLock)
         {
-            terminateParticipant(participant, null, null, /* do not send session-terminate */ false);
+            terminateParticipant(participant, null, null, /* do not send session-terminate */ false, true);
```
gpolitis, line 1401: just for uniformity maybe add the param name in a comment before the value?

gpolitis:
I couldn't grasp when we not *not* send source remove. Maybe it would be good to add a short description in the commit message?

src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
```diff
@@ -1832,25 +1837,27 @@ private XMPPError removeSources(
...
+                    .forEach(
+                            otherParticipant ->
+                            {
+                                if (otherParticipant.isSessionEstablished())
+                                {
+                                    jingle.sendRemoveSourceIQ(
+                                            removedSources,
+                                            removedGroups,
+                                            otherParticipant.getJingleSession());
+                                } else
```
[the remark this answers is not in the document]
bgrozev: Yep, that got autoformatted.

src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
bgrozev: I wish this was kotlin :)

bgrozev:
> I couldn't grasp when we _not_ send source remove. Maybe it would be good to add a short description in the commit message?

I added a comment, and I'll also add to the commit message when doing squash-merging.

JonathanLennox (approved):

src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
```diff
@@ -1395,7 +1400,12 @@ public XMPPError onSessionTerminate(JingleSession session, JingleIQ iq)
 
         synchronized (participantLock)
         {
-            terminateParticipant(participant, null, null, /* do not send session-terminate */ false);
+            terminateParticipant(
+                    participant,
+                    null,
+                    null,
+                    /* do not send session-terminate */ false,
+                    /* do send source-remove */ true);
```
gpolitis, line 1408: Can't we treat this case the same way as `onMemberLeft`?

src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
```diff
@@ -1832,25 +1844,28 @@ private XMPPError removeSources(
...
-                            otherParticipant.getJingleSession());
-                    }
-                    else
-                    {
-                        logger.warn("Remove source: no jingle session for " + participantJid);
-                        otherParticipant.scheduleSourcesToRemove(removedSources);
-                        otherParticipant.scheduleSourceGroupsToRemove(removedGroups);
-                    }
-                });
+        if (sendSourceRemove)
```
gpolitis, line 1847: Maybe negate this (& return) and move it above the logger statement?

src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
gpolitis: Soon..

src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
bgrozev: I thought I'd leave it, in case the participant sends a session-terminate, but does not leave the MUC. I don't know if thata happens in practice.

src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
bgrozev: I know this would save some indentation, but I prefer it like this. If we add more code to the function, the early `return` makes it easier to do something unindented.

bgrozev: e.g. this would probably be a bug:
```
if (!sendSourceRemove)
    return null;
    
 sendSourceRemove();
 doSomethingElse());   
 return null;
```

PR #721, opened 2021-04-07
feat: Adds a ping to the healthcheck.

damencho:
Adds a ping that will get its response in the single thread that handles the sync stanza listeners and will detect if something is blocking it.

### DIFF
--- a/src/main/java/org/jitsi/jicofo/health/JicofoHealthChecker.java
+++ b/src/main/java/org/jitsi/jicofo/health/JicofoHealthChecker.java
@@ -17,6 +17,7 @@ package org.jitsi.jicofo.health;
 
 import kotlin.*;
 import org.jitsi.health.*;
+import org.jitsi.impl.protocol.xmpp.*;
 import org.jitsi.jicofo.*;
 import org.jitsi.jicofo.bridge.*;
 import org.jitsi.jicofo.xmpp.*;
@@ -193,8 +194,9 @@ public class JicofoHealthChecker implements HealthCheckService
         CountDownLatch pingResponseWait = new CountDownLatch(1);
         Ping p = new Ping(JidCreate.bareFrom(XmppConfig.client.getXmppDomain()));
 
-        XMPPConnection connection = Objects.requireNonNull(JicofoServices.jicofoServicesSingleton).getXmppServices()
-            .getClientConnection().getXmppConnection();
+        XmppProvider provider = Objects.requireNonNull(JicofoServices.jicofoServicesSingleton).getXmppServices()
+            .getClientConnection();
+        XMPPConnection connection = provider.getXmppConnection();
         StanzaListener listener = packet -> pingResponseWait.countDown();
         try
         {
@@ -203,8 +205,7 @@ public class JicofoHealthChecker implements HealthCheckService
             );
             connection.sendStanza(p);
 
-            // the xmpp server is on localhost so 500ms is more than enough to receive the response
-            if (!pingResponseWait.await(500, TimeUnit.MILLISECONDS))
+            if (!pingResponseWait.await(provider.getConfig().getReplyTimeout().toMillis(), TimeUnit.MILLISECONDS))
             {
                 throw new RuntimeException("did not receive ping from the xmpp server");
             }
--- a/src/main/java/org/jitsi/jicofo/health/JicofoHealthChecker.java
+++ b/src/main/java/org/jitsi/jicofo/health/JicofoHealthChecker.java
@@ -205,7 +205,8 @@ public class JicofoHealthChecker implements HealthCheckService
             );
             connection.sendStanza(p);
 
-            if (!pingResponseWait.await(provider.getConfig().getReplyTimeout().toMillis(), TimeUnit.MILLISECONDS))
+            // will wait for 5 seconds to receive the ping
+            if (!pingResponseWait.await(5, TimeUnit.SECONDS))
             {
                 throw new RuntimeException("did not receive ping from the xmpp server");
             }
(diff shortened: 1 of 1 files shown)


paweldomas:
Usually the requests for conference being created (the first one seen for given conference name) takes longer than usually and I don't remember if are blocking. I'd recommend testing this throughly as it's a risky one.

Isn't it enough to check how long the conference creation took? Or what problem does this PR solve?

paweldomas:
Also if it's the `focusManager.getConference(roomName)` call that is stuck then ping will not detect it

damencho:
It detects whether the single-threaded executor in smack is blocked.

damencho:
> Usually the requests for conference being created (the first one seen for given conference name) takes longer than usually and I don't remember if are blocking. I'd recommend testing this throughly as it's a risky one.

Yeah, it waits for the creation of the room (response to the presence) and is executed in the smack reader thread but that code had not changes, I just add an extra packet for the check and its response ... 
I don't understand what worries you here, is that we would not receive the ping for 500ms on localhost?
I can adjust the time, for example in jvb the reply timeout is 5 seconds, where in jicofo and jigasi it is 15 seconds.

paweldomas:
> I don't understand what worries you here, is that we would not receive the ping for 500ms on localhost?

So the first thing that may not work here is the case where the health check will block itself (the conference request). This will not be detected by ping, because it's sent after and if this call returns.

Another thing is that 500ms is too short and even if the IQ processor is not blocked, but it's just busy creating the conference this limit will likely be exceeded.

Also there's the assumption that everyone deploys jicofo and prosody on the same host.

> I can adjust the time, for example in jvb the reply timeout is 5 seconds, where in jicofo and jigasi it is 15 seconds.

Right, so 5 seconds and 15 that is for regular IQs. What problem will adjusting the timeout solve here?

damencho:
> Another thing is that 500ms is too short and even if the IQ processor is not blocked, but it's just busy creating the conference this limit will likely be exceeded.

This is done after the conference is created, I don't see the connection here...

> Right, so 5 seconds and 15 that is for regular IQs. What problem will adjusting the timeout solve here?

That the wait will not be 500ms.

paweldomas:
>> Another thing is that 500ms is too short and even if the IQ processor is not blocked, but it's just busy creating the conference this limit will likely be exceeded.

> This is done after the conference is created, I don't see the connection here...

But a user can be creating a conference at the time when healtcheck tries the ping.

damencho:
> But a user can be creating a conference at the time when healtcheck tries the ping.

And? ... there is nothing that is blocking the creation of a conference. Creating a conference in healthcheck or from a user is just new muc created and joined, which is a presence we send and we receive its reply. The reply comes from a stanza collector which is executed in smacks reader thread. 

The ping is sent directly using the connection and we create a stanza listener which is added to the sync listeners and we check whether it is received, the wait is in a separate thread and the reply comes from a single-threaded executor, which is used for example for delivering presence updates on member joined.
There are no touchpoints between the ping and the creation of the room other than both received packets which are read from the reader thread in smack, but here we are blocking another thread. 

Where do you think ping can block the creation of a conference for a user?

paweldomas:
> Where do you think ping can block the creation of a conference for a user?

I'm not saying that ping can block conference creation, but that ping can be blocked by a normal conference creation request that is just taking more time. That is probably more of a problem with the timeout set to 500ms, otherwise I guess it's fine.

damencho:
Changed it to the reply timeout, whatever is configured.

paweldomas:
> Changed it to the reply timeout, whatever is configured.

But do you agree that what I'm writing about could be the case? Because I'm no longer sure I understand everything here.

damencho:
Well, I saw that the iqRequest handlers are sync mode and those are coming from another single thread and we block it till the  room is created, the presence response is delivered. We do not block the smacks reader thread. So I don't think creating of the rooms block in anyway ping. But I agree about the assumption the jicofo and prosody are on the same machine, and waiting for the timeout is also fine, it is longer ... hum ... it will miss shorter blockages ... for example because of dialout we saw it was blocked for 10 seconds, and if the replyTimeout is 15 seconds healthcheck will succeed and will miss that ...  @bgrozev WDYT, as you were also looking at this?

bgrozev:
Ahh, now that ConferenceIq is handled in the smack sync thread, we end up executing JitsiMeetConferenceImpl#start there (which blocks waiting to join the MUC). This can explain the delays Pawel is seeing (I think we saw some indirect evidence of these too in prod). We'll have to solve this somehow, probably by offloading the smack sync thread again.

I think doing a ping in the health checker is fine -- if something blocks it for 5 seconds that's clearly a problem whatever the cause.

paweldomas:
> Well, I saw that the iqRequest handlers are sync mode and those are coming from another single thread and we block it till the room is created, the presence response is delivered. We do not block the smacks reader thread.

If conference/muc queues are separate from ping then I don't know what the ping is supposed to detect.

damencho:
> Ahh, now that ConferenceIq is handled in the smack sync thread, we end up executing JitsiMeetConferenceImpl#start there (which blocks waiting to join the MUC).

Ah, I see where the confusion comes from (confusing me), so I was talking about presence listeners which are smack sync listeners https://github.com/jitsi/Smack/blob/4fa73bde03b27b9166809ca6c2e5ebafe7d844b7/smack-extensions/src/main/java/org/jivesoftware/smackx/muc/MultiUserChat.java#L315 which are executed in the same thread as the ConferenceIqs, we want to detect whether something is blocking that thread.
So if that thread is blocked the sync listener we add with this change will also be blocked.

@bgrozev So the point here is that this PR is supposed to detect those 10 sec locks we saw ... but waiting 15 second for the ping will actually miss them. I will lower the wait time to 5 secs, wdyt?

bgrozev:
This PR makes sense to me. @paweldomas what do you think?

I'll be working separately on handing ConferenceIq in a different thread.

bgrozev (approved):

PR #725, opened 2021-04-16
Handle ConferenceIq in the IO pool (and cleanup).

bgrozev:


src/test/java/org/jitsi/jicofo/OSGiHandler.java
```diff
@@ -72,6 +73,12 @@ public MockXmppProvider getXmppProvider()
 
     public void shutdown()
     {
+        System.clearProperty("org.jitsi.jicofo.PING_INTERVAL");
+        System.clearProperty(XmppClientConnectionConfig.legacyXmppDomainPropertyName);
+        System.clearProperty(XmppClientConnectionConfig.legacyDomainPropertyName);
+        System.clearProperty(XmppClientConnectionConfig.legacyUsernamePropertyName);
+        JitsiConfig.Companion.reloadNewConfig();
+
```
JonathanLennox, line 81: This doesn't logically seem part of the "Move XmppTest" commit - what's it doing?

src/main/kotlin/org/jitsi/jicofo/JicofoServices.kt
```diff
@@ -201,7 +184,7 @@ open class JicofoServices {
             it.stop()
         }
         healthChecker?.stop()
-        channelAllocationExecutor.shutdownNow()
+        TaskPools.ioPool.shutdownNow()
```
JonathanLennox, line 179: I feel like it'd be cleaner to have a single shutdown method in the `TaskPools` companion object.

spotbugs-exclude.xml
```diff
@@ -26,6 +26,7 @@
     <Or>
       <Bug pattern="RV_RETURN_VALUE_IGNORED_NO_SIDE_EFFECT"/>
       <Bug pattern="NP_NULL_ON_SOME_PATH_FROM_RETURN_VALUE"/>
+      <Bug pattern="DE_MIGHT_IGNORE"/>
```
JonathanLennox, line 29: What warning are you suppressing here?

src/test/java/org/jitsi/jicofo/OSGiHandler.java
bgrozev: Moving the test changed the order in which tests are run, which exposed a bug with some other tests changing the configuration. Now that I look at it more carefully (though I still don't understand it fully), it should probably be JicofoConfigTest that needs to clean-up after itself.

spotbugs-exclude.xml
bgrozev: It doesn't need to be suppressed anymore, removing.

src/test/java/org/jitsi/jicofo/OSGiHandler.java
bgrozev: Aha. It used to be that `OSGiHandler.init` never ran before `XmppConfigTest`. But now `AllocateConferenceTest` runs prior to `XmppConfigTest`, and if `OSGiHandler` leaves these properties behind it brakes things.

JonathanLennox (approved):

PR #734, opened 2021-04-29
A/V moderation

damencho:


src/main/java/org/jitsi/impl/protocol/xmpp/ChatRoomImpl.java
```diff
@@ -844,6 +852,51 @@ public void processPresence(Presence presence)
...
+
+        // there is no whitelists received yet
+        if (this.whitelists == null)
+        {
+            return false;
+        }
+
+        List<String> whitelist = this.whitelists.get(mediaType.toString());
+
+        return whitelist == null ? false : this.whitelists.get(mediaType.toString()).contains(jid.toString());
```
bgrozev, line 897: `this.whitelists.get(mediaType.toString())` is already saved in the local `whitelist`.

src/main/java/org/jitsi/impl/protocol/xmpp/ChatRoomImpl.java
```diff
@@ -844,6 +852,51 @@ public void processPresence(Presence presence)
...
+     */
+    public boolean isMemberAllowedToUnmute(Jid jid, MediaType mediaType)
+    {
+        if (!this.isAvModerationEnabled())
+        {
+            return true;
+        }
+
+        // there is no whitelists received yet
+        if (this.whitelists == null)
```
bgrozev, line 890: Can we make that `@NotNull`?

src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
```diff
@@ -2115,11 +2115,25 @@ public MuteResult handleMuteRequest(Jid muterJid, Jid toBeMutedJid, boolean doMu
...
+        if (!doMute)
         {
-            logger.warn("Unmute now allowed, mutedJid=" + muterJid + ", toBeMutedJid=" + toBeMutedJid);
-            return MuteResult.NOT_ALLOWED;
+            // when A/V moderation is enabled we need to check the whitelists
+            if (this.chatRoom.isAvModerationEnabled())
+            {
+                if (!this.chatRoom.isMemberAllowedToUnmute(toBeMutedJid, mediaType))
+                {
+                    logger.warn("Unmute now allowed due to av moderation, mutedJid="
```
bgrozev, line 2126: s/now/not/

src/main/kotlin/org/jitsi/jicofo/xmpp/XmppServices.kt
```diff
@@ -45,6 +58,11 @@ class XmppServices(xmppProviderFactory: XmppProviderFactory) {
         setOf(clientConnection.xmppConnection, serviceConnection.xmppConnection)
     )
 
+    val avModerationHandler = AvModerationHandler(clientConnection.xmppConnection).also {
+        clientConnection.xmppConnection.addAsyncStanzaListener(it, MessageTypeFilter.NORMAL)
+        clientConnection.addRegistrationListener(it)
```
damencho, line 63: @bgrozev I was wondering is it possible for the connection to be already registered when we do this and skip the disco-info?

damencho: I added a check for that ... not sure how good it is with no synchronization ...

damencho: Aaaand I broke the tests with that :)

damencho:
```
SEVERE: Error checking for av_moderation component
org.jivesoftware.smack.SmackException$NoResponseException: No response received within reply timeout. Timeout was 15000ms (~15s). Waited for response using: IQReplyFilter: iqAndIdFilter (AndFilter: (OrFilter: (IQTypeFilter: type=error, IQTypeFilter: type=result), StanzaIdFilter: id=hP27l-141)), : fromFilter (OrFilter: (FromMatchesFilter (full): test.domain.net)).
	at org.jivesoftware.smack.SmackException$NoResponseException.newWith(SmackException.java:111)
	at org.jivesoftware.smack.SmackException$NoResponseException.newWith(SmackException.java:98)
	at org.jivesoftware.smack.StanzaCollector.nextResultOrThrow(StanzaCollector.java:260)
	at org.jivesoftware.smack.StanzaCollector.nextResultOrThrow(StanzaCollector.java:214)
	at org.jivesoftware.smackx.disco.ServiceDiscoveryManager.discoverInfo(ServiceDiscoveryManager.java:540)
	at org.jivesoftware.smackx.disco.ServiceDiscoveryManager.discoverInfo(ServiceDiscoveryManager.java:506)
	at org.jitsi.jicofo.xmpp.AvModerationHandler.registrationChanged(XmppServices.kt:154)
	at org.jitsi.jicofo.xmpp.XmppServices.<init>(XmppServices.kt:64)
```

@bgrozev as you are playing with the tests, is there a way to skip this step when testing?

damencho: I dropped that commit. 

But now I see Mute test is broken, so there must be some problem ... will be checking that.

damencho: And that's only in FF case ... hum is this what you were seeing yesterday for some other PR?

damencho: Oh that is a problem in the bridge not forwarding the video stream, and as audio is muted nothing is received and it fails

damencho: Yep, retrying it, worked, it is now just the known desktop sharing one.

src/main/java/org/jitsi/impl/protocol/xmpp/ChatRoomImpl.java
```diff
@@ -844,6 +852,45 @@ public void processPresence(Presence presence)
...
+     * {@inheritDoc}
+     */
+    public boolean isMemberAllowedToUnmute(Jid jid, MediaType mediaType)
+    {
+        if (!this.isAvModerationEnabled())
+        {
+            return true;
+        }
+
+        List<String> whitelist = this.whitelists.get(mediaType.toString());
```
bgrozev, line 889: Can we make the `whitelist` field `@NotNull` (it's already effectively not null) instead of doing null checks?

src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
```diff
@@ -2115,11 +2115,25 @@ public MuteResult handleMuteRequest(Jid muterJid, Jid toBeMutedJid, boolean doMu
...
+        if (!doMute)
         {
-            logger.warn("Unmute now allowed, mutedJid=" + muterJid + ", toBeMutedJid=" + toBeMutedJid);
-            return MuteResult.NOT_ALLOWED;
+            // when A/V moderation is enabled we need to check the whitelists
+            if (this.chatRoom.isAvModerationEnabled())
+            {
+                if (!this.chatRoom.isMemberAllowedToUnmute(toBeMutedJid, mediaType))
+                {
+                    logger.warn("Unmute not allowed due to av moderation, mutedJid="
```
bgrozev, line 2126: s/mutedJid/muterJid/

src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
```diff
@@ -2115,11 +2115,25 @@ public MuteResult handleMuteRequest(Jid muterJid, Jid toBeMutedJid, boolean doMu
...
+                        + muterJid + ", toBeMutedJid=" + toBeMutedJid);
+                    return MuteResult.NOT_ALLOWED;
+                }
+            }
+            // do not allow unmuting other participants even for the moderator
+            else if (!muterJid.equals(toBeMutedJid))
```
bgrozev, line 2132: It is strange that this behavior is different based on whether moderation is enabled or not.

For the purpose of moderated meetings, clients need to expect that their channel could be unmuted sever-side, thus they must make sure not to send media unless the user initiates it. If they already have that logic, we may as well separate the local mute and the server-side mute, and allow moderators to do the latter. 

But (even if you agree) this is probably outside the scope of this PR.

bgrozev: Still, we want to double make sure that clients can not be unmuted remotely without the user's consent in the case of moderation (which includes e.g. someone turning moderation on/off at the right time)

src/main/kotlin/org/jitsi/jicofo/xmpp/XmppServices.kt
```diff
@@ -80,3 +99,62 @@ class XmppServices(xmppProviderFactory: XmppProviderFactory) {
...
 
 enum class XmppConnectionEnum { Client, Service }
+
+class AvModerationHandler(val xmppConnection: AbstractXMPPConnection) : StanzaListener, RegistrationListener {
+    var conferenceStore: ConferenceStore = EmptyConferenceStore()
+    private val jsonParser = JSONParser()
+    private var avModerationAddress: String? = null
+
+    override fun processStanza(stanza: Stanza) {
+        if (!stanza.from.equals(avModerationAddress)) {
```
bgrozev, line 109: `stanza.from != avModerationAddress`? Can `stanza.from` ever be `null`?

src/main/kotlin/org/jitsi/jicofo/xmpp/XmppServices.kt
```diff
@@ -80,3 +99,62 @@ class XmppServices(xmppProviderFactory: XmppProviderFactory) {
...
+
+        val jsonMessage = stanza.getExtension<JsonMessageExtension>(
+            JsonMessageExtension.ELEMENT_NAME, JsonMessageExtension.NAMESPACE
+        ) ?: return Unit.also {
+            logger.warn("Skip processing stanza without JsonMessageExtension")
+        }
+
+        val incomingJson = jsonParser.parse(jsonMessage.json) as JSONObject
+        if (incomingJson["type"] == "av_moderation") {
+            val conferenceJid = JidCreate.entityBareFrom(incomingJson["room"] as String)
```
bgrozev, line 121: I think `.toString()` would be safer

src/main/kotlin/org/jitsi/jicofo/xmpp/XmppServices.kt
```diff
@@ -80,3 +99,62 @@ class XmppServices(xmppProviderFactory: XmppProviderFactory) {
...
+
+            val conference = conferenceStore.getConference(conferenceJid) ?: return Unit.also {
+                logger.warn("Not processing message for invalid conferenceJid=$conferenceJid")
+            }
+
+            val enabled = incomingJson["enabled"]
```
bgrozev, line 127: Why do we encode a boolean as a string in json?

src/main/kotlin/org/jitsi/jicofo/xmpp/XmppServices.kt
```diff
@@ -80,3 +99,62 @@ class XmppServices(xmppProviderFactory: XmppProviderFactory) {
...
+            val lists = incomingJson["whitelists"] as JSONObject?
+
+            if (enabled != null) {
+                conference.chatRoom.isAvModerationEnabled = (enabled as String).toBoolean()
+            } else if (lists != null) {
```
bgrozev, line 132: Why do we not support updating the lists when `enabled` is present?

src/main/kotlin/org/jitsi/jicofo/xmpp/XmppServices.kt
bgrozev: > @bgrozev I was wondering is it possible for the connection to be already registered when we do this and skip the disco-info?

Yes, it is. For now you can add something like `it.registrationChanged(clientConnection.isRegistered())` which should make the probability of the race very low. I'll probably change this soon and I'll have it in mind.

src/main/java/org/jitsi/impl/protocol/xmpp/ChatRoomImpl.java
damencho: This is what is received from the client and the list indeed can be null.

src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
damencho: We need to talk about it ... I'm confused.

src/main/kotlin/org/jitsi/jicofo/xmpp/XmppServices.kt
damencho: I don't think so, but I will add the check.
using equals is intentional as it is overloaded equals of DomainPartJid comparing with a string. (kotlin gives an error with !=)

src/main/kotlin/org/jitsi/jicofo/xmpp/XmppServices.kt
```diff
@@ -80,3 +99,62 @@ class XmppServices(xmppProviderFactory: XmppProviderFactory) {
...
+            return
+        }
+
+        val jsonMessage = stanza.getExtension<JsonMessageExtension>(
+            JsonMessageExtension.ELEMENT_NAME, JsonMessageExtension.NAMESPACE
+        ) ?: return Unit.also {
+            logger.warn("Skip processing stanza without JsonMessageExtension")
+        }
+
+        val incomingJson = jsonParser.parse(jsonMessage.json) as JSONObject
```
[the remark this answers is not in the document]
damencho: 👍

src/main/kotlin/org/jitsi/jicofo/xmpp/XmppServices.kt
damencho: Ah its coming from lua ... as we take it directly from the stanza, I will change that

src/main/kotlin/org/jitsi/jicofo/xmpp/XmppServices.kt
damencho: It's two separate/different messages coming from prosody ... and these are two different logical steps in the usecase.

src/main/java/org/jitsi/impl/protocol/xmpp/ChatRoomImpl.java
bgrozev: Ah, of course, I misread.

src/main/kotlin/org/jitsi/jicofo/xmpp/XmppServices.kt
bgrozev: Ah okay. Why do we keep the address as a String then?

damencho: Ok, I will make it jid ... it used to be jid but I move it several times while debugging an issue

src/main/kotlin/org/jitsi/jicofo/xmpp/XmppServices.kt
damencho: Done

src/main/kotlin/org/jitsi/jicofo/xmpp/XmppServices.kt
```diff
@@ -80,3 +99,62 @@ class XmppServices(xmppProviderFactory: XmppProviderFactory) {
...
+        if (incomingJson["type"] == "av_moderation") {
+            val conferenceJid = JidCreate.entityBareFrom(incomingJson["room"] as String)
+
+            val conference = conferenceStore.getConference(conferenceJid) ?: return Unit.also {
+                logger.warn("Not processing message for invalid conferenceJid=$conferenceJid")
```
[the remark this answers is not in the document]
damencho: Fixed

src/main/kotlin/org/jitsi/jicofo/xmpp/XmppServices.kt
damencho: Fixed

src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
damencho: Fixed

src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
damencho: It is just this comment that had left from this PR.

damencho: Ok, it takes me a day to get it, but in the end it came to me :) 
I think I fixed it.

damencho:
I think that all comments across the PRs are fixed, but I realized there is one flow missing in the logic ...  but I think we need to discuss it, I have a simple resolution of it, but we need to take care of the concurrency ... so let's discuss it offline on Monday.

src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
```diff
@@ -2154,6 +2170,69 @@ public MuteResult handleMuteRequest(Jid muterJid, Jid toBeMutedJid, boolean doMu
         return succeeded ? MuteResult.SUCCESS : MuteResult.ERROR;
     }
 
+    /**
+     * {@inheritDoc}
+     */
+    public void muteAllNonModeratorParticipants(MediaType mediaType)
+    {
+        for (Participant participant : participants)
```
bgrozev, line 2178: I think we need the lock on `participantsLock` here, but we should not hold it while waiting for a response from the bridge.

src/main/java/org/jitsi/impl/protocol/xmpp/ChatRoom.java
```diff
@@ -198,4 +199,40 @@ void join()
...
+     * @param value the new value.
+     */
+    void setAvModerationEnabled(Jid actorJid, MediaType mediaType, boolean value);
+
+    /**
+     * Returns the actor that performed the last enable/disable action.
+     * @param mediaType the media type.
+     * @return the jid of the actor.
+     */
+    Jid getAvModerationActor(MediaType mediaType);
```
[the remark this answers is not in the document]
damencho: This is the participant we had stored, that had enabled the AV moderation for that mediaType.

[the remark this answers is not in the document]
damencho: And yes, this is to be passed to handleMuteRequest only. As it can happen when that is enabled, but we store it for all participants which join later and we need to use it later.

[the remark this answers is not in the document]
damencho: Hum, good point ... which I'm not sure how to handle that ...

src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
```diff
@@ -2154,6 +2170,69 @@ public MuteResult handleMuteRequest(Jid muterJid, Jid toBeMutedJid, boolean doMu
...
+    }
+
+    /**
+     * Mutes a participant.
+     * @param participant the participant to mute and is not moderator and hasn't been already muted.
+     * @param mediaType the media type for the operation.
+     */
+    public void muteParticipant(Participant participant, MediaType mediaType)
+    {
+        boolean isMuterModerator = participant.getChatMember().getRole().hasModeratorRights();
```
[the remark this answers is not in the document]
bgrozev: Why is this specific to "initial av moderation"? Can we just check if it's muted (we even have a TODO to keep the muted state somewhere, right?)?

src/main/java/org/jitsi/jicofo/ParticipantChannelAllocator.java
```diff
@@ -166,6 +166,17 @@ else if (reInvite)
                     participant.getSourcesCopy(),
                     participant.getSourceGroupsCopy());
         }
+
+        // if participant is not av moderated but we need to let's do it
+        if (meetConference.getChatRoom().isAvModerationEnabled(MediaType.AUDIO))
```
bgrozev, line 171: I assume we already have the correct `role` for the participant at this stage?

src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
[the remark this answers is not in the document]
damencho: So the one that enabled the moderation, and then the client see that the same participant has muted them. 
After we mute on the bridge side we send the muteIQ to clients, and clients see the notification the participant X has muted them, the same way it was not a force mute, but a regular remote mute.

src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
damencho: It is done in muteParticipant. You want me to rename those?

src/main/kotlin/org/jitsi/jicofo/xmpp/XmppServices.kt
```diff
@@ -61,6 +77,11 @@ class XmppServices(xmppProviderFactory: XmppProviderFactory) {
         )
     } else null
 
+    private val avModerationHandler = AvModerationHandler(clientConnection.xmppConnection).also {
+        clientConnection.xmppConnection.addAsyncStanzaListener(it, MessageTypeFilter.NORMAL)
```
bgrozev, line 81: I think we should use a sync listener and then offload to one of our IO threads.

src/main/kotlin/org/jitsi/jicofo/xmpp/XmppServices.kt
damencho: I added it, but it breaks the test. Will leave it there commented.

src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
```diff
@@ -2115,11 +2120,23 @@ public MuteResult handleMuteRequest(Jid muterJid, Jid toBeMutedJid, boolean doMu
...
 
-        // do not allow unmuting other participants even for the moderator
-        if (!doMute && !muterJid.equals(toBeMutedJid))
+        // process unmuting
+        if (!doMute)
         {
-            logger.warn("Unmute now allowed, mutedJid=" + muterJid + ", toBeMutedJid=" + toBeMutedJid);
-            return MuteResult.NOT_ALLOWED;
+            // do not allow unmuting other participants even for the moderator
+            if (muterJid != null && !muterJid.equals(toBeMutedJid))
```
bgrozev, line 2127: I think this should be `muterJid == null || !muterJir.equals(toBeMutedJid)` to allow unmute only when a participant unmutes themselves.

src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
```diff
@@ -2115,11 +2120,23 @@ public MuteResult handleMuteRequest(Jid muterJid, Jid toBeMutedJid, boolean doMu
...
+            {
+                logger.warn("Unmute not allowed, muterJid=" + muterJid + ", toBeMutedJid=" + toBeMutedJid);
+                return MuteResult.NOT_ALLOWED;
+            }
+            else if (this.chatRoom.isAvModerationEnabled(mediaType))
```
bgrozev, line 2132: This first check is redundant, since `isMemberAllowedToUnmute` performs it.

src/main/kotlin/org/jitsi/jicofo/xmpp/AvModerationHandler.kt
```diff
@@ -0,0 +1,120 @@
...
+ * Adds the A/V moderation handling. Process incoming messages and when audio or video moderation is enabled,
+ * muted all participants in the meeting (that are not moderators). Moderators are always allowed to unmute.
+ */
+class AvModerationHandler(val xmppProvider: XmppProvider) : RegistrationListener {
+    var conferenceStore: ConferenceStore = EmptyConferenceStore()
+    private val jsonParser = JSONParser()
+    private var avModerationAddress: DomainBareJid? = null
+
+    init {
+        xmppProvider.xmppConnection.addAsyncStanzaListener(
```
bgrozev, line 47: I think this should be `sync`

src/main/kotlin/org/jitsi/jicofo/xmpp/AvModerationHandler.kt
```diff
@@ -0,0 +1,120 @@
...
+                val conference = conferenceStore.getConference(conferenceJid) ?: return Unit.also {
+                    logger.warn("Not processing message for not existing conference conferenceJid=$conferenceJid")
+                }
+
+                val enabled = incomingJson["enabled"] as Boolean?
+                val lists = incomingJson["whitelists"] as JSONObject?
+
+                if (enabled != null) {
+                    val mediaType = MediaType.parseString(incomingJson["mediaType"] as String)
+                    val oldEnabledValue = conference.chatRoom.isAvModerationEnabled(mediaType)
```
bgrozev, line 83: nit: I think it would be useful to have logs when moderation is enabled

src/main/java/org/jitsi/impl/protocol/xmpp/colibri/ColibriConferenceImpl.java
```diff
@@ -687,11 +687,32 @@ public boolean muteParticipant(ColibriConferenceIQ channelsInfo, boolean mute, M
         }
 
         request.addContent(requestContent);
-        UtilKt.tryToSendStanza(connection, request);
+        try
+        {
+            IQ response = UtilKt.sendIqAndGetResponse(connection, request);
```
bgrozev, line 692: This makes `muteParticipant` blocking. I can make the mute IQ handler async if you want.

damencho: Well, I check that the iq handlers are executed async in a smacks cached thread pool.

bgrozev (dismissed):

bgrozev (approved):

PR #742, opened 2021-05-20
fix: Catch exceptions in IQ handlers.

bgrozev:


bgrozev:
@paweldomas you brought up the need for these changes, can you please review them?

paweldomas (approved):

PR #744, opened 2021-05-25
Fix jigasi iq handler

jerry2013:
I don't know what I'm doing - all I know is that jigasi couldn't find any conference given it's using EmptyConferenceStore to do lookup (log shows "rejected ... non-existent conference").

The patch appears to work fine for me. Please advice.


bgrozev:
Thanks for the fix!

Jenkins, please test this please

bgrozev (approved):

PR #747, opened 2021-06-08
Use the service connection (if configured) for the Jigasi detector

jbg:
When Jicofo is configured to use the service XMPP connection for Jigasi, it still accesses the Jigasi brewery MUC using the client connection. This patch fixes it to use the connection selected in jicofo.conf for jigasi.

### DIFF
--- a/src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
+++ b/src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
@@ -439,7 +439,9 @@ public class JitsiMeetConferenceImpl
         rolesAndPresence.init();
 
         transcriberManager = new TranscriberManager(
-            getClientXmppProvider(),
+            jicofoServices.getXmppServices().getXmppConnectionByName(
+                JigasiConfig.config.xmppConnectionName()
+            ),
             this,
             jicofoServices.getXmppServices().getJigasiDetector(),
             logger);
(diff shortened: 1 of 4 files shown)


jbg:
CLA already signed previously.

bgrozev:
Thanks for the fix. But it needs to use the connection configured for jigasi, [like we do for jibri](https://github.com/jitsi/jicofo/blob/master/src/main/kotlin/org/jitsi/jicofo/JicofoServices.kt#L122), not just the service connection unconditionally.

jbg:
Ah yeah, sorry, I misinterpreted [the lines above](https://github.com/jitsi/jicofo/pull/747/files#diff-dcafb0e3062273505592f993acc661f425b57856453242e9cc701bc1ac2336e1L44) as taking care of that but they are actually just checking if a service connection is defined at all. I'll update it to use the connection configured for Jigasi a bit later, thanks.

bgrozev:
@jbg do you still plan to make the change?

jbg:
@bgrozev Sorry for the delay, have been busy with other priorities. I'll try to get this PR updated tomorrow.

jbg:
@bgrozev I've updated this to use the configured connection now. In testing I also noticed that the client connection was used unconditionally for contacting Jigasi when starting transcription; I've resolved that now too.

bgrozev (approved):

bgrozev:
@jbg thanks, good catch!

Jenkins, please test this please

bgrozev:
Test failures are unrelated

PR #748, opened 2021-06-14
ref: Make JitsiMeetConference and ChatRoom more separate

bgrozev:
This PR makes some progress towards using an API and a single API between ChatRoomImpl and the code that uses it (JitsiMeetConference, BaseBrewery, etc). The goal is to move the XMPP/MUC logic out of JitsiMeetConference and separate ChatRoom.

This is not complete, because I found that the next step requires us to read endpoint supported features from presence, which will be another PR.


src/main/java/org/jitsi/impl/protocol/xmpp/ChatRoomImpl.java
```diff
@@ -116,9 +116,6 @@
     private final CopyOnWriteArrayList<ChatRoomLocalUserRoleListener> localUserRoleListeners
             = new CopyOnWriteArrayList<>();
 
-    /**
-     * Nickname to member impl class map.
-     */
```
JonathanLennox, line 122: Why remove this doc?

src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
```diff
@@ -539,7 +539,7 @@ private void leaveTheRoom()
      *
      * @param chatRoomMember the new member that has just joined the room.
      */
-    protected void onMemberJoined(final ChatRoomMember chatRoomMember)
+    private void onMemberJoined(final ChatRoomMember chatRoomMember)
```
JonathanLennox, line 542: Should this `chatRoomMember` be annotated `@NotNull` now too?

src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
```diff
@@ -529,7 +529,7 @@ private void leaveTheRoom()
             chatRoom.leave();
         }
 
-        chatRoom.setListener(null);
+        chatRoom.removeListener(null);
```
JonathanLennox, line 532: The argument here should be `chatRoomListener` not `null`.

src/main/java/org/jitsi/impl/protocol/xmpp/ChatRoomImpl.java
bgrozev: Because it's wrong (the keys are the full JIDs, not just the nickname).

src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
bgrozev: Good catch, I also annotated the interface.

src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
bgrozev: Sure. I removed the `final`, since we don't usually use it and I don't see a special reason to have it.

JonathanLennox (approved):

PR #749, opened 2021-06-15
Recognize jigasi by presence (not disco#info).

bgrozev:


JonathanLennox:
Is this authenticated? Will anything bad happen if an endpoint falsely claims to be a Jigasi?

bgrozev:
> Is this authenticated? Will anything bad happen if an endpoint falsely claims to be a Jigasi?

It is not authenticated, but neither was disco#info.

bgrozev:
Thanks!

PR #751, opened 2021-06-16
feat: Only recognize jibri members from trusted domains.

bgrozev:


src/main/resources/reference.conf
```diff
@@ -301,5 +301,7 @@ jicofo {
       // Only disable this if your xmpp connection is on loopback!
       use-tls = true
     }
+
+    trusted-domains = []
```
JonathanLennox, line 305: Have a comment here, i.e. trusted to do what?

src/main/resources/reference.conf
```diff
@@ -301,5 +301,9 @@ jicofo {
       // Only disable this if your xmpp connection is on loopback!
       use-tls = true
     }
+
+    // The list of domains with trusted services. Only members logged in to these domains can declare themselves to be
+    // Jibri instances.
+    trusted-domains = []
```
damencho, line 307: Can we change it to default to `recorder .${xmpp.domain}` so we avoid breaking all deployments on next update?

damencho (approved):

PR #753, opened 2021-06-21
ref: Simplify ChatRoomImpl, remove ChatRoomRoleAndPresence.'

bgrozev:


bgrozev:
This is accidentally on top of #751, I'll rebase when it is merged.

src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
```diff
@@ -534,12 +534,6 @@ private void leaveTheRoom()
         chatRoom = null;
     }
 
-    /**
-     * Method called by {@link #rolesAndPresence} when new member joins
-     * the conference room.
-     *
-     * @param chatRoomMember the new member that has just joined the room.
-     */
     private void onMemberJoined(@NotNull ChatRoomMember chatRoomMember)
```
damencho, line 545: Why removing javadocs from code? This is already enforced in all js code maybe we need to see whether we can enforce it with checkstyle plugin we use. 

https://jitsi.github.io/handbook/docs/dev-guide/dev-guide-contributing#comments

damencho (commented):
Is it on purpose that Kotlin code is without any docs? 
We should be documenting all methods and classes everywhere.

src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
bgrozev: I removed it, because it is no longer correct. I also think we should document what the method does, not which class calls it. I didn't write a new description, because I didn't see anything useful I could add. I added a doc now.

bgrozev:
Force-pushed to rebase on top of master

bgrozev:
> Is it on purpose that Kotlin code is without any docs?

No, I just added docs for the code I introduced.

> We should be documenting all methods and classes everywhere.

I strongly disagree. I know we have a convention for java, so I added docs for the method you mentioned, but we don't have the same policy for kotlin and I think we are better off this way. 

I agree that we need good documentation, but I don't think requiring it for methods and classes is a good way to achieve it. We can have a separate discussion about it, but here are a few points:
* The "methods and classes" line is arbitrary. Do we require it for fields? For method parameters? For variables defined in a function? When they are lambdas?
* Useless documentation has it's price. Documenting `public void setId(String id)` with `Sets the ID. @param id the id` adds nothing useful, but it takes space, wastes time for the person writing the code and wastes time for the person reading the code.
* Useless docs make it easier for docs to become stale (inaccurate).
* An abundance of useless/inaccurate docs makes people ignore them. 
* Mandatory docs lead to authors leaving useless docs just to check a box, and to reviewers not checking that the docs make sense (since the box is checked).


It would be interesting to somehow compare the quality of the docs in our java code vs kotlin.

damencho:
I agree with some of the points you mentioned, but there were two classes with zero docs. I see you have added docs there, and some of them have very useful information now.

damencho (approved):

bgrozev:
> but there were two classes with zero docs

Absolutely agree, and I appreciate you mentioning it! I think that's the right way to do docs -- if something needs them bring it up

PR #759, opened 2021-07-10
feat(breakout-rooms): add breakout rooms support

wernf:
This patch adds support for the breakout rooms feature. 

It allows participants to start breakout room conferences without authentication. Access to breakout rooms is assured by uuids. The prosody breakout_rooms_muc module only allows joining breakout rooms which were previously created by moderators.

Additionally it allows participants to join/start the main room when all participants are in breakout rooms.

This is a dependency of [this](https://github.com/jitsi/jitsi-meet/pull/9131) jitsi-meet pull request.

### DIFF
--- a/src/main/kotlin/org/jitsi/jicofo/xmpp/ConferenceIqHandler.kt
+++ b/src/main/kotlin/org/jitsi/jicofo/xmpp/ConferenceIqHandler.kt
@@ -17,6 +17,8 @@
  */
 package org.jitsi.jicofo.xmpp
 
+import org.jitsi.impl.protocol.xmpp.RegistrationListener
+import org.jitsi.impl.protocol.xmpp.XmppProvider
 import org.jitsi.jicofo.FocusManager
 import org.jitsi.jicofo.TaskPools
 import org.jitsi.jicofo.auth.AuthenticationAuthority
@@ -24,31 +26,38 @@ import org.jitsi.jicofo.auth.ErrorFactory
 import org.jitsi.jicofo.reservation.ReservationSystem
 import org.jitsi.utils.logging2.createLogger
 import org.jitsi.xmpp.extensions.jitsimeet.ConferenceIq
-import org.jivesoftware.smack.AbstractXMPPConnection
 import org.jivesoftware.smack.iqrequest.AbstractIqRequestHandler
 import org.jivesoftware.smack.iqrequest.IQRequestHandler
 import org.jivesoftware.smack.packet.IQ
 import org.jivesoftware.smack.packet.StanzaError
+import org.jxmpp.jid.DomainBareJid
+import org.jxmpp.jid.impl.JidCreate
 
 /**
  * Handles XMPP requests for a new conference ([ConferenceIq]).
  */
 class ConferenceIqHandler(
-    val connection: AbstractXMPPConnection,
+    val xmppProvider: XmppProvider,
     val focusManager: FocusManager,
     val focusAuthJid: String,
     val isFocusAnonymous: Boolean,
     val authAuthority: AuthenticationAuthority?,
     val reservationSystem: ReservationSystem?,
     val jigasiEnabled: Boolean
-) : AbstractIqRequestHandler(
+) : RegistrationListener, AbstractIqRequestHandler(
     ConferenceIq.ELEMENT_NAME,
     ConferenceIq.NAMESPACE,
     IQ.Type.set,
     IQRequestHandler.Mode.sync
 ) {
-
+    private val connection = xmppProvider.xmppConnection
+    private var breakoutAddress: DomainBareJid? = null
     private val logger = createLogger()
+
+    init {
+        xmppProvider.addRegistrationListener(this)
+        registrationChanged(xmppProvider.isRegistered)
+    }
 
     private fun handleConferenceIq(query: ConferenceIq): IQ {
         val response = ConferenceIq()
@@ -104,9 +113,12 @@ class ConferenceIqHandler(
     private fun processExtensions(query: ConferenceIq, response: ConferenceIq?, roomExists: Boolean): IQ? {
         val peerJid = query.from
         var identity: String? = null
+        val room = query.room
+        val isBreakoutRoom = breakoutAddress != null && room.domain == breakoutAddress
 
-        // Authentication
-        if (authAuthority != null) {
+        // Authentication. We do not perform authentication for breakout rooms, expecting the breakout room prosody
+        // module to handle it.
+        if (!isBreakoutRoom && authAuthority != null) {
             val authErrorOrResponse = authAuthority.processAuthentication(query, response)
 
             // Checks if authentication module wants to cancel further
@@ -116,17 +128,26 @@ class ConferenceIqHandler(
             }
             // Only authenticated users are allowed to create new rooms
             if (!roomExists) {
-                identity = authAuthority.getUserIdentity(peerJid)
-                if (identity == null) {
-                    // Error not authorized
-                    return ErrorFactory.createNotAuthorizedError(query, "not authorized user domain")
+                var breakoutRoomExists: Boolean = false
+                for (conference in focusManager.getConferences()) {
+                    val name = conference.getRoomName()
+                    if (name.domain == breakoutAddress && name.localpart.startsWith("${room.localpart}_")) {
+                        breakoutRoomExists = true
+                        break
+                    }
+                }
+                if (!breakoutRoomExists) {
+                    identity = authAuthority.getUserIdentity(peerJid)
+                    if (identity == null) {
+                        // Error not authorized
+                        return ErrorFactory.createNotAuthorizedError(query, "not authorized user domain")
+                    }
                 }
             }
         }
 
         // Check room reservation?
         if (!roomExists && reservationSystem != null) {
-            val room = query.room
             val result: ReservationSystem.Result = reservationSystem.createConference(identity, room)
             logger.info("Create room result: $result for $room")
             if (result.code != ReservationSystem.RESULT_OK) {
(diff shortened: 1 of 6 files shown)


src/main/kotlin/org/jitsi/jicofo/xmpp/ConferenceIqHandler.kt
```diff
@@ -104,9 +104,11 @@ class ConferenceIqHandler(
     private fun processExtensions(query: ConferenceIq, response: ConferenceIq?, roomExists: Boolean): IQ? {
         val peerJid = query.from
         var identity: String? = null
+        val room = query.room
+        val isBreakoutRoom = room.toString().endsWith("@breakout.${XmppConfig.client.xmppDomain}")
```
saghul, line 108: Maybe this should be configurable? Almost any component name is.

paweldomas: Yes, almost any component name is currently configurable, but I'm wondering how many users are changing them at all? Even if it would be configurable it would be great to have a default, so there's less config properties to setup.

saghul: The Docker setup does change them :-)

wernf: Agree that this should be done, including adjustments in lib-jitsi-meet, jitsi-meet and prosody. Preferably I would do this in a separate PR introducing a customisation feature for the breakout rooms domain. Does that sound reasonable?

wernf: Removed extra muc component definition for breakout rooms. They are using the main muc domain now. Breakout rooms are distinguished by a #breakout_uuid suffix.

src/main/kotlin/org/jitsi/jicofo/xmpp/ConferenceIqHandler.kt
```diff
@@ -104,9 +104,12 @@ class ConferenceIqHandler(
     private fun processExtensions(query: ConferenceIq, response: ConferenceIq?, roomExists: Boolean): IQ? {
         val peerJid = query.from
         var identity: String? = null
+        val room = query.room
+        val breakoutRoomRegex = Regex("#breakout_[-\\da-f]{36}@${XmppConfig.client.conferenceMucJid}$")
+        val isBreakoutRoom = room.toString().contains(breakoutRoomRegex)
```
bgrozev, line 109: Does this allow users to bypass authentication by using a special room name?

wernf: Thank you very much @bgrozev for your thorough review. You are right, special breakout room names will bypass the authentication check. This is to allow users to join a breakout room without having a moderator to join the room first. Authorisation is checked [here](https://github.com/jitsi/jitsi-meet/pull/9131/files#diff-844d74e9c13f71420b4cba715cd6167cfae74192cf534166c4e04b8b41001512R230) on Prosody by rejecting any breakout room which is not valid. So more concrete: Prosody will reject the creation of the room on the muc-room-pre-create event. Security is achieved by the random uuid in the breakout room names.

src/main/kotlin/org/jitsi/jicofo/xmpp/ConferenceIqHandler.kt
```diff
@@ -116,17 +119,27 @@ class ConferenceIqHandler(
...
-                identity = authAuthority.getUserIdentity(peerJid)
-                if (identity == null) {
-                    // Error not authorized
-                    return ErrorFactory.createNotAuthorizedError(query, "not authorized user domain")
+                var breakoutRoomExists: Boolean = false
+                val breakoutRoomsNamePrefix = "${room.toString().substringBefore('@')}_breakout-"
+                for (conference in focusManager.getConferences()) {
+                    val name = conference.getRoomName()
+                    if (name.startsWith(breakoutRoomsNamePrefix)) {
+                        breakoutRoomExists = true;
```
saghul, line 127: Maybe calling this `mainRoomExists` is more accurate.

saghul (approved):
LGTM with a nit.

src/main/kotlin/org/jitsi/jicofo/xmpp/ConferenceIqHandler.kt
bgrozev: Sorry for the slow response, I missed that! Prosody only rejects the creation of a breakout room when `mod_breakout_rooms` is loaded, correct? Without it anyone can create a room that matches `breakoutRoomRegexp` and bypass jicofo's authentication?

bgrozev (changes requested):
Please don't merge until the question about authentication has been discussed.

src/main/kotlin/org/jitsi/jicofo/xmpp/ConferenceIqHandler.kt
wernf: Good catch, @bgrozev! 

I see two options to get this fixed:

1. make the main muc in prosody block rooms with the breakout room name scheme by default and have the breakout rooms module allow these with a higher priority.
1. re-introduce  a new breakout room muc component

I tend to prefer the first option as it avoids the overhead of introducing a breakout domain handling to jitsi-meet, lib-jitsi-meet and jicofo. But I admit that the second solution would allow a clearer distinction between the room types. So on the long run we might need to go down that road. I would just like to try to avoid setting the hurdle to hight for the initial release of this feature.

Or do you have another idea? What do you think?

saghul: Can you elaborate on option 2? How would a new room muc component helps us fix the problem?

wernf: It would mostly come down to roll back these changes:

- https://github.com/jitsi/jitsi-meet/pull/9131/commits/02888917e463eab054f9b9d3257189eea284fd18
- https://github.com/jitsi/jitsi-meet/pull/9131/commits/e40f76287592af013854e14426e6a26a4f312035
- https://github.com/jitsi/lib-jitsi-meet/pull/1654/commits/26ef339f6c4a3e742f55029e375dc83e1ee467ee
- https://github.com/jitsi/docker-jitsi-meet/pull/1088/commits/faf72082ae986d51d719d12f84cbc5f0e3e0a1b7
- jicofo: https://github.com/jitsi/jicofo/pull/759/commits/4a990db4b4862c1c8432736f59f0a6dd20a3318b

So actually it was like that before. However comments about the introduction of a new muc and the handling of the domain name made me consider to renounce to that approach.

This would fix the problem as only the breakout module would handle breakout mucs. So there would be no chance for somebody to start a breakout room without the breakout module being active (and checking whether it is a valid room).

saghul: > This would fix the problem as only the breakout module would handle breakout mucs. So there would be no chance for somebody to start a breakout room without the breakout module being active (and checking whether it is a valid room).

That sounds quite good actually. Can you elaborate on the drawbacks? Sorry for the back and forth here :-/

wernf: Well, drawbacks come down to what is in those patches. That is:

- We will need to teach jitsi-meet, lib-jitsi-meet, prosody and jicofo the new muc domain (in contrast to only let them know about the "weird" breakout room names)
- The bit more complicated configuration of Prosody
- @saghul asking to make the domain name configurable :D

saghul: That saghul guy.... :-P

Let's try to brainstorm about this next week and figure out a way forward.

src/main/kotlin/org/jitsi/jicofo/xmpp/ConferenceIqHandler.kt
```diff
@@ -116,17 +129,27 @@ class ConferenceIqHandler(
...
             if (!roomExists) {
-                identity = authAuthority.getUserIdentity(peerJid)
-                if (identity == null) {
-                    // Error not authorized
-                    return ErrorFactory.createNotAuthorizedError(query, "not authorized user domain")
+                var breakoutRoomExists: Boolean = false
+                val breakoutRoomsNamePrefix = "${room.toString().substringBefore('@')}_"
+                for (conference in focusManager.getConferences()) {
+                    val name = conference.getRoomName()
+                    if (name.endsWith(breakoutAddress.toString()) && name.startsWith(breakoutRoomsNamePrefix)) {
```
bgrozev, line 136: nit: Since the names are `Jid`s you can use something like
```
if (name.localpart.startsWith("${room.localpart}_") && name.domain == breakoutAddress)
```

bgrozev (dismissed):

src/main/kotlin/org/jitsi/jicofo/xmpp/ConferenceIqHandler.kt
wernf: Thanks @bgrozev. This is an excellent comment. I have adjusted the code accordingly.

bgrozev (approved):

saghul:
@bgrozev I rebased this with latest master, can you PTAL?

saghul:
Updated with automagic detection of breakout rooms and skipping the auto-owner thing. @bgrozev @damencho can you PTAL? I tested it and it works as expected.

bgrozev (approved):

bgrozev:
Note that there's a delay between jicofo creating a `JitsiMeetConference` object and it receiving the MUC form. In this period the conference will not be recognized as breakout and anonymous users will not be allowed in the associated main room. It shouldn't be a problm in practice.

saghul:
Good observation! I don't think that's a problem for now.

saghul:
Continues on https://github.com/jitsi/jicofo/pull/818

PR #768, opened 2021-08-02
Refactor and optimize SSRC validation

bgrozev:


src/main/kotlin/org/jitsi/jicofo/conference/source/Source.kt
```diff
@@ -0,0 +1,73 @@
...
+    /** Serializes this [SsrcGroup] to XML */
+    fun toPacketExtension(
+        /** An optional JID for the owner of this source to encode in the XML extension. */
+        owner: Jid? = null
+    ) = SourcePacketExtension().apply {
+        ssrc = this@Source.ssrc
+        if (owner != null) {
+            SSRCInfoPacketExtension().also {
+                it.owner = owner
+                addChildExtension(it)
```
JonathanLennox, line 73: I find calling `addChildExtension(it)` in the `also` a bit confusing - maybe it'd be clearer to call `addChildExtension` on the `SSRCInfoPacketExtension` constructor?

src/main/kotlin/org/jitsi/jicofo/conference/source/SsrcGroupSemantics.kt
```diff
@@ -0,0 +1,35 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jitsi.jicofo.conference.source
+
+import kotlin.jvm.Throws
+
+enum class SsrcGroupSemantics {
```
JonathanLennox, line 20: This could use a kdoc; probably specifically noting that we're only including those semantics we support.

src/main/kotlin/org/jitsi/jicofo/conference/source/ConferenceSourceMap.kt
```diff
@@ -43,15 +44,21 @@ data class ConferenceSourceMap(
         }
     )
 
+    /**
+     * An unmodifiable view of this [ConferenceSourceMap].
```
JonathanLennox, line 64: The view is unmodifiable, but it's not a COW snapshot of the map, right?

src/main/java/org/jitsi/impl/protocol/xmpp/colibri/ColibriConferenceImpl.java
```diff
@@ -52,6 +53,36 @@
 
     private final static Logger logger = new LoggerImpl(ColibriConferenceImpl.class.getName());
 
+    private static Pair<Map<String, List<SourcePacketExtension>>, Map<String, List<SourceGroupPacketExtension>>>
```
JonathanLennox, line 56: Can this type get named?  It's pretty verbose.  (Assuming it's going to stick around for the final version of the PR.)

bgrozev: Yes, for sure.

src/main/kotlin/org/jitsi/jicofo/conference/source/ConferenceSourceMap.kt
bgrozev: Correct. I'll be looking more closely at this, now that I know all the ways it's used (I still need to make changes for thread safety).

src/main/kotlin/org/jitsi/jicofo/conference/source/SsrcGroupSemantics.kt
bgrozev: Fixed.

src/main/kotlin/org/jitsi/jicofo/conference/source/Source.kt
bgrozev: Agreed and fixed.

src/main/kotlin/org/jitsi/jicofo/conference/source/EndpointSourceSet.kt
```diff
@@ -0,0 +1,143 @@
...
+import org.jitsi.utils.MediaType
+import org.jitsi.xmpp.extensions.colibri.SourcePacketExtension
+import org.jitsi.xmpp.extensions.jingle.ContentPacketExtension
+import org.jitsi.xmpp.extensions.jingle.RtpDescriptionPacketExtension
+import org.jitsi.xmpp.extensions.jingle.SourceGroupPacketExtension
+import org.jxmpp.jid.Jid
+import java.lang.IllegalArgumentException
+import kotlin.jvm.Throws
+
+/** A set [Source]s and [SsrcGroup]s (optionally) associated with a specific endpoint (identified by [owner]). */
```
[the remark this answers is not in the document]
bgrozev: An old version of the code had an `owner` field. Fixed.

src/main/kotlin/org/jitsi/jicofo/conference/source/ValidatingConferenceSourceMap.kt
```diff
@@ -18,67 +18,302 @@
...
+ */
+class ValidatingConferenceSourceMap(
+    private val maxSsrcsPerUser: Int = ConferenceConfig.config.maxSsrcsPerUser
+) : ConferenceSourceMap() {
     val logger = createLogger()
 
+    /**
+     * Maps an SSRC to the JID of the endpoint that owns it. Used to detect cross-endpoint conflicts efficiently.
+     *
+     * Note that since the owner is nullable [Map.containsKey] must be used when checking for.
```
JonathanLennox, line 44: "when checking for" is missing an object, presumably "map membership" or the like.  Likewise below.

src/main/kotlin/org/jitsi/jicofo/conference/source/ValidatingConferenceSourceMap.kt
```diff
@@ -18,67 +18,302 @@
...
-            )
-            add(addedConferenceSourceMap)
-            return addedConferenceSourceMap
-        } catch (e: InvalidSSRCsException) {
-            throw ValidationFailedException(e.message)
+        val acceptedGroups = mutableSetOf<SsrcGroup>()
+        sourcesToAdd.ssrcGroups.forEach {
+            // We just ignore empty groups and groups which have been signaled multiple times.
+            when {
+                it.ssrcs.isEmpty() -> logger.info("Empty group signaled, ignoring.")
```
JonathanLennox, line 102: How noisy will these logs be? I'm wondering if they need to be `debug`.

src/main/kotlin/org/jitsi/jicofo/conference/source/ValidatingConferenceSourceMap.kt
```diff
@@ -18,67 +18,302 @@
...
+    /**
+     * Attempts to remove [sourcesToRemove] as sources owned by [owner]. The attempt is successful if all referenced
+     * sources and groups are indeed owned by [owner], and the resulting set of sources for [owner] is valid (see
+     * [validateEndpointSourceSet]).
+     *
+     * If the attempt is successful, the sources are removed from [owner]'s sources and the set of removed sources is
+     * returned. The set of removed sources might differ from [sourcesToRemove] because:
+     * 1. We only match a source's SSRC, so a source added with parameters (MSID, etc) can be removed by simply
+     * referencing the SSRC.
+     * 2. We automatically remove any groups that contain any of the removed sources.
```
JonathanLennox, line 128: Should this be "that contain all of the removed sources"?  What do we do if one source in a group is removed?  Or will that cause a validation failure?

src/main/kotlin/org/jitsi/jicofo/conference/source/ValidatingConferenceSourceMap.kt
```diff
@@ -18,67 +18,302 @@
...
+            //
+            // This is technically O(s*g^2) where s is the number of sources and g is the number of groups. But in
+            // practice both s and g are very small (at most 20 and 4 respectively).
+            val grouped: Map<SsrcGroup, List<Source>> =
+                endpointSourceSet.sources.groupBy { groupBySimulcastGroup(it.ssrc, endpointSourceSet.ssrcGroups) }
+
+            // Here we check for MSID conflicts, i.e. we make sure that each group has a unique MSID.
+            val msidsSeen = mutableSetOf<String?>()
+            grouped.values.map { it[0] }.forEach { source ->
+                if (msidsSeen.contains(source.msid)) throw MsidConflictException(source.msid ?: "NONE")
```
JonathanLennox: Since a source's msid is nullable, does `msidsSeen.contains(null)` make sense?

src/main/kotlin/org/jitsi/jicofo/conference/source/ValidatingConferenceSourceMap.kt
```diff
@@ -18,67 +18,302 @@
...
+            // FID(111, 222)
+            // And non-grouped source 333 this will associate the SIM group with sources 1..6, the lonely FID group
+            // with sources 111 and 222, and a dummy single-source group with source 333.
+            //
+            // This is technically O(s*g^2) where s is the number of sources and g is the number of groups. But in
+            // practice both s and g are very small (at most 20 and 4 respectively).
+            val grouped: Map<SsrcGroup, List<Source>> =
+                endpointSourceSet.sources.groupBy { groupBySimulcastGroup(it.ssrc, endpointSourceSet.ssrcGroups) }
+
+            // Here we check for MSID conflicts, i.e. we make sure that each group has a unique MSID.
```
JonathanLennox: Will this make it harder in the future to support endpoints with multiple video sources?  I'm thinking that the test should be that if the msid is the same, then the epId should be the same.  But maybe this should just get a comment for now, since we don't support that case yet.

src/main/kotlin/org/jitsi/jicofo/conference/source/ValidatingConferenceSourceMap.kt
bgrozev: They are currently warnings :)

src/main/kotlin/org/jitsi/jicofo/conference/source/ValidatingConferenceSourceMap.kt
bgrozev: We require that all groups' SSRCs be described as sources. If one source in a group is removed, it may result in a validation failure or not, depending on what remains. 

For example, if we remove source 1 from this we end up with a valid set (the group is removed, only source 2 remains):
```
sources = 1, 2
groups = sim(1,2)
```

But if we remove source 1 from this, we end up with an invalid set (sources 2 and 3 have the same MSID but are not grouped in any way):
```
sources = 1, 2, 3
groups = sim(1, 2, 3)
```

I think this is new. Previously we didn't automatically remove groups. If you think it may be confusing or lead to problems I can remove it.

src/main/kotlin/org/jitsi/jicofo/conference/source/ValidatingConferenceSourceMap.kt
bgrozev: Yes, because the set is of nullable strings. The following works as one would expect:
```
val s = mutableSetOf<String?>()

s.contains(null) //false
s.add(null)
s.contains(null) //true
```

src/main/kotlin/org/jitsi/jicofo/conference/source/ValidatingConferenceSourceMap.kt
bgrozev: I don't know. Wouldn't the client use a new MSID for its other stream? Honestly, I don't understand why jicofo does any validation apart from detecting conflicts between endpoints, but I'm hesitant to remove it (this check is effectively there in the existing code, but it's not clear if the intention is to check for colisions within the same endpoint or not, since it works on the map of all sources in the conference).

src/main/kotlin/org/jitsi/jicofo/conference/source/ValidatingConferenceSourceMap.kt
JonathanLennox: Yeah, I'm wondering what would happen if an endpoint decided to go from a three-source SIM group to a two-source SIM group.  Do we even have a way to signal that now?

bgrozev: Hm, no. It would have to do a `remove` followed by an `add`.

src/main/kotlin/org/jitsi/jicofo/conference/source/ConferenceSourceMap.kt
```diff
@@ -21,27 +21,29 @@ import org.jitsi.xmpp.extensions.jingle.ContentPacketExtension
...
 
-    constructor(vararg entries: Pair<Jid?, EndpointSourceSet>) : this(
-        mutableMapOf<Jid?, EndpointSourceSet>().apply {
-            entries.forEach { (k, v) ->
-                this[k] = v
-            }
-        }
-    )
-
+    /** C */
```
JonathanLennox, line 44: A bit sparse as a doc...

src/main/kotlin/org/jitsi/jicofo/conference/source/ConferenceSourceMap.kt
```diff
@@ -120,8 +128,6 @@ open class ConferenceSourceMap(
         return extensions
     }
 
-    fun copy(): ConferenceSourceMap = ConferenceSourceMap(endpointSourceSets.toMutableMap())
-
     fun removeInjected() = this.apply {
```
JonathanLennox, line 131: Does this need to be synchronized?

src/main/kotlin/org/jitsi/jicofo/conference/source/ValidatingConferenceSourceMap.kt
```diff
@@ -177,16 +179,20 @@ class ValidatingConferenceSourceMap(
...
-    override fun add(owner: Jid?, endpointSourceSet: EndpointSourceSet) = super.add(owner, endpointSourceSet).also {
-        sourceSetAdded(owner, endpointSourceSet)
+    override fun add(owner: Jid?, endpointSourceSet: EndpointSourceSet) = synchronized(syncRoot) {
+        super.add(owner, endpointSourceSet).also {
+            sourceSetAdded(owner, endpointSourceSet)
+        }
     }
 
-    private fun sourceSetAdded(owner: Jid?, endpointSourceSet: EndpointSourceSet) {
+    private fun sourceSetAdded(owner: Jid?, endpointSourceSet: EndpointSourceSet) = synchronized(syncRoot) {
```
JonathanLennox, line 199: Is `sourceSetAdded` ever called in a non-synchronized context?  If not you shouldn't need to synchronize it again.  (And if it is, that seems like it could be a race.). Likewise for `sourceSetRemoved`.

src/main/kotlin/org/jitsi/jicofo/conference/source/ValidatingConferenceSourceMap.kt
JonathanLennox: I _think_ clients use the same MSID for all sources in a PeerConnection?  But actually I'm not sure.  Conceptually I think it has to do with whether they're supposed to be synchronized, but I'm not 100% positive what browsers actually do.

src/main/kotlin/org/jitsi/jicofo/conference/source/ValidatingConferenceSourceMap.kt
JonathanLennox: That should be fine then, I guess.

src/main/kotlin/org/jitsi/jicofo/conference/source/ConferenceSourceMap.kt
bgrozev: Yes, thank you.

src/main/kotlin/org/jitsi/jicofo/conference/source/ValidatingConferenceSourceMap.kt
bgrozev: My preference is to keep the redundant lock aquisition to make it clear that the function requires the lock. I think the performance impact is negligible (we had a similar discussion a long time ago and I ran some basic tests, but I don't have the results anymore). But I'm open to changing it if you insist.

src/main/kotlin/org/jitsi/jicofo/conference/source/ValidatingConferenceSourceMap.kt
bgrozev: I think I agree. I don't see a reason why jicofo would enforce this. Do you mind leaving it for a follow-up PR?

src/main/kotlin/org/jitsi/jicofo/conference/source/ValidatingConferenceSourceMap.kt
JonathanLennox: Ok.

JonathanLennox (approved):

src/main/kotlin/org/jitsi/jicofo/conference/source/ValidatingConferenceSourceMap.kt
JonathanLennox: No, that's fine.

PR #771, opened 2021-08-03
Improvements to jicofo's PacketQueue usage.

JonathanLennox:


src/main/kotlin/org/jitsi/jicofo/util/QueueExecutor.kt
```diff
@@ -35,7 +35,7 @@ class QueueExecutor(
 ) : Executor, PacketQueue<Runnable>(
     Integer.MAX_VALUE,
     true,
-    id,
+    "QueueExecutor",
```
bgrozev, line 38: I added the ID field in case `JitsiMeetConferenceImpl` is not the only user of `QueueExecutor`. If this is unnecessary, we should also remove the `id` param from the constructor.

JonathanLennox: Ah, it's got a static name, rather than a per-conference name.  In that case the passed-in `id` is fine.

bgrozev (approved):

PR #777, opened 2021-08-05
feat: Remove limitations for MSID signaled by an endpoint.

bgrozev:
SSRCValidator used to put all SSRCs and groups in a conference in one place and run checks on the whole sett. In #768 I changed that to check for conflicts between endpoints and then check for consistency whithin the sources singaled from a single endpoint. To preserve the existing behavior I added the same check that, grouped into "simulcast groups", MSIDs need to be unique. This means that an endpoint is not allowed to use the same MSID for more than one of its sources. I don't think this was intended by SSRCValidator (@paweldomas perhaps you remember?), and the limitation seems unnecessary, so this PR removes it.

In effect this PR allows an endpoint to use whatever MSIDs it likes, as long as they don't conflict with another endpoint, and the groups have a consistent MSID value.

### DIFF
--- a/src/main/kotlin/org/jitsi/jicofo/conference/source/ValidatingConferenceSourceMap.kt
+++ b/src/main/kotlin/org/jitsi/jicofo/conference/source/ValidatingConferenceSourceMap.kt
@@ -18,7 +18,6 @@
 package org.jitsi.jicofo.conference.source
 
 import org.jitsi.jicofo.ConferenceConfig
-import org.jitsi.utils.MediaType
 import org.jitsi.utils.logging2.createLogger
 import org.jxmpp.jid.Jid
 import java.lang.IllegalStateException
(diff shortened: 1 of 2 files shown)


paweldomas:
Does it remove the limitation of 1 MSID signaled for an endpoint per media type? Is it fixing a bug (I would be afraid it would mask a bug) or for a feature?

JonathanLennox (approved):

bgrozev:
> Does it remove the limitation of 1 MSID signaled for an endpoint per media type? 

I don't think we ever had that limitation. I'll try to verify.

>Is it fixing a bug (I would be afraid it would mask a bug) or for a feature?

No, just removing code which seems artificial and redundant.

Note: See the updated description above.

paweldomas:
I think the limitation was to not allow have the same MSID for two ungrouped sources (SSRC numbers at the time). Which would effectively not allow to have two tracks of one media type I think? I'm just trying to understand the implications of this change

paweldomas:
>> Does it remove the limitation of 1 MSID signaled for an endpoint per media type?

> I don't think we ever had that limitation. I'll try to verify.

I think there is such limitation on the client that you can not add more than one 1 video track for example.

paweldomas:
> SSRCValidator used to put all SSRCs and groups in a conference in one place and run checks on the whole sett.

The motivation for checking it as a whole set was to verify that set remote description error would not happen on the client. Basically to prevent from breaking conference for everyone when one of the clients adds sources that result in invalid remote SDP. I'm not saying it's not possible to do in the new approach just reminding what was the goal of those validations.

bgrozev:
> The motivation for checking it as a whole set was to verify that set remote description error would not happen on the client. Basically to prevent from breaking conference for everyone when one of the clients adds sources that result in invalid remote SDP. I'm not saying it's not possible to do in the new approach just reminding what was the goal of those validations.

Okay, that's good to know. What I'm trying to understand is: the code prohibits an endpoint to signal, for example, two ungrouped SSRCs with the same MSID. Is this intentional, or a side-effect of the way the check was implemented?

bgrozev:
> I think the limitation was to not allow have the same MSID for two ungrouped sources (SSRC numbers at the time). Which would effectively not allow to have two tracks of one media type I think? I'm just trying to understand the implications of this change

Yes, I think that's correct. With this PR, an endpoint is allowed to signal multiple SSRCs with the same media type and the same MSID and no groups.

paweldomas:
> Okay, that's good to know. What I'm trying to understand is: the code prohibits an endpoint to signal, for example, two ungrouped SSRCs with the same MSID. Is this intentional, or a side-effect of the way the check was implemented?

I think it was intentional, because lib-jitsi-meet doesn't allow that. What's the use case for two SSRCs with the same MSID?

bgrozev:
> I think it was intentional, because lib-jitsi-meet doesn't allow that. What's the use case for two SSRCs with the same MSID?

None that I know of.

JonathanLennox:
> I think it was intentional, because lib-jitsi-meet doesn't allow that. What's the use case for two SSRCs with the same MSID?

If (using WebRTC terminology) two MediaStreamTracks are part of the same MediaStream, they get the same MSID.

Which is to say, if the tracks (sources) are supposed to be synchronized, they get the same MSID.  (This is why the audio and video sources have the same MSID).

We're not likely to support this any time soon, but I wanted to make sure we weren't painting ourselves into a corner if we did want to support it in the future.

PR #778, opened 2021-08-06
Add more debug logs for changes to sources.

bgrozev:


src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
```diff
@@ -1622,7 +1625,12 @@ private XMPPError onSessionAcceptInternal(
         participant.setRTPDescription(contents);
         participant.addTransportFromJingle(contents);
 
+        String participantId = participant.getEndpointId();
         EndpointSourceSet sourcesAdvertised = EndpointSourceSet.fromJingle(contents);
+        if (logger.isDebugEnabled())
```
JonathanLennox, line 1630: Why do this with `isDebugEnabled()` rather than `debug(() ->`?

bgrozev: Because using a lambda requires `sourcesAdvertised` to be final, but it's modified below.

JonathanLennox (approved):

PR #779, opened 2021-08-06
fix: Fix the way source-add/source-remove are queued before the session is established.

bgrozev:


JonathanLennox:
I don't see the definition of `SourcesToAddOrRemove here`, did you forget a `git add`?

src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
```diff
@@ -1682,18 +1683,20 @@ private XMPPError onSessionAcceptInternal(
...
+        for (SourcesToAddOrRemove sourcesToAddOrRemove : participant.clearQueuedRemoteSourceChanges())
         {
-            jingle.sendAddSourceIQ(remoteSourcesToAdd, jingleSession);
-        }
-        if (!remoteSourcesToRemove.isEmpty())
-        {
-            jingle.sendRemoveSourceIQ(remoteSourcesToRemove, jingleSession);
+            AddOrRemove action = sourcesToAddOrRemove.getAction();
+            ConferenceSourceMap sources = sourcesToAddOrRemove.getSources();
+            logger.info("Sending a queued source-" + action + " to " + participantId + ", sources:" + sources);
```
JonathanLennox, line 1690: Will this log as a `source-Add` and `source-Remove` (with the capital letters) because of how the enum is spelled?  Not the end of the world but a little weird.

bgrozev: Yes, I was lazy :)

JonathanLennox (approved):

PR #782, opened 2021-08-09
fix: Do not remove octo sources twice.

bgrozev:


src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java
```diff
@@ -2223,14 +2223,14 @@ private void reInviteParticipants(List<Participant> participants)
         }
     }
 
-    private BridgeSession terminateParticipantBridgeSession(@NotNull Participant participant)
+    private BridgeSession terminateParticipantBridgeSession(@NotNull Participant participant, boolean removeSources)
```
JonathanLennox, line 2226: Do we want a Javadoc explaining this?  (Under our rules this should have a Javadoc regardless...)

JonathanLennox (approved):

PR #783, opened 2021-08-09
Strip simulcast streams when signaling receivers.

bgrozev:


JonathanLennox (dismissed):

JonathanLennox:
The one comment I'd make is that documentation of **why** we can strip simulcast (because jvb is anyway projecting all the sources down to a single SSRC anyway) is probably a good idea.  Also to point out that its logic has to be in sync with jvb's.

src/main/java/org/jitsi/protocol/xmpp/AbstractOperationSetJingle.java
```diff
@@ -277,13 +277,6 @@ public void sendAddSourceIQ(ConferenceSourceMap sources, JingleSession session)
     @Override
     public void sendRemoveSourceIQ(ConferenceSourceMap sourcesToRemove, JingleSession session)
     {
-        sourcesToRemove = sourcesToRemove.copy().removeInjected();
```
JonathanLennox, line 287: Don't you still need to handle the `isEmpty()` case here?

JonathanLennox:
Given how often you have
```
        if (ConferenceConfig.config.stripSimulcast())
        {
            sources.stripSimulcastAndInjected();
        }
        else
        {
            sources.removeInjected();
        }
```
would it make sense to make that a method?

JonathanLennox (approved):

PR #785, opened 2021-08-19
Split SourcesTest in multiple files.

bgrozev:


JonathanLennox (approved):
I'm assuming these were copy&paste'd so I don't have to review them in detail - the moves look fine.

PR #802, opened 2021-09-07
Bump Smack version to 4.4.3; make API changes accordingly.

JonathanLennox:


JonathanLennox:
Depends on https://github.com/jitsi/jitsi-xmpp-extensions/pull/47, https://github.com/jitsi/jicoco/pull/153, and https://github.com/jitsi/jitsi-videobridge/pull/1719.

bgrozev:
LGTM

bgrozev (approved):

bgrozev:
> It generates both per-stanza logs and per read/write logs. The per-stanza logs are more useful, so this is ignoring the read/write logs.

Do the per-stanza logs pass through the XML parser? I wonder if they are representative of what's on the wire. If they pass through the XML parser they might:
a) be missing elements which were not parsed by jitsi-xmpp-extensions
b) include extra namespaces (I've seen that happen, I don't know why)

bgrozev (approved):

PR #812, opened 2021-09-28
test: Updates the test to cover jicofo/pull/809.

damencho:


JonathanLennox (dismissed):

src/test/java/org/jitsi/jicofo/xmpp/ConferenceIqProviderTest.java
```diff
@@ -78,6 +79,30 @@ public void testParseConferenceIq()
...
+                        " room='somename@email.com@example.com' ready='true'>" +
+                    "</conference>" +
+                "</iq>";
+
+        try
+        {
+            ConferenceIqProvider provider = new ConferenceIqProvider();
+            IQUtils.parse(iqXml, provider);
+        }
+        catch(Exception e)
```
paweldomas, line 99: how about putting a specific exception class in the catch?

src/test/java/org/jitsi/jicofo/xmpp/ConferenceIqProviderTest.java
```diff
@@ -78,6 +79,30 @@ public void testParseConferenceIq()
...
+        {
+            // we expect XmppStringprepException
+            assertEquals(XmppStringprepException.class, e.getClass());
```
bgrozev, line 102: This does not assert that an exception is thrown

bgrozev (approved):

src/test/java/org/jitsi/jicofo/xmpp/ConferenceIqProviderTest.java
```diff
@@ -91,16 +91,22 @@ public void testParseConferenceIqWithWrongRoom()
...
         {
             ConferenceIqProvider provider = new ConferenceIqProvider();
             IQUtils.parse(iqXml, provider);
         }
         catch(Exception e)
         {
-            // we expect XmppStringprepException
-            assertEquals(XmppStringprepException.class, e.getClass());
-        }   
+            resultException = e;
```
paweldomas, line 103: I think it's cleaner to do the catch specifically for `XmppStringprepException` and put a fail statement at line 100 after parse which is the case when exception is not thrown.

paweldomas: and the fail message says "XmppStringprepException was not thrown as expected"

damencho: But isn't it clearer this way, in case of the failure you will have:

[ERROR] Failures:
[ERROR]   ConferenceIqProviderTest.testParseConferenceIqWithWrongRoom:102 expected:<class org.jxmpp.stringprep.XmppStringprepException> but was:<class java.lang.ArrayIndexOutOfBoundsException>

While in the other case the test throwing an error ... and you need to find out why, while here it is written: expected is this, not that.

paweldomas: it's just not a standard pattern for testing thrown exceptions

bgrozev: 🙄
```
shouldThrow<XmppStringprepException> {
    IQUtils.parse(iqXml, provider)
}
```

PR #815, opened 2021-10-01
fix: Avoid a race in Smack 4.4.3 which causes

bgrozev:


JonathanLennox (approved):
Approved, but try to update the description on commit not to be incomplete?

PR #816, opened 2021-10-01
Hack to work around a race in Smack.

JonathanLennox:
(See commit 3f570999 for an explanation of the race.)

### DIFF
--- a/src/main/java/org/jitsi/impl/protocol/xmpp/ChatRoomImpl.java
+++ b/src/main/java/org/jitsi/impl/protocol/xmpp/ChatRoomImpl.java
@@ -1094,11 +1094,14 @@ public class ChatRoomImpl
         Field occupantsMapField = null;
         try
         {
-            occupantsMapField = muc.getClass().getField("occupantsMap");
+            occupantsMapField = muc.getClass().getDeclaredField("occupantsMap");
             occupantsMapField.setAccessible(true);
 
             Map<EntityFullJid, Presence> occupantsMap = (Map<EntityFullJid, Presence>)occupantsMapField.get(muc);
-            occupantsMap.clear();
+            if (!occupantsMap.isEmpty()) {
+                logger.warn("MultiUserChat occupantsMap is not empty, clearing.");
+                occupantsMap.clear();
+            }
         }
         catch (NoSuchFieldException | IllegalAccessException e)
         {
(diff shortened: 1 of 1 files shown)


bgrozev (dismissed):

src/main/java/org/jitsi/impl/protocol/xmpp/ChatRoomImpl.java
```diff
@@ -1076,4 +1081,28 @@ public void roomDestroyed(MultiUserChat alternateMUC, String reason)
...
+        assert(!muc.isJoined());
+
+        Field occupantsMapField = null;
+        try
+        {
+            occupantsMapField = muc.getClass().getField("occupantsMap");
+            occupantsMapField.setAccessible(true);
+
+            Map<EntityFullJid, Presence> occupantsMap = (Map<EntityFullJid, Presence>)occupantsMapField.get(muc);
+            occupantsMap.clear();
```
damencho, line 1101: What about printing a log if there was something in the map?

src/main/java/org/jitsi/impl/protocol/xmpp/ChatRoomImpl.java
```diff
@@ -1076,4 +1081,28 @@ public void roomDestroyed(MultiUserChat alternateMUC, String reason)
...
+     * chat instance.  This function uses reflection to hack the internal state to fix the problem.
+     */
+    private void clearMucOccupantsMap(MultiUserChat muc)
+    {
+        assert(!muc.isJoined());
+
+        Field occupantsMapField = null;
+        try
+        {
+            occupantsMapField = muc.getClass().getField("occupantsMap");
```
damencho, line 1097: This should be getDeclaredField

damencho:
I did some automation and reverted the previous change and not I cannot hit it ... it is so strange ... I even add in prosody code to send a self presence update before the unavailable message and I see it in jicofo xmpp logs but still nothing, I ran 50 attempts ... will leave it now to do more attempts during the day. 
I'm waiting for a print I added on `if (muc.getOccupants().size() > 0)` or at least the test will fail to join the muc and I will see that ...

damencho:
Ok, here is an update. Out of 200 runs managed to repro it once and the fix with clearing the map worked, no locking. 
The runs were without prosody modifications and just with reverted the latest change in master about the setting of bridge 0.

damencho:
And leaving just the fix from Friday I was able to repro in around 60 attempts. So I think we need this one :)

bgrozev (approved):

damencho (approved):

PR #821, opened 2021-10-27
fix: Do not consider bridges lost if they have no conferences.

bgrozev:


bgrozev:
Not necessary since jitsi/jitsi-videobridge#1756

PR #826, opened 2021-11-04
fix: Fix error message when a jibri session already exists.

bgrozev:


bgrozev:
This changes the error condition, so it will be [rendered differently in jitsi-meet](https://github.com/jitsi/jitsi-meet/blob/master/react/features/recording/middleware.js#L254). I think that's fine, as it's either a rare race condition or a bug in the client. Any thoughts @saghul  (see JID-8726 for more context)

yanas (approved):

PR #827, opened 2021-11-09
Limit video senders

jqdrqgnq:
Keep track of the number of video senders in a conference, that is, the number of participants who have unmuted their video source. When the limit is reached, return an error if another participant attempts to unmute their video.

This is an attempt to avoid issues that may arise from the storm of signaling that may occur when a large number of participants unmute at the same time.

### DIFF
--- a/src/main/java/org/jitsi/jicofo/conference/JitsiMeetConferenceImpl.java
+++ b/src/main/java/org/jitsi/jicofo/conference/JitsiMeetConferenceImpl.java
@@ -1504,6 +1504,16 @@ public class JitsiMeetConferenceImpl
             return StanzaError.from(StanzaError.Condition.bad_request, e.getMessage()).build();
         }
 
+        // $ optimize. add methods to EndpointSourceSet that don't need to copy ssrcs.
+        boolean hasVideo = !sourcesAdvertised.getVideoSsrcs().isEmpty();
+
+        if (hasVideo && chatRoom.getVideoSendersCount() >= ConferenceConfig.config.getMaxVideoSenders())
+        {
+            String errorMsg = "Source add rejected. Maximum number of video senders reached.";
+            logger.warn(() -> participantId + ": " + errorMsg);
+            return StanzaError.from(StanzaError.Condition.resource_constraint, errorMsg).build();
+        }
+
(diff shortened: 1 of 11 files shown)


src/main/kotlin/org/jitsi/jicofo/conference/source/EndpointSourceSet.kt
```diff
@@ -41,6 +41,16 @@ data class EndpointSourceSet(
      */
     fun isEmpty() = sources.isEmpty() && ssrcGroups.isEmpty()
 
+    /**
+     * Whether there are any audio sources in this set.
+     */
+    fun hasAudio() = sources.any { it.mediaType == AUDIO }
```
bgrozev, line 47: Nit since you're optimizing anyway :)
`EndpointSourceSet` is immutable so this could be a `val` calculated once. And if you use the `lazy` delegate it will only be initialized on demand: see e.g. the definition of `audioSsrcs` below.

src/main/java/org/jitsi/jicofo/conference/JitsiMeetConferenceImpl.java
```diff
@@ -2425,6 +2443,12 @@ public void bridgeAdded(Bridge bridge)
 
     private class ChatRoomListenerImpl implements ChatRoomListener
     {
+        private final JitsiMeetConferenceImpl conference;
+        private ChatRoomListenerImpl(JitsiMeetConferenceImpl conference)
```
bgrozev, line 2447: Since `ChatRoomListener` is an inner class of `JitsiMeetConferenceImpl` it has an implicit reference to it. You can just call `numVideoSendersChanged()` (see e.g. how `roomDestroyed` calls `JitsiMeetConferenceImpl.stop()`).

jqdrqgnq: thanks.

src/main/kotlin/org/jitsi/jicofo/conference/source/EndpointSourceSet.kt
jqdrqgnq: cool, thanks.

PR #830, opened 2021-11-11
feat: Add an enable-rtcstats field to ColibriConferenceIQ.

gpolitis:


pom.xml
```diff
@@ -101,7 +101,7 @@
     <dependency>
       <groupId>${project.groupId}</groupId>
       <artifactId>jitsi-xmpp-extensions</artifactId>
-      <version>1.0-30-g48113ca</version>
+      <version>1.0-SNAPSHOT</version>
```
damencho, line 104: I guess this is temporary?

gpolitis: yep

src/main/java/org/jitsi/jicofo/conference/colibri/ColibriConferenceImpl.java
```diff
@@ -125,6 +125,16 @@ public void setGID(String gid)
         conferenceState.setGID(gid);
     }
 
+    @Override
+    public void setRtcStatsEnabled(boolean rtcStatsEnabled) {
+        conferenceState.setRtcStatsEnabled(rtcStatsEnabled);
```
bgrozev, line 130: Have you tested that this? The code is super hairy and hard to follow, but I think the requests go through `colibriBuilder`, which gets reset at certain times, and `reset()` doesn't re-read the new flags

gpolitis: I just did and it doesn't work :-( I guess you nailed the problem.. I'll dig into it.

gpolitis:
Blocked by https://github.com/jitsi/jicofo/pull/832

bgrozev (approved):

PR #831, opened 2021-11-12
Add RFC 2198 fmtp line for Opus RED

paweldomas:
Opus RED is now enabled on Chrome M96, but it requires this extra FMTP line.

### DIFF
--- a/src/main/java/org/jitsi/jicofo/codec/JingleOfferFactory.java
+++ b/src/main/java/org/jitsi/jicofo/codec/JingleOfferFactory.java
@@ -449,6 +449,11 @@ public class JingleOfferFactory
             {
                 PayloadTypePacketExtension red = addPayloadTypeExtension(rtpDesc, config.opus.red.pt(), "red", 48000);
                 red.setChannels(2);
+
+                // RFC 2198 fmtp line
+                // Indicates Opus payload type (111) as primary/secondary encoding for RED (112)
+                // fmtp:112 111/111
+                addParameterExtension(red, null, config.opus.pt() + "/" + config.opus.pt());
             }
 
             // a=rtpmap:111 opus/48000/2


nils-ohlmeier (approved):
LGTM

paweldomas:
jenkins test this please

paweldomas:
jenkins test this please

src/main/java/org/jitsi/jicofo/codec/JingleOfferFactory.java
```diff
@@ -449,6 +449,11 @@ private static void addAudioToContent(ContentPacketExtension content, OfferOptio
             {
                 PayloadTypePacketExtension red = addPayloadTypeExtension(rtpDesc, config.opus.red.pt(), "red", 48000);
                 red.setChannels(2);
+
+                // RFC 2198 fmtp line
+                // Indicates Opus payload type (111) as primary/secondary encoding for RED (112)
+                // fmtp:112 111/111
+                addParameterExtension(red, null, config.opus.pt() + "/" + config.opus.pt());
```
bgrozev, line 456: What are the semantics of "111/111"? As far as I understand 2198 (`we define the format
   parameters to be a slash "/" separated list of RTP payload types`) these are just formats that are declared usable with RED, so just "111" should suffice. But I haven't read the RFC carefully.

paweldomas: you mean that if there's just one then it should be used for both primary and secondary?

paweldomas: I did it this way, because that's what Phippo's post mentions here https://groups.google.com/g/discuss-webrtc/c/5761etCrSuA

and also I was able to confirm that it's working with Chrome. Not sure about the other way around.

bgrozev: I read it again now, and I think I understand what you're saying.

bgrozev (approved):

PR #832, opened 2021-11-16
Cleanup, update dependencies

ibauersachs:
Cleanups, test with Java 17

Needs
- jitsi/jitsi-xmpp-extensions#55
- jitsi/jicoco#157
- jitsi/jitsi-videobridge#1760


pom.xml
```diff
@@ -385,7 +441,7 @@
       <plugin>
         <groupId>com.github.spotbugs</groupId>
         <artifactId>spotbugs-maven-plugin</artifactId>
-        <version>${spotbugs.version}</version>
+        <version>4.4.2.2</version>
```
bgrozev, line 444: Does this intentionally update only the maven plugin and not `spotbugs` and `spotbugs-annotations`?

src/main/java/org/jitsi/impl/protocol/xmpp/ChatMemberImpl.java
```diff
@@ -297,12 +297,12 @@ void processPresence(Presence presence)
         }
 
         boolean wasVideoMuted = isVideoMuted;
-        StandardExtensionElement videoMutedExt = presence.getExtension("videomuted", "jabber:client");
-        isVideoMuted = videoMutedExt == null || Boolean.valueOf(videoMutedExt.getText()); /* defaults to true */
+        VideoMutedExtension videoMutedExt = presence.getExtension(VideoMutedExtension.class);
```
bgrozev, line 300: I think this is incorrect. The `VideoMutedExtension` class uses a custom name space (for whatever reason), while the jitsi-meet client does not bother adding a namespace and it inherits "jabber:client".

bgrozev:
Can you merge master please?

pom.xml
ibauersachs: Yes, for now. Spotbugs 4.5 fails with messages about exposed arrays. I didn't want to address that in this PR as it is big enough by itself. The maven plugin update is necessary for Java 17.

src/main/java/org/jitsi/impl/protocol/xmpp/ChatMemberImpl.java
ibauersachs: I'm tempted to update jitsi-xmpp-extensions to set `org.jitsi.xmpp.extensions.jitsimeet.VideoMutedExtension#NAMESPACE` to `jabber:client`. The element isn't used anywhere before the above change.
WDYT?

bgrozev:
> I'm tempted to update jitsi-xmpp-extensions to set `org.jitsi.xmpp.extensions.jitsimeet.VideoMutedExtension#NAMESPACE` to `jabber:client`. The element isn't used anywhere before the above change.

This makes sense. 

> Edit: the other option would be to update the client to send a proper namespace. The code is relatively new (just one week Nov. 9, 2021), so I guess there's not a lot of client base yet.

I don't follow. The code in the client has been there for a while. However, I'm pretty sure the client's XML parser/code ignores namespaces so it wouldn't be a breaking change. Still, I'd rather not run the risk of sometthing getting out of sync just for fix a namespace.

ibauersachs:
> > I'm tempted to update jitsi-xmpp-extensions to set `org.jitsi.xmpp.extensions.jitsimeet.VideoMutedExtension#NAMESPACE` to `jabber:client`. The element isn't used anywhere before the above change.
> 
> This makes sense.
> 
> > Edit: the other option would be to update the client to send a proper namespace. The code is relatively new (just one week Nov. 9, 2021), so I guess there's not a lot of client base yet.
> 
> I don't follow.

The video muted extension is only parsed since that date above in Jicofo, so I assumed it's new in the client too.

> The code in the client has been there for a while. However, I'm pretty sure the client's XML parser/code ignores namespaces so it wouldn't be a breaking change.

The client would still need to add the namespace, so it would still be breaking change.

> Still, I'd rather not run the risk of sometthing getting out of sync just for fix a namespace.

Well, part of the mess in jitsi-xmpp-extensions is the lack of setting a proper namespace to the extensions in the first place.

bgrozev:
> The client would still need to add the namespace, so it would still be breaking change.

The code in jicofo is not actively used yet. It's behind a config flag, waiting for changes in the client.

src/main/java/org/jitsi/impl/protocol/xmpp/ChatMemberImpl.java
paweldomas: if namespaces will be changed, be careful for breaking older mobile clients if they would not be sending the new one and if jicofo could ignore it?

paweldomas: oh I didn't read the whole thing at first, Boris wrote it's behind a feature flag? then should be fine

ibauersachs: @paweldomas see jitsi/jitsi-xmpp-extensions#56. The `VideoMutedExtension` was never used - unless the GitHub code search is tricking me.

bgrozev:
LGTM, but can you resolve the conflicts in the JVB PR so the tests can run (the test output is very misleading, it contains successful torture tests results from a previous run, in reality the job failed before even trying to build the projects).

ibauersachs:
JVB is rebased, I'm not sure what I need to look at in those log outputs.

JonathanLennox:
I'm slightly concerned that the on-the-wire XML namespaces of the types you split (RtcpMuxPacketExtension and CandidatePacketExtension) will now end up getting encoded whereas previously they weren't.  I don't think this will cause problems for the client (which as I understand it ignores namespaces) but I'm concerned this might make our XMPP messaging bigger, increasing load on the Prosody server (this is already a bottleneck for large conferences).  Can you take a look at the jicofo-xmpp logs before and after this change and see if there's any problem there?

JonathanLennox:
> JVB is rebased, I'm not sure what I need to look at in those log outputs.

```
Auto-merging pom.xml
CONFLICT (content): Merge conflict in pom.xml
Auto-merging jvb/src/main/kotlin/org/jitsi/videobridge/transport/ice/IceTransport.kt
```

bgrozev:
Reading `AbstractPacketExtension.toXml` it looks like [it will not encode the namespace when it matches the parent](https://github.com/jitsi/jitsi-xmpp-extensions/blob/master/src/main/java/org/jitsi/xmpp/extensions/AbstractPacketExtension.java#L158) (this now happens in [XmlStringBuilder](https://github.com/igniterealtime/Smack/blob/master/smack-core/src/main/java/org/jivesoftware/smack/util/XmlStringBuilder.java#L58) actually).

bgrozev:
```
2021-11-19 20:16:05.174 SEVERE: [22] org.jivesoftware.smack.AbstractXMPPConnection.lambda$invokeStanzaCollectorsAndNotifyRecvListeners$8: Exception in packet listener
java.lang.IllegalStateException: Extension element is not of expected class 'org.jitsi.xmpp.extensions.jitsimeet.VideoMutedExtension', because there is no according extension element provider registered with ProviderManager for {jabber:client}videomuted. WARNING: This indicates a serious problem with your Smack setup, probably causing Smack not being able to properly initialize itself.
	at org.jivesoftware.smack.util.XmppElementUtil.castOrThrow(XmppElementUtil.java:111)
	at org.jivesoftware.smack.packet.StanzaView.getExtension(StanzaView.java:98)
	at org.jitsi.impl.protocol.xmpp.ChatMemberImpl.processPresence(ChatMemberImpl.java:300)
	at org.jitsi.impl.protocol.xmpp.ChatRoomImpl.processOtherPresence(ChatRoomImpl.java:782)
	at org.jitsi.impl.protocol.xmpp.ChatRoomImpl.processPresence(ChatRoomImpl.java:842)
	at org.jivesoftware.smackx.muc.MultiUserChat$3.processStanza(MultiUserChat.java:294)
	at org.jivesoftware.smack.AbstractXMPPConnection.lambda$invokeStanzaCollectorsAndNotifyRecvListeners$8(AbstractXMPPConnection.java:1626)
	at org.jivesoftware.smack.AsyncButOrdered$Handler.run(AsyncButOrdered.java:151)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:748)
```

Does VideoMutedExtension need to be registered somewhere?

ibauersachs:
> Does VideoMutedExtension need to be registered somewhere?

Yes, in the ProviderManager of Smack. I'll push an update.

ibauersachs:
> > JVB is rebased, I'm not sure what I need to look at in those log outputs.
> 
> ```
> Auto-merging pom.xml
> CONFLICT (content): Merge conflict in pom.xml
> Auto-merging jvb/src/main/kotlin/org/jitsi/videobridge/transport/ice/IceTransport.kt
> ```

Was that from on old test before I rebased on Boris' ice4j cleanup?

JonathanLennox:
> Reading `AbstractPacketExtension.toXml` it looks like [it will not encode the namespace when it matches the parent](https://github.com/jitsi/jitsi-xmpp-extensions/blob/master/src/main/java/org/jitsi/xmpp/extensions/AbstractPacketExtension.java#L158) (this now happens in [XmlStringBuilder](https://github.com/igniterealtime/Smack/blob/master/smack-core/src/main/java/org/jivesoftware/smack/util/XmlStringBuilder.java#L58) actually).

Ok - it looks like both Jingle and Colibri use the `urn:xmpp:jingle:transports:ice-udp:1` namespace for `<transport>` objects, so copying CandidatePacketExtensions between them is still safe even as the extensions themselves get namespaces.

ibauersachs:
Seems there's some broken stuff left :-(
- Something with the presence extension/filter @JonathanLennox already flagged in jitsi/jicoco#157
- An unparsable stanza, but I don't know which one yet

JonathanLennox:
> Was that from on old test before I rebased on Boris' ice4j cleanup?

No, I think it's just the pom?  Sorry, the other merge notification was probably a red herring.

ibauersachs:
jenkins test this please

ibauersachs:
jenkins test this please

bgrozev:
LGTM, just the jitsi-videobridge version remains to be fixed.

bgrozev (approved):

bgrozev:
Thank you for your work, Ingo!

PR #836, opened 2021-11-19
limit audio senders, as was done for video senders.

jqdrqgnq:


src/main/java/org/jitsi/impl/protocol/xmpp/ChatMemberImpl.java
```diff
@@ -296,13 +304,26 @@ void processPresence(Presence presence)
             statsId = statsIdPacketExt.getStatsId();
         }
 
+        boolean wasAudioMuted = isAudioMuted;
+        StandardExtensionElement audioMutedExt = presence.getExtension("audiomuted", "jabber:client");
```
bgrozev, line 308: Ingo has been doing some work around extensions in #832 and he fixed `VideoMutedExtension`. Can you please use the same approach for audio?

src/main/java/org/jitsi/impl/protocol/xmpp/ChatMemberImpl.java
```diff
@@ -296,13 +304,26 @@ void processPresence(Presence presence)
...
+        isAudioMuted = audioMutedExt == null || Boolean.valueOf(audioMutedExt.getText()); /* defaults to true */
+
+        if (isAudioMuted != wasAudioMuted)
+        {
+            logger.debug(() -> toString() + ". isAudioMuted = " + isAudioMuted + ".");
```
bgrozev, line 313: If you need extra context for the `ChatMemberImpl` instance in the logs you can add it to the logger, e.g.
`logger.addContext("occupantJid", occupantJid.toString());`. We could also hook it up to the "parent" logger so it has context for the chat room/conference.

src/main/java/org/jitsi/impl/protocol/xmpp/ChatMemberImpl.java
jqdrqgnq: done

src/main/java/org/jitsi/impl/protocol/xmpp/ChatMemberImpl.java
jqdrqgnq: added occupantJid to the logging context.

jqdrqgnq:
Requires #839 to eliminate the accumulating of old conference-properties elements within presence updates.

bgrozev (approved):

PR #838, opened 2021-11-22
ref: Simplify code sending presence.

bgrozev:


bgrozev:
This is wrong, I overlooked

PR #845, opened 2021-12-02
fix logger context.

jqdrqgnq:
shouldn't have set per-instance context on a static logger object.


bgrozev:
Can you make `ChatMemberImpl.logger` an instance field and initialize it using the `ChatRoomImpl`'s logger as a parent logger (see for example how `Participant` initializes its logger)?

bgrozev:
Test failures are probably unrelated. @damencho do you know if these tests are known to fail?

damencho:
👍  Yep ignore those ...

PR #853, opened 2021-12-17
Fix null references

bgrozev:


src/main/kotlin/org/jitsi/jicofo/xmpp/ConferenceIqHandler.kt
```diff
@@ -130,9 +130,10 @@ class ConferenceIqHandler(
                 // authentication for the main room so users can go back to it.
                 var breakoutRoomExists: Boolean = false
                 for (conference in focusManager.getConferences()) {
-                    if (conference.chatRoom.isBreakoutRoom && room.toString() == conference.chatRoom.mainRoom) {
-                        breakoutRoomExists = true
-                        break
+                    conference.chatRoom?.let {
+                        if (it.isBreakoutRoom && room.toString() == it.mainRoom) {
+                            breakoutRoomExists = true
```
damencho: Is dropping the break intentional?

damencho: O that's covered in a later commit.

damencho (approved):

src/main/kotlin/org/jitsi/jicofo/xmpp/ConferenceIqHandler.kt
bgrozev: Yes, nice catch, it was an awkward way to do it.

PR #854, opened 2021-12-17
chore: Remove the sentry and agafua-syslog runtime dependencies.

bgrozev:
This is to stop shipping non-essential jars and reduce potential exposure to vulnerabilities. Setting up sentry requires setting an environment variable (and potentially more changes?), so it should be easy to adapt to also include the sentry jar in the classpath.

cc @saghul @sapkra

### DIFF
--- a/lib/logging.properties
+++ b/lib/logging.properties
@@ -31,8 +31,8 @@ com.agafua.syslog.SyslogHandler.formatter = org.jitsi.utils.logging2.JitsiLogFor
 com.agafua.syslog.SyslogHandler.escapeNewlines = false
 com.agafua.syslog.SyslogHandler.filter = org.jitsi.impl.protocol.xmpp.log.ExcludeXmppPackets
 
-# Sentry (uncomment handler to use)
-io.sentry.jul.SentryHandler.level=WARNING
+# Sentry (also add the sentry handler to "handlers" to use)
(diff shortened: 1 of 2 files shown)


sapkra:
@bgrozev How will I be able to use error tracking in the future? I'm not that familiar with java setups. To enable a sentry handler it needs to be installed somehow.

damencho:
If you put the jar file in /usr/share/jicofo/libs before starting it and have the correct logging config, that is enough.

bgrozev:
Or add the sentry jar file to the classpath. We can add an env variable for additional classpath entries if that helps.

saghul:
Hasn't Sentry released a fixed version?

We expose this in the Docker setup, so it would be broken now.

Sentry is pretty well regarded in the industry, FWIW.

bgrozev:
How do we expose it in docker? Can we ship the jar and insert it in the classpath in the docker setup (ideally behind a flag, since not every docker uses uses sentry)? I'd be happy to work on this if you can point me to the right place.

I don't doubt that sentry is well regarded. Still, it's additional surface area and an additional dependency to be maintained, which is only used by a very small subset of users. And there's a simple way to avoid it.

saghul:
Fair point.

We esport env variables to turn it on, they are part of the config template file.

Yeah, we could ship it there indeed.

PR #855, opened 2021-12-20
colibri2 and secure octo 

JonathanLennox:


src/main/resources/reference.conf
```diff
@@ -153,6 +153,10 @@ jicofo {
     }
   }
 
+  colibri {
+    // Whether to use the new version of COLIBRI
+    enable-colibri2 = true
```
bgrozev, line 158: Note to self: flip this before merging.

src/main/kotlin/org/jitsi/jicofo/conference/colibri/v2/ColibriV2SessionManager.kt
```diff
@@ -79,20 +79,27 @@ class ColibriV2SessionManager(
...
-        val participantInfo = participants[participant.endpointId]
-            ?: run {
-                logger.warn("No ParticipantInfo for participant $participant")
-                return
-            }
-        participantInfo.session.expire(participantInfo)
-        remove(participantInfo)
+        val participantInfos = participants.mapNotNull { this.participants[it.endpointId] }
+        if (participantInfos.size != participants.size) {
+            logger.error("Can not remove every participant requested, ParticipantInfo missing: " +
```
JonathanLennox: Is this a "should not happen", a "could happen in a race", or normal?

bgrozev: It's actually a "happens during normal operation because of awkward APIs": bridge goes down, JitsiMeetConference calls colibriSessionManager.removeBridges (which in v2 removes the participants), JitsiMeetConfernce calls removeParticipants. I've been thinking about how to avoid it, I think I'll just hide the message for now. When we remove colibri v1 we can clean it up.

bgrozev (approved):
Approving with @JonathanLennox 's consent

JonathanLennox (commented):
I approve, but since this is nominally my PR, I can't stamp it.

PR #860, opened 2022-01-09
feat: Add a debug HTTP interface.

bgrozev:


JonathanLennox:
How hard would it be to add a unit test for this?  (Particularly, making sure the debug output is valid JSON, we've run into bugs with that in JVB in the past.)

bgrozev:
Good point, I've added some tests that cover most of the code.

PR #862, opened 2022-01-11
fix(multi-stream): generate source name

paweldomas:
...for injected audio sources.

It is required, because of this constraint:

https://github.com/jitsi/jitsi-videobridge/pull/1793/files#diff-9d9ade953a28940dd8dc6ad1a9378bacde2eab16400e03959d3bac1cb539f46fL295

I think it's okay to generate the source-name here, because it's Jicofo who's in charge of signaling this one and my understanding is that it's removed eventually.


paweldomas:
> The injected SSRC is actually not removed from the participant's sources. We noticed this yesterday. It probably should be removed, but I don't want to waste time trying to figure out how -- colibri2 is very close and we don't use it there. For that reason -- would participantId + "-a0" clash with a source name advertised by the client?

Yeah, I think it will clash. This is still subject to change depending on how the backwards compatibility mode will work exactly (will it be in jicofo or jvb etc.). I need to at least put it behind a flag. The main purpose of this is to unblock work on multi stream, because the conference started crashing ever since I added a hard requirement for source name to always be there in the jvb.

bgrozev:
I don't even think this needs to be behind a flag. Just make it `participantId-injected0` so it doesn't clash later?

bgrozev (approved):

PR #868, opened 2022-01-24
Fix removing sources for octo

bgrozev:
- fix: Remove sources when a participant leaved regardless of the presense of a jingle session.
- fix: Remove sources from octo  when a participant leaves.


src/main/java/org/jitsi/jicofo/conference/JitsiMeetConferenceImpl.java
```diff
@@ -854,7 +854,7 @@ private void terminateParticipant(
                 // This is inefficient because it sends one colibri request to remove the sources and another one
                 // to expire the channels/endpoint.
```
JonathanLennox, line 855: Is this comment still true?

JonathanLennox (approved):

PR #877, opened 2022-02-11
feat: Add SCTP with colibri2.

bgrozev:


src/main/kotlin/org/jitsi/jicofo/conference/colibri/ColibriAllocation.kt
```diff
@@ -14,5 +14,7 @@ data class ColibriAllocation(
     /** The region of the bridge */
     val region: String?,
     /** An identifier for the session */
-    val bridgeSessionId: String?
+    val bridgeSessionId: String?,
+    /** The SCTP port (used for both local and remote port) advertised by the bridge, if any */
+    val sctpPort: Int?
```
bbaldino, line 19: Would it be more 'obvious' to put the default port here rather than in `ParticipantInviteRunnable`?

src/main/kotlin/org/jitsi/jicofo/conference/colibri/v2/ColibriV2SessionManager.kt
```diff
@@ -276,6 +276,7 @@ class ColibriV2SessionManager(
         val session: Colibri2Session
         val created: Boolean
         val participantInfo: ParticipantInfo
+        val useSctp = contents.any { it.name == "data" }
```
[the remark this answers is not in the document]
bgrozev: No :)
It's just that the internal representation of an "offer" in jicofo uses a list of Jingle contents. I'm looking forward to cleaning that up once colibri1 is removed (and SCTP is the last feature needed for that, apart from more testing and fixes).

src/main/kotlin/org/jitsi/jicofo/conference/colibri/ColibriAllocation.kt
bgrozev: I don't want to set the default port here, because in the future when there's only colibri2 there will be no default port and `ColibriAllocation.sctpPort` will control whether SCTP is signaled to the client or not.

I removed the double definition, I think it's a bit less confusing now

JonathanLennox:
LGTM, but will need the jvb version bump before merging.

PR #879, opened 2022-02-15
feat: Take into account region groups for selection.

bgrozev:
- feat: Add region groups config.
- feat: Implement selection with region groups.

### DIFF
--- a/src/main/java/org/jitsi/jicofo/bridge/BridgeSelectionStrategy.java
+++ b/src/main/java/org/jitsi/jicofo/bridge/BridgeSelectionStrategy.java
@@ -272,7 +272,7 @@ public abstract class BridgeSelectionStrategy
 
         if (result.isPresent())
         {
-            totalNotLoadedInRegion++;
+            totalNotLoadedInRegionGroup++;
             updateSplitStats(conferenceBridges, result.get(), null, participantRegionGroup);
             logSelection(result.get(), conferenceBridges, null, participantRegionGroup);
         }
@@ -355,7 +355,7 @@ public abstract class BridgeSelectionStrategy
 
         if (result.isPresent())
         {
-            totalLeastLoadedAlreadyInConferenceInRegion++;
+            totalLeastLoadedAlreadyInConferenceInRegionGroup++;
             logSelection(result.get(), conferenceBridges, null, participantRegionGroup);
         }
 
--- a/src/main/java/org/jitsi/jicofo/bridge/BridgeSelectionStrategy.java
+++ b/src/main/java/org/jitsi/jicofo/bridge/BridgeSelectionStrategy.java
@@ -111,7 +111,7 @@ public abstract class BridgeSelectionStrategy
             Bridge bridge = doSelect(bridges, conferenceBridges, participantRegion);
             if (bridge != null)
             {
-                logger.info("Selected initial bridge " + bridge
+                logger.debug("Selected initial bridge " + bridge
                         + " with reported stress=" + bridge.getLastReportedStressLevel()
                         + " for participantRegion=" + participantRegion
                         + " using strategy " + this.getClass().getSimpleName());
--- a/src/main/java/org/jitsi/jicofo/bridge/BridgeSelectionStrategy.java
+++ b/src/main/java/org/jitsi/jicofo/bridge/BridgeSelectionStrategy.java
@@ -1,5 +1,6 @@
 package org.jitsi.jicofo.bridge;
 
+import org.jetbrains.annotations.*;
 import org.jitsi.utils.logging2.*;
 import org.json.simple.*;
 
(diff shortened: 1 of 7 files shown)


src/main/kotlin/org/jitsi/jicofo/bridge/BridgeConfig.kt
```diff
@@ -118,6 +121,17 @@ class BridgeConfig private constructor() {
         "jicofo.bridge.xmpp-connection-name".from(JitsiConfig.newConfig)
     }
 
+    val regionGroups: Set<Set<String>> by config {
+        "jicofo.bridge".from(JitsiConfig.newConfig).convertFrom<ConfigObject> {
+            val regionGroups = it["region-groups"] as? ConfigList ?: emptyList<ConfigValue>()
```
JonathanLennox, line 126: Having both these vals called `regionGroups` is a bit confusing

src/main/java/org/jitsi/jicofo/bridge/BridgeSelectionStrategy.java
```diff
@@ -208,17 +260,49 @@ private void logSelection(
...
+            Set<String> participantRegionGroup)
+    {
+        Optional<Bridge> result = bridges.stream()
+                .filter(not(b -> isOverloaded(b, conferenceBridges)))
+                .filter(inRegionGroup(participantRegionGroup))
+                .findFirst();
+
+        if (result.isPresent())
+        {
+            totalNotLoadedInRegion++;
```
JonathanLennox, line 275: Shouldn't this be incrementing `totalNotLoadedInRegionGroup`?

src/main/java/org/jitsi/jicofo/bridge/RegionBasedBridgeSelectionStrategy.java
```diff
@@ -67,11 +91,17 @@ public Bridge doSelect(
...
                 () -> notLoadedInRegion(bridges, conferenceBridges, participantRegion).orElseGet(
+                () -> notLoadedInRegionGroup(bridges, conferenceBridges, getRegionGroup(participantRegion)).orElseGet(
                 () -> leastLoadedAlreadyInConferenceInRegion(bridges, conferenceBridges, participantRegion).orElseGet(
+                () -> leastLoadedAlreadyInConferenceInRegionGroup(
+                        bridges, conferenceBridges, getRegionGroup(participantRegion)).orElseGet(
                 () -> leastLoadedInRegion(bridges, conferenceBridges, participantRegion).orElseGet(
+                () -> leastLoadedInRegionGroup(bridges, conferenceBridges, getRegionGroup(participantRegion)).orElseGet(
                 () -> nonLoadedAlreadyInConference(bridges, conferenceBridges, participantRegion).orElseGet(
-                () -> leastLoaded(bridges, conferenceBridges, participantRegion).orElse(null))))));
+                () -> leastLoaded(bridges, conferenceBridges, participantRegion).orElse(null))))))))));
```
JonathanLennox, line 104: I'm worried this code is beginning to traverse the same list of bridges over and over again with different criteria, it might be worth thinking about how we could make this more efficient (especially as we start having larger bridge pools).

src/main/kotlin/org/jitsi/jicofo/bridge/BridgeConfig.kt
```diff
@@ -123,8 +123,8 @@ class BridgeConfig private constructor() {
 
     val regionGroups: Set<Set<String>> by config {
         "jicofo.bridge".from(JitsiConfig.newConfig).convertFrom<ConfigObject> {
-            val regionGroups = it["region-groups"] as? ConfigList ?: emptyList<ConfigValue>()
-            regionGroups.map { regionGroupConfigList ->
+            val regionGroupsConfigList = it["region-groups"] as? ConfigList ?: emptyList<ConfigValue>()
+            regionGroupsConfigList.map { regionGroupConfigList ->
```
JonathanLennox, line 127: Now you have two vals called `regionGroupConfigList`.  🙂

src/main/java/org/jitsi/jicofo/bridge/RegionBasedBridgeSelectionStrategy.java
JonathanLennox: Yeah, if your performance tests don't show anything ridiculous, that's fine.

src/main/kotlin/org/jitsi/jicofo/bridge/BridgeConfig.kt
bgrozev: No, one is `regionGroup*s*ConfigList` (which is the list of region groups) and the other is `regionGroupConfigList` (which is the list of regions in the group). I tried to improve the naming a bit. I didn't intentionally make this hard to read, I promise :)

src/main/java/org/jitsi/jicofo/bridge/RegionBasedBridgeSelectionStrategy.java
bgrozev: I added some perf tests. On my machine selection from 500 bridges takes about 0.5 ms. I forced the worst-case for the iteration over the lists, though the sorting might be slower in practice. I think that's reasonable.

JonathanLennox (approved):

PR #881, opened 2022-02-16
Use drain mode flag in bridge selection.

jqdrqgnq:
Prefer active bridges to draining ones.
Fail selection if pinned to a version different than the
one the conference is already using.

### DIFF
--- a/src/test/kotlin/org/jitsi/jicofo/bridge/BridgeReleaseTest.kt
+++ b/src/test/kotlin/org/jitsi/jicofo/bridge/BridgeReleaseTest.kt
@@ -17,7 +17,6 @@
  */
 package org.jitsi.jicofo.bridge
 
-import io.kotest.core.annotation.Ignored
 import io.kotest.core.spec.IsolationMode
 import io.kotest.core.spec.Spec
 import io.kotest.core.spec.style.ShouldSpec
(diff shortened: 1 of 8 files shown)


src/test/kotlin/org/jitsi/jicofo/util/ConfigHelpers.kt
```diff
@@ -23,3 +23,9 @@ fun ShouldSpec.context(config: String, name: String, block: () -> Unit) = contex
         block()
     }
 }
+
+fun ShouldSpec.xcontext(config: String, name: String, block: () -> Unit) = xcontext(name) {
```
bgrozev, line 27: Do you use this anywhere?

src/test/kotlin/org/jitsi/jicofo/bridge/BridgeReleaseTest.kt
```diff
@@ -0,0 +1,162 @@
...
+import io.kotest.matchers.shouldBe
+import org.jitsi.config.setNewConfig
+import org.jitsi.test.time.FakeClock
+import org.jxmpp.jid.impl.JidCreate
+
+/**
+ * This test simulates the intended procedure to upgrade the set of bridges connected to a jicofo instance, and
+ * verifies the selection logic.
+ */
+@Ignored
```
bgrozev, line 33: Let's enable the test now that it passes :)

bgrozev (approved):

PR #888, opened 2022-03-04
Improve jibri logs

bgrozev:
- log: Inherit log context in JibriSession.
- log: Log when at jibri IQ is dropped.
- log: Log all received jibri requests.

### DIFF
--- a/src/main/kotlin/org/jitsi/jicofo/jibri/BaseJibri.kt
+++ b/src/main/kotlin/org/jitsi/jicofo/jibri/BaseJibri.kt
@@ -72,6 +72,7 @@ abstract class BaseJibri internal constructor(
 
     fun handleJibriRequest(request: JibriRequest): IqProcessingResult =
         if (accept(request.iq)) {
+            logger.info("Accepted jibri request: ${request.iq.toXML()}")
             incomingIqQueue.add(request)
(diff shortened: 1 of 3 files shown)


src/main/kotlin/org/jitsi/jicofo/jibri/BaseJibri.kt
```diff
@@ -72,6 +72,7 @@ abstract class BaseJibri internal constructor(
 
     fun handleJibriRequest(request: JibriRequest): IqProcessingResult =
         if (accept(request.iq)) {
+            logger.info("Accepted jibri request: ${request.iq.toXML()}")
```
JonathanLennox, line 75: Most "ordinary" XMPP logging is done at either `debug` or `trace` level - do you really want this as `info`?

bgrozev: Yes, I erred on the side of logging too much here, because it's hard to log all the useful information otherwise (is the request coming from a client or a jibri, which client/jibri is it coming from, is it a request or status update, is there an error condition, etc). Jibri messages tend to be small and not that many, so the volume should be okay.

JonathanLennox (approved):

PR #889, opened 2022-03-10
RFC: ExternalBridgeSelectionStrategy

jbg:
This is very similar to a patch we've been using at AVStack for a while, and I think it could be useful for others.

Basically our bridge selection logic is a bit more complex than we'd like to express inside Jicofo itself, so we have Jicofo call an external (to Jicofo) service for bridge selection decisions. A fallback to another strategy is implemented in case it's unable to reach the service.

This lets us look up the customer's bridge preferences in a database (allowing customers to set up rules about which regions are allowed or denied themselves), as well as implement failover to nearby regions etc.

I've marked this as RFC because I'm not sure if it's something you want included in Jicofo itself, and also because the use of `HttpClient` enforces a Java 11 minimum. I'm not sure if there is a published policy about minimum Java version; if we still need to support older than 11 then we could swap out the use of `HttpClient` for something else.


jbg:
Well, I guess the existence of CI testing Java 8 tells me that using `HttpClient` is a no-no — let me know if this is something you'd be interested in merging, and if so I'll work on swapping `HttpClient` out for something Java 8 friendly.

saghul:
We've been deploying jicofo on Java 11 AFAIK. Maybe we ought to update the CI checks? /cc @bgrozev ?

bgrozev:
Thanks, that's definitely a good contribution! I'll check back with the team to see if there's anything preventing us from dropping support for java 8.

damencho:
Maybe Ubuntu 18.04 ... but that is 4 years old now ... 
And anyway it will skip the update and for new installation, and you will need to update java to 11 ... so maybe just edit in the quick install guide as we have for prosody update to 11, that is all we need ...

bgrozev:
We're working on dropping java8 support, and will review/merge when that's done.

src/main/kotlin/org/jitsi/jicofo/bridge/ExternalBridgeSelectionStrategy.kt
```diff
@@ -0,0 +1,110 @@
...
+import java.net.URI
+import java.net.http.HttpClient
+import java.net.http.HttpRequest
+import java.net.http.HttpResponse
+
+@Suppress("unused")
+class ExternalBridgeSelectionStrategy() : BridgeSelectionStrategy() {
+    private val httpClient: HttpClient = HttpClient
+        .newBuilder()
+        .connectTimeout(ExternalBridgeSelectionStrategyConfig.config.timeout)
```
bgrozev, line 16: You can import `ExternalBridgeSelectionStrategyConfig.config` to make these calls shorter

src/main/kotlin/org/jitsi/jicofo/bridge/ExternalBridgeSelectionStrategy.kt
```diff
@@ -0,0 +1,110 @@
...
+    override fun doSelect(
+        bridges: MutableList<Bridge>?,
+        conferenceBridges: MutableMap<Bridge, Int>?,
+        participantRegion: String?
+    ): Bridge {
+        val url = ExternalBridgeSelectionStrategyConfig.config.url
+            ?: throw Exception("ExternalBridgeSelectionStrategy requires url to be provided")
+
+        val requestBody = JSONObject()
+        requestBody["bridges"] = bridges?.map {
```
bgrozev, line 52: I just realized that you are providing the full list of bridges here. Can you make that optional (behind a config flag) please? We intend to use it with a service which already has the list of available bridges.

src/main/kotlin/org/jitsi/jicofo/bridge/ExternalBridgeSelectionStrategy.kt
```diff
@@ -0,0 +1,110 @@
...
+        }
+
+        val selectedBridgeIndex = responseBody["selected_bridge_index"] as? Number
+
+        if (selectedBridgeIndex == null) {
+            logger.error("ExternalBridgeSelectionStrategy: HTTP response selected_bridge_index missing or invalid, using fallback strategy")
+            return fallback(bridges, conferenceBridges, participantRegion)
+        }
+
+        val bridge = bridges!![selectedBridgeIndex.toInt()]
```
bgrozev, line 106: Can you fail early (before sending the request) if `bridges` is null or empty?

bgrozev, line 106: What do you think about having the external service return an ID (JID) of the selected bridge instead of an index? We want to have the service return a bridge that jicofo doesn't yet know about (this will require more changes, but it will be easier to have the HTTP API ready to support it and not have to change).

src/main/kotlin/org/jitsi/jicofo/bridge/ExternalBridgeSelectionStrategy.kt
```diff
@@ -0,0 +1,110 @@
...
+        requestBody["conference_bridges"] = conferenceBridges?.mapKeys { it.key.jid.toString() }
+        requestBody["participant_region"] = participantRegion
+        requestBody["fallback_strategy"] = ExternalBridgeSelectionStrategyConfig.config.fallbackStrategy
+
+        val request = HttpRequest
+            .newBuilder()
+            .POST(HttpRequest.BodyPublishers.ofString(requestBody.toJSONString()))
+            .uri(URI.create(url))
+            .headers("Content-Type", "application/json")
+            .timeout(ExternalBridgeSelectionStrategyConfig.config.timeout)
```
bgrozev, line 74: Will this work as expected when the timeout is `null`? You could either only call `.timeout` if the config has a value, or make the config value required (`by config` instead of `by optionalconfig`) and add a reasonable default value in `reference.conf`

src/main/kotlin/org/jitsi/jicofo/bridge/ExternalBridgeSelectionStrategyConfig.kt
```diff
@@ -0,0 +1,29 @@
...
+        "${BASE}.url".from(JitsiConfig.newConfig)
+    }
+    fun url() = url
+
+    val timeout: Duration? by optionalconfig {
+        "${BASE}.timeout".from(JitsiConfig.newConfig)
+    }
+    fun timeout() = timeout
+
+    val fallbackStrategy: String? by optionalconfig {
```
bgrozev, line 18: Perhaps it's better to make this required (`by config`) and provide a good default value?

bgrozev:
One more thing: can you please add a commented out section to `reference.conf` with the new options? For the purpose of documentation

jbg:
@bgrozev Thanks for the review and sorry for the long delay before getting back to this. I'll rebase and address your comments shortly.

src/main/kotlin/org/jitsi/jicofo/bridge/ExternalBridgeSelectionStrategy.kt
jbg: Yes, this seems sensible. Interesting use case to use a bridge it doesn't know about yet. Launching bridges on-demand?

src/main/kotlin/org/jitsi/jicofo/bridge/ExternalBridgeSelectionStrategy.kt
jbg: Good catch, `.timeout(null)` [will throw](https://github.com/AdoptOpenJDK/openjdk-jdk11/blob/master/src/java.net.http/share/classes/jdk/internal/net/http/HttpRequestBuilderImpl.java#L231). I think making it required with a default in `reference.conf` is sensible; there is no rational reason to want to block forever...

src/main/kotlin/org/jitsi/jicofo/bridge/ExternalBridgeSelectionStrategy.kt
```diff
@@ -0,0 +1,110 @@
...
+                "region" to it.region,
+                "stress" to it.stress,
+                "operational" to it.isOperational,
+                "graceful_shutdown" to it.isInGracefulShutdown,
+                "draining" to it.isDraining,
+            )
+        }
+        requestBody["conference_bridges"] = conferenceBridges?.mapKeys { it.key.jid.toString() }
+        requestBody["participant_region"] = participantRegion
+        requestBody["fallback_strategy"] = ExternalBridgeSelectionStrategyConfig.config.fallbackStrategy
```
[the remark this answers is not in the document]
jbg: We were using it to signal which mode the external service should use (region based or intra region), matched to the fallback strategy Jicofo would use if it couldn't contact the external service for any reason. We're actually now passing the selected mode in the URL, so this is no longer needed. I'll remove it, absent any use case for it.

src/main/kotlin/org/jitsi/jicofo/bridge/ExternalBridgeSelectionStrategy.kt
jbg: It's necessary as long as the index is used in the response to identify the bridge. But if the JID is used as suggested, this can even be removed entirely rather than made optional IMO (our service also knows the list of available bridges).

But I was concerned how Jicofo would react if told to use a bridge it doesn't already know about. Even if not deliberate, I can imagine when bridges start up or shut down there could be brief periods where the external selection service has a different idea of available bridges than Jicofo.

PR #894, opened 2022-03-17
feat(ColibriV2): source names capability

paweldomas:
Depends on https://github.com/jitsi/jitsi-xmpp-extensions/pull/68

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -131,7 +131,7 @@
     <dependency>
       <groupId>${project.groupId}</groupId>
       <artifactId>jitsi-xmpp-extensions</artifactId>
-      <version>1.0-44-g707c6af</version>
+      <version>1.0-SNAPSHOT</version>
     </dependency>
     <dependency>
       <groupId>org.slf4j</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -131,7 +131,7 @@
     <dependency>
       <groupId>${project.groupId}</groupId>
       <artifactId>jitsi-xmpp-extensions</artifactId>
-      <version>1.0-SNAPSHOT</version>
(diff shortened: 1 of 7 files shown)


src/main/java/org/jitsi/jicofo/discovery/DiscoveryUtil.java
```diff
@@ -81,6 +81,11 @@
      */
     public final static String FEATURE_DTLS = "urn:xmpp:jingle:apps:dtls:0";
 
+    /**
+     * Source name signalling required for the multi-stream mode.
+     */
+    public final static String FEATURE_SOURCE_NAMES = "'http://jitsi.org/source-name'";
```
bgrozev, line 87: Are the quotes intentional (both `"` and `'`)?

paweldomas: no, but since this was not tested yet, that's why it wouldn't work I guess :) thanks

pom.xml
```diff
@@ -131,7 +131,7 @@
     <dependency>
       <groupId>${project.groupId}</groupId>
       <artifactId>jitsi-xmpp-extensions</artifactId>
-      <version>1.0-44-g707c6af</version>
+      <version>1.0-SNAPSHOT</version>
```
bgrozev, line 134: This can be fixed now

paweldomas: well actually not, for some reason it didn't build yet? https://github.com/jitsi/jitsi-maven-repository/commits/master

paweldomas: checked with Damyan about this

bgrozev (approved):

Issue #895, opened 2022-03-18
Failed to grant owner status: Bare JID expected, got full JID


denzill:
### Description:
I install Jitsi-meet by this doc: https://jitsi.github.io/handbook/docs/devops-guide/devops-guide-quickstar.
Then enable authentication by this: https://jitsi.github.io/handbook/docs/devops-guide/secure-domain.
And now after create conference i get an error in jicofo.log:

`Jicofo 2022-03-18 16:13:23.094 SEVERE: [118] ChatRoomRoleManager.grantOwner#46: Failed to grant owner status to mpyze6sgt1f5rgcabavcgo5j@guest.jitsi.minsk.isida.by/JayE684PsEBe
java.lang.RuntimeException: Failed to grant owner: <iq xmlns='jabber:client' to='focus@auth.jitsi.minsk.isida.by/focus' from='crowdedphilosopherslistenresponsibly@conference.jitsi.minsk.isida.by' id='BQN8J-179' type='error'><error type='modify'><jid-malformed xmlns='urn:ietf:params:xml:ns:xmpp-stanzas'/><text xmlns='urn:ietf:params:xml:ns:xmpp-stanzas'>Bare JID expected, got full JID</text></error></iq>
	at org.jitsi.impl.protocol.xmpp.ChatRoomImpl.grantOwnership(ChatRoomImpl.java:482)
	at org.jitsi.jicofo.xmpp.muc.ChatRoomRoleManager.grantOwner(ChatRoomRoleManager.kt:43)
	at org.jitsi.jicofo.xmpp.muc.AuthenticationRoleManager.authenticationListener$lambda-2(ChatRoomRoleManager.kt:127)
	at org.jitsi.jicofo.auth.AbstractAuthAuthority.notifyUserAuthenticated(AbstractAuthAuthority.java:343)
	at org.jitsi.jicofo.auth.AbstractAuthAuthority.authenticateJidWithSession(AbstractAuthAuthority.java:433)
	at org.jitsi.jicofo.auth.XMPPDomainAuthAuthority.processAuthLocked(XMPPDomainAuthAuthority.java:100)
	at org.jitsi.jicofo.auth.AbstractAuthAuthority.processAuthentication(AbstractAuthAuthority.java:356)
	at org.jitsi.jicofo.xmpp.ConferenceIqHandler.processExtensions(ConferenceIqHandler.kt:118)
[...]

PR #896, opened 2022-03-18
Fix granting ownership

bgrozev:
- ref: Pass a ChatRoomMember to grantOwnership().
- fix(#895): Use bare JID when granting priviledges.

### DIFF
--- a/src/main/java/org/jitsi/impl/protocol/xmpp/ChatRoom.java
+++ b/src/main/java/org/jitsi/impl/protocol/xmpp/ChatRoom.java
@@ -99,10 +99,9 @@ public interface ChatRoom
     * ownership privileges to other users. An owner is allowed to change
     * defining room features as well as perform all administrative functions.
     *
-    * @param address the user address of the user to grant ownership
-    * privileges (e.g. "user@host.org").
+    * @param member the member to grant ownershit to.
     */
-    void grantOwnership(String address);
+    void grantOwnership(@NotNull ChatRoomMember member);
 
     /**
      * Destroys the chat room.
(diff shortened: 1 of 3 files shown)


saghul (dismissed):
LGTM! Q: this change is backwards compatible, right?

src/main/java/org/jitsi/impl/protocol/xmpp/ChatRoom.java
```diff
@@ -99,10 +99,9 @@ void join()
     * ownership privileges to other users. An owner is allowed to change
     * defining room features as well as perform all administrative functions.
     *
-    * @param address the user address of the user to grant ownership
-    * privileges (e.g. "user@host.org").
+    * @param member the member to grant ownershit to.
```
mwild1, line 102: > the member to grant ownershit to.

:speak_no_evil:

mwild1 (commented):
LGTM. I notice a few mentions of "roles". Although I don't see anything problematic, it's worth noting that XEP-0045 distinguishes between affiliations (owner, admin, member) which are persistent and based on the user's real *bare* JID and roles (moderator, participant).

Roles in XEP-0045 are typically temporary permissions (i.e. lost if they leave the chat). Use-cases for roles are rare these days - if you're using them, know that they are based on nicks, not JIDs, which can be confusing.

I'm assuming the term "role" in this code is referring to Jitsi's own concept of a role in the conference, rather than XEP-0045 roles, but just wanted to mention it in case.

damencho (approved):

bgrozev:
> I'm assuming the term "role" in this code is referring to Jitsi's own concept of a role in the conference, rather than XEP-0045 roles, but just wanted to mention it in case.

That's correct, jicofo's concept of a "role" applies just to a member of a room and [combines the XMPP affiliation and role](https://github.com/jitsi/jicofo/blob/07035c8d70dae924ff108fe231819b4cff94122f/src/main/kotlin/org/jitsi/jicofo/xmpp/muc/MemberRole.kt#L50). I'm not even sure XMPP roles are actively used.

PR #897, opened 2022-03-18
feat: Add an option to use presence instead of sending health checks.

bgrozev:
- Read bridge health from presence.
- Save the time presence was last received.
- feat: Add an option to use presence instead of sending health checks.

### DIFF
--- a/src/main/java/org/jitsi/jicofo/bridge/Bridge.java
+++ b/src/main/java/org/jitsi/jicofo/bridge/Bridge.java
@@ -137,6 +137,9 @@ public class Bridge
     @NonNull
     private final Clock clock;
 
+    @NonNull
+    private Instant lastPresenceReceived = Instant.MIN;
+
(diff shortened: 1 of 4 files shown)


aaronkvanmeerten (approved):

src/main/java/org/jitsi/jicofo/bridge/Bridge.java
```diff
@@ -223,6 +243,12 @@ public void setStats(ColibriStatsExtension stats)
         {
             this.relayId = relayId;
         }
+
+        String healthy = stats.getValueAsString("healthy");
+        if (healthy != null)
+        {
+            this.healthy = Boolean.parseBoolean(healthy);
+        }
```
gpolitis, line 251: We don't expect a null here, right? Maybe we could print a warning message, if this happens?

gpolitis (dismissed):

aaronkvanmeerten (approved):

PR #903, opened 2022-04-20
Mute fixes

bgrozev:
- fix: Set forceMute state with the initial colibri allocate request and jingle offer.
- fix: Send a single coolibri message for force-mute, do not block for a response.

Incidentally, this fixes a race condition between terminateParticipant() and muteAllParticipants() which throws an exception when trying to mute a participant that has been removed.

### DIFF
--- a/src/main/java/org/jitsi/jicofo/conference/JitsiMeetConferenceImpl.java
+++ b/src/main/java/org/jitsi/jicofo/conference/JitsiMeetConferenceImpl.java
@@ -1597,15 +1597,9 @@ public class JitsiMeetConferenceImpl
      */
     public void muteParticipant(Participant participant, MediaType mediaType)
     {
-        if (participant.getChatMember().isJigasi() && !participant.hasAudioMuteSupport())
+        if (participant.shouldSuppressForceMute())
         {
-            logger.warn("Will not mute jigasi with not audioMute support: " + participant);
-            return;
-        }
-
-        if (participant.getChatMember().isJibri())
-        {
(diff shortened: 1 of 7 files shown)


damencho (approved):

src/main/java/org/jitsi/jicofo/conference/JitsiMeetConferenceImpl.java
```diff
@@ -1511,25 +1512,17 @@ else if (!participant.hasModeratorRights()
...
-            logger.warn("Mute not allowed, toBeMuted is jigasi.");
-            return MuteResult.NOT_ALLOWED;
-        }
-
-
-        if (doMute && participant.getChatMember().isJibri())
+        if (participant.shouldSuppressForceMute())
         {
-            logger.warn("Mute not allowed, toBeMuted is jibri.");
+            logger.warn("Force mute suppressed, returning NOT_ALLOWED:" + participant);
```
JonathanLennox, line 1517: This case used to only trigger if doMute was true - this will now return NOT_ALLOWED for unmuting a Jigasi too.  Is this what we want?

bgrozev: I think the result is equivalent because "unmute" is only allowed when the request comes from the endpoint itself, and since jigasi doesn't support muting it never tries to unmute itself either (besides, it can't be muted) :)

But this was not intenional and I see a little bit of cleanup I can do. If you think the check makes more sense I'll add it back.

PR #912, opened 2022-04-27
Check client support for ssrc-rewriting functionality and signal it t…

jqdrqgnq:
…o jvb.

### DIFF
--- a/src/main/java/org/jitsi/jicofo/conference/Participant.java
+++ b/src/main/java/org/jitsi/jicofo/conference/Participant.java
@@ -237,6 +237,14 @@ public class Participant
     }
 
     /**
+     * @return {@code true} if this participant supports SSRC rewriting functionality.
+     */
+    public boolean hasSsrcRewritingSupport()
+    {
+        return supportedFeatures.contains(DiscoveryUtil.FEATURE_SSRC_REWRITING);
+    }
+
+    /**
      * @return {@code true} if this participant supports receiving Jingle sources encoded as JSON instead of the
      * standard Jingle encoding.
      */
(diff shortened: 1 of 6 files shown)


src/main/java/org/jitsi/jicofo/conference/ParticipantInviteRunnable.java
```diff
@@ -468,8 +468,16 @@ private boolean doInviteOrReinvite(Jid address, Offer offer, ColibriAllocation c
...
-        // Take all sources from participants in the conference.
-        ConferenceSourceMap conferenceSources = meetConference.getSources().copy();
+        ConferenceSourceMap conferenceSources;
+
+        if (!participant.hasSsrcRewritingSupport()) {
+            // Take all sources from participants in the conference.
+            conferenceSources = meetConference.getSources().copy();
+        } else {
+            // Bridge will signal sources in this case.
+            conferenceSources = new ValidatingConferenceSourceMap();
```
bgrozev, line 478: You can use a regular `ConferenceSourceMap` here, no validation (checking for ssrc conflicts, etc) needed.

JonathanLennox:
I feel like there should be some infra-side configuration about whether we want to do ssrc-rewriting, not just the client capability (unless sending the client capability is triggered by some server config?  I haven't read the client-side PRs.)

I guess this will need to be a jicofo-level configuration option, since that would control whether all the conference sources are forwarded?

JonathanLennox:
This seems good to me, but should wait to land until the JVB part is done.  (We'll need to make sure JVB and Jicofo are updated in sync with each other.)

bgrozev (approved):

bgrozev:
> This seems good to me, but should wait to land until the JVB part is done. (We'll need to make sure JVB and Jicofo are updated in sync with each other.)

Since jicofo's config option defaults to `false` it should be safe to merge. But it makes sense to wait and merge together

JonathanLennox (approved):

bgrozev (approved):

JonathanLennox:
It's getting a failure in the unit tests, do we understand why?

PR #923, opened 2022-05-20
Use larger buckets for conference size stats.

bgrozev:
- fix: Fix @NotNull annotations.
- chore: Update jitsi-utils, use apache commons instead of guava.
- ref: Use larger buckets for conference size stats.


bgrozev:
This is on top of #922

bgrozev:
It uses larger buckets for conference sizes, instead of size-1 buckets from 1 to 20, then a single bucket for 21+ (that's where all the fermented jibris used to be stored. 

Also, the new version of jitsi-utils changed the JSON format so it's cleanly exportable to e.g. wavefront
```
   "conference_sizes": {
        "average": null,
        "max": 0,
        "total_value": 0,
        "total_count": 0,
        "buckets": {
            "0": 0,
            "1": 0,
            "2": 0,
            "3_to_5": 0,
            "6_to_10": 0,
            "11_to_20": 0,
            "21_to_50": 0,
            "51_to_100": 0,
            "101_to_200": 0,
            "201_to_300": 0,
            "301_to_400": 0,
            "401_to_500": 0,
            "501_to_max": 0,
            "p99_upper_bound": -1,
            "p999_upper_bound": -1
        }
    },
```


@aaronkvanmeerten any thughts on the selectin of buckets?

src/main/java/org/jitsi/jicofo/FocusManager.java
```diff
@@ -428,26 +432,21 @@ public JSONObject getStats()
...
         stats.put("largest_conference", largestConferenceSize);
         stats.put("participants", numParticipants);
-        JSONArray conferenceSizesJson = new JSONArray();
-        for (int size : conferenceSizes)
-        {
-            conferenceSizesJson.add(size);
-        }
-        stats.put("conference_sizes", conferenceSizesJson);
+        stats.put("conference_sizes", conferenceSizes.toJson());
+        stats.put("endpoint_pairs", endpointPairs);
```
JonathanLennox, line 449: I'm not sure about "endpoint_pairs" as the name of this.  What's the motivation for this?

bgrozev: To see how this number correlates with load on the machine or the prosody process (in order to potentially use it to set the shard's weight for load balancing)

JonathanLennox (approved):

PR #926, opened 2022-05-26
fix: Reduce the core size of the scheduled pool to 3.

bgrozev:
It is unlimited, so it will grow as needed. There's no need to
keep 200 threads alive at all time.


JonathanLennox (approved):

bbaldino:
Does Jicofo have that much scheduled work?  I think JVB just has a single thread in the scheduled pool.

PR #928, opened 2022-05-26
Turn off Opus inband FEC when RED is in use

nils-ohlmeier:
Currently when RED is being used as FEC we also signal the usage of Opus inband FEC. Running these two FECs algorithms together just wastes bandwidth for little to no gain. Since the results for RED for heavier packet loss is better compared to Opus FEC, this PR turns off Opus FEC when RED is in use.

### DIFF
--- a/src/main/kotlin/org/jitsi/jicofo/codec/CodecUtil.kt
+++ b/src/main/kotlin/org/jitsi/jicofo/codec/CodecUtil.kt
@@ -113,7 +113,6 @@ class CodecUtil {
             options: OfferOptions = OfferOptions()
         ): Collection<PayloadTypePacketExtension> = buildList {
             if (config.opus.enabled()) {
-                var addedRed: Boolean = false
                 // Though RED has a payload type of its own and can be used to encode multiple other payload types, we need
                 // it to be advertised with the same clock rate as opus, so it's defined here.
                 // Add the RED payload type before Opus, so that it is the selected codec.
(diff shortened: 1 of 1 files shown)


src/main/kotlin/org/jitsi/jicofo/codec/CodecUtil.kt
```diff
@@ -145,7 +147,8 @@ class CodecUtil {
                 }
                 if (config.opus.useInbandFec()) {
                     // fmtp:111 useinbandfec=1
-                    opus.addParameterExtension("useinbandfec", "1")
+                    val fecval = addedRed.compareTo(true).toString()
```
bgrozev, line 150: If we've added RED then we add `useinbandfec=false`, but if `opus.useInbandFec` is false we don't add an extensin at all. Is this intentional? If not, you could simplify to just `if (config.opus.useInbandFec() && !config.opus.red.enabled())`

nils-ohlmeier: > If we've added RED then we add `useinbandfec=false`, but if `opus.useInbandFec` is false we don't add an extensin at all. Is this intentional? If not, you could simplify to just `if (config.opus.useInbandFec() && !config.opus.red.enabled())`

Yeah it is kind of intentional. This results in `useinbandfec=0`. I wanted to be explicit about it, rather than rely on absence of the `useinbandfec` parameter to do the right thing.
And I decided to not rely on the config option in case someone later changes the if condition for the RED block.
But if you think it is important to change to your version I'll update my commit.

bgrozev (approved):

PR #935, opened 2022-06-02
feat: Re-invite participants when a bridge goes to SHUTTING_DOWN.

bgrozev:


src/main/java/org/jitsi/jicofo/bridge/JvbDoctor.java
```diff
@@ -129,6 +130,11 @@ public void bridgeAdded(Bridge bridge)
         logger.info("Scheduled health-check task for: " + bridge);
     }
 
+    @Override
+    public void bridgeIsShuttingDown(@NotNull Bridge bridge)
```
JonathanLennox, line 134: Why is this needed? Does the default Kotlin implementation not apply to Java implementations of interfaces?

bgrozev: The kotlin function needs a `@JvmDefault` annotation, but that requires to run with `-Xjvm-default` and I didn't want to do that.

JonathanLennox (approved):

PR #936, opened 2022-06-08
Prefer local region for initial selection

bgrozev:
- feat: Default to localRegion when participantRegion is null.
- test: Add a test for initial selection in region group.
- feat: Prefer local region for initial selection.
- doc: Cleanup docs.

### DIFF
--- a/src/main/kotlin/org/jitsi/jicofo/bridge/BridgeSelector.kt
+++ b/src/main/kotlin/org/jitsi/jicofo/bridge/BridgeSelector.kt
@@ -105,11 +105,9 @@ class BridgeSelector @JvmOverloads constructor(
     }
 
     /**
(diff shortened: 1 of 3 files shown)


src/main/kotlin/org/jitsi/jicofo/bridge/BridgeSelector.kt
```diff
@@ -105,11 +105,9 @@ class BridgeSelector @JvmOverloads constructor(
     }
 
     /**
-     * Removes Jitsi Videobridge XMPP address from the list videobridge
-     * instances available in the system .
+     * Removes Jitsi Videobridge XMPP address from the list videobridge instances available in the system.
```
JonathanLennox, line 108: "list **of** videobridge instances"

damencho (approved):

PR #910, opened 2022-04-22
Support visitor role for endpoints.

JonathanLennox:

PR #940, opened 2022-06-27
WIP: Complex bridge topologies.

JonathanLennox:
This only has the abstract representation of topologies so far.


src/main/kotlin/org/jitsi/jicofo/conference/colibri/v2/Colibri2Session.kt
```diff
@@ -48,18 +49,18 @@ internal class Colibri2Session(
...
     private val xmppConnection = colibriSessionManager.xmppConnection
     val id = UUID.randomUUID().toString()
     /**
      * Save the relay ID locally since it is possible for the relay ID of the Bridge to change and we don't want it to
-     * change in the context of a session. We maintain the invariant that whenever a a conference has multiple sessions,
+     * change in the context of a session. We maintain the invariant that whenever a conference has multiple sessions,
      * they all have non-null relay IDs.
      */
-    val relayId: String? = bridge.relayId
+    override val relayId: String? = bridge.relayId
```
bgrozev, line 63: We could use `bridge.jid` as the relay instead (which is already non-nullable)

src/main/kotlin/org/jitsi/jicofo/cascade/Cascade.kt
```diff
@@ -137,26 +137,40 @@ fun <N : CascadeNode<N, L>, L : CascadeLink> Cascade<N, L>.removeNode(
...
+        "$from does not have a link to $toward"
+    }
+    getNodesBehind(link, toward, nodes)
+    return nodes
+}
+
+fun <N : CascadeNode<N, L>, L : CascadeLink> Cascade<N, L>.getNodesBehind(fromMesh: String, toward: N): Set<N> {
+    val nodes = HashSet<N>()
+    nodes.add(toward)
+    toward.relays.values.forEach {
```
bgrozev, line 153: nit: Maybe `filter` would be cleaner than `return@forEach`

src/main/kotlin/org/jitsi/jicofo/conference/colibri/v2/ColibriV2SessionManager.kt
```diff
@@ -244,8 +245,12 @@ class ColibriV2SessionManager(
...
     }
 
-    override fun addLinkBetween(node: Colibri2Session, otherNode: Colibri2Session, meshId: String) {
-        TODO("Not yet implemented")
+    override fun addLinkBetween(session: Colibri2Session, otherSession: Colibri2Session, meshId: String) {
+        val participantsBehindSession = getNodesBehind(meshId, session).flatMap { getSessionParticipants(it) }
+        val participantsBehindOtherSession = getNodesBehind(meshId, otherSession).flatMap { getSessionParticipants(it) }
+
+        session.createRelay(otherSession.relayId!!, participantsBehindOtherSession, initiator = true)
+        otherSession.createRelay(session.relayId!!, participantsBehindSession, initiator = false)
```
bgrozev, line 261: That's just cool :)

src/main/kotlin/org/jitsi/jicofo/conference/colibri/v2/Colibri2Session.kt
JonathanLennox: This would be a change from what we do today, yes? And thus potentially more disruptive of code I'm already feeling like I have to be fairly delicate about...

JonathanLennox:
This work was finished in https://github.com/jitsi/jicofo/pull/960 because the logic to actually _do_ something with the topologies needed the visitors work from https://github.com/jitsi/jicofo/pull/910.

PR #952, opened 2022-07-27
fix: Clear old state when MUC is re-joined.

bgrozev:


src/main/java/org/jitsi/impl/protocol/xmpp/ChatRoomImpl.java
```diff
@@ -225,6 +226,30 @@ public String getMainRoom() {
         return mainRoom;
     }
 
+    /**
+     * Prepare this {@link ChatRoomImpl} for a call to {@link #joinAs(Resourcepart)}, which send initial presence to
+     * the MUC. Resets any state that might have been set the previous time the MUC was joined.
+     */
+    private void resetState()
```
paweldomas, line 233: I thought that ChatRoomImpl instances are destroyed and a new one would be created? Is there a rejoin scenario?

bgrozev: They are cached and potentially reused in `XmppProtocolProvider.Muc`. I'm not sure it's necessary, but I don't understand it well enough to change right now.

This fix is based on a dump where a chat room was re-joined and had an old entry in `members` which was never removed as it is not an occupant of the Smack MUC.

src/main/java/org/jitsi/impl/protocol/xmpp/ChatRoomImpl.java
```diff
@@ -225,6 +233,33 @@ public String getMainRoom() {
...
+            lastPresenceSent = null;
+            meetingId = null;
+            logger.addContext("meeting_id", "");
+            isBreakoutRoom = false;
+            mainRoom = null;
+            avModerationEnabled.clear();
+            whitelists.clear();
+        }
+    }
     private void joinAs(Resourcepart nickname) throws SmackException, XMPPException, InterruptedException
```
JonathanLennox, line 263: Blank line between methods

src/main/java/org/jitsi/impl/protocol/xmpp/ChatRoomImpl.java
```diff
@@ -380,7 +415,7 @@ private void resetRoleForOccupant(EntityFullJid occupantJid)
         }
         else
         {
-            ChatMemberImpl member = members.get(occupantJid);
+            ChatMemberImpl member = getChatMember(occupantJid);
```
JonathanLennox, line 418: Does this need this change? It seems like the only difference between the two methods is what happens if `occupantJid` is null, which already can't be true?

bgrozev: I changed it because I was trying to fix spotbugs "inconsistent synchronization" warnings, but it's not needed. I reverted it.

JonathanLennox (approved):

PR #956, opened 2022-08-22
Split a jicofo-selector maven artefact.

bgrozev:
- ref: Encapsulate allocate() params in a class, do not use Participant.
- ref: Move the sctp flag to ParticipantAllocationOptions.
- ref: Move jingle->colibri conversion out of colibri.
- ref: Pass properties to ColibriV2SessionManager.
- ref: Do not pass "conference" to ColibriV2SessionManager.
- ref: Move everything to a mvn module.
- ref: Split maven artifacts: jicofo-common, jicofo-selector, jicofo.


src/main/kotlin/org/jitsi/jicofo/conference/colibri/ColibriSessionManager.kt
```diff
@@ -90,3 +87,13 @@ interface ColibriSessionManager {
         )
     }
 }
+
+data class ParticipantAllocationOptions(
```
JonathanLennox, line 88: I might call this `ParticipantAllocationParameters` rather than `ParticipantAllocationOptions`, since it includes things like the various IDs that aren't optional.

src/main/kotlin/org/jitsi/jicofo/conference/colibri/v2/ParticipantInfo.kt
```diff
@@ -32,9 +33,19 @@ internal class ParticipantInfo(
     var audioMuted: Boolean = false,
     /** Whether the participant is force video muted. */
     var videoMuted: Boolean = false,
-    /* Whether the participant supports source name signaling */
+    /** Whether the participant supports source name signaling */
```
JonathanLennox, line 38: Do we still need the long-form constructor here?

src/main/kotlin/org/jitsi/jicofo/conference/Util.kt
```diff
@@ -0,0 +1,46 @@
...
+ * limitations under the License.
+ */
+package org.jitsi.jicofo.conference
+
+import org.jitsi.utils.MediaType
+import org.jitsi.xmpp.extensions.colibri2.Media
+import org.jitsi.xmpp.extensions.jingle.ContentPacketExtension
+import org.jitsi.xmpp.extensions.jingle.RtpDescriptionPacketExtension
+
+class Util {
```
JonathanLennox: Is there a better name for this class?

jicofo-selector/src/main/java/org/jitsi/jicofo/bridge/JvbDoctor.java
```diff
@@ -174,7 +174,7 @@ protected void doHealthCheck()
 
             logger.debug("Sending health-check request to: " + bridge);
 
-            IQ response = UtilKt.sendIqAndGetResponse(connection, newHealthCheckIQ(bridge));
+            IQ response = org.jitsi.jicofo.xmpp.UtilKt.sendIqAndGetResponse(connection, newHealthCheckIQ(bridge));
```
JonathanLennox, line 177: Why does this need to be spelled out?  Do we have multiple `UtilKt` classes floating around now?

bgrozev: Yes, we have one in jicofo-common and one in jicofo-selector

JonathanLennox (approved):

PR #961, opened 2022-09-06
Send Jingle source parameters with the correct namespace

jbg:
Ref: jitsi-xmpp-extensions#82

This shouldn't be merged until jitsi-xmpp-extensions is updated.


linkmauve:
Hi, if I were to rebase this branch, would you merge it eventually?  This would make it easier to support in other implementations.

PR #967, opened 2022-09-16
Add the msid for feedback sources internally.

bgrozev:


bgrozev:
Moving this logic from jvb to jicofo because it's just a pculiarity of what the client expects (and this way we can avoid using ParameterPacketExtensin in colibri2)

JonathanLennox (approved):

PR #969, opened 2022-09-19
Bridge selection logs and metric tweaks

bgrozev:
- log: Log selection at INFO.
- Add metric for the number of different bridge versions available.


bgrozev:
The only failure is the log verification, which is expected. I opened https://github.com/8x8Cloud/jitsi-pr-testing-scripts/pull/108 to adapt the scripts

PR #978, opened 2022-09-26
fix: Fix leaking conferences.

bgrozev:
I don't understand how the leak occurs, but I see multiple conferences
left as listeners of XmppProviderImpl, which are not in FocusManager
anymore. Fire events via FocusManager instead of registering individual
conferences.


JonathanLennox:
I take it this doesn't need to be deregistered because the focus manager is an eternal singleton?

bgrozev:
> I take it this doesn't need to be deregistered because the focus manager is an eternal singleton?

Ops, I added the missing removeRegistrationListener it and added a call to FocusManager.stop() that was missing

JonathanLennox (approved):

PR #986, opened 2022-10-04
Move XMPP config to jicofo.conf

bgrozev:
This needs more testing. But the idea is that command line arguments are no longer used. All options that used command line arguments are now set in jicofo.conf instead, and `postinst` takes care of moving things from `/etc/jitsi/jicofo/config` to `jicofo.conf`.

- debian: Only pass command line params if the options are defined.
- fix: Read xmpp-domain from jicofo.conf.
- debian: Move xmpp config from "config" to jicofo.conf.
- debian: Do not pass /etc/jitsi/jicofo/config values as command line arguments.
- ref: Do not read command line arguments (use jicofo.conf instead).
- Add docs to reference.conf.

### DIFF
--- a/debian/postinst
+++ b/debian/postinst
@@ -84,10 +84,18 @@ jicofo {
 
         # Set the client XMPP config in jicofo.conf. The values were either read from an existing
         # /etc/jitsi/jicofo/config or generated above.
-        hocon -f $HOCON_CONFIG set "jicofo.xmpp.client.xmpp-domain" "$JICOFO_HOSTNAME"
-        hocon -f $HOCON_CONFIG set "jicofo.xmpp.client.domain" "$JICOFO_AUTH_DOMAIN"
-        hocon -f $HOCON_CONFIG set "jicofo.xmpp.client.username" "$JICOFO_AUTH_USER"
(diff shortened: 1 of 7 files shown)


debian/postinst
```diff
@@ -101,6 +82,13 @@ jicofo {
         # Make sure the client-proxy is set correctly.
         hocon -f $HOCON_CONFIG set jicofo.xmpp.client.client-proxy "focus.$JVB_HOSTNAME"
 
+        # Set the client XMPP config in jicofo.conf. The values were either read from an existing
+        # /etc/jitsi/jicofo/config or generated above.
+        hocon -f $HOCON_CONFIG set "jicofo.xmpp.client.xmpp-domain" "$JICOFO_HOSTNAME"
```
damencho, line 87: If you want to migrate config, this will not work. This `if` is when there was no old config or we are reconfiguring jicofo because the domain changed. 
And if you drop the handling in the code, those that are updating will stop working for them.

bgrozev: It's actually outside the `if`

damencho (approved):

PR #989, opened 2022-10-05
Fix stopping jibri sessions

bgrozev:
- fix: Set JibriRecorder.jibriSession prioor to running start().
- fix: Run JibriSession.stop() in an IO thread as it might block.

### DIFF
--- a/jicofo/src/main/kotlin/org/jitsi/jicofo/jibri/JibriRecorder.kt
+++ b/jicofo/src/main/kotlin/org/jitsi/jicofo/jibri/JibriRecorder.kt
@@ -18,6 +18,7 @@
 package org.jitsi.jicofo.jibri
 
 import org.apache.commons.lang3.StringUtils
+import org.jitsi.jicofo.TaskPools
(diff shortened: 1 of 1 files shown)


jicofo/src/main/kotlin/org/jitsi/jicofo/jibri/JibriRecorder.kt
```diff
@@ -108,10 +108,10 @@ class JibriRecorder(
                     jibriDetector,
                     false, null, iq.displayName, iq.streamId, iq.youtubeBroadcastId, sessionId, iq.appData,
                     logger
-                ).apply {
-                    start()
-                    logger.info("Started Jibri session")
-                }
+                )
+                this.jibriSession = jibriSession
```
JonathanLennox, line 115: I see why you removed the `apply`, but why do you need the `val`?

bgrozev: Otherwise I have to deal with `this.jibriSession` potentially being null

PR #990, opened 2022-10-06
log: Do not log expiration of health check conferences.

bgrozev:


JonathanLennox (approved):
Possibly `includeInStatistics` should be changed to be named something more general given this change, but this is fine.

PR #993, opened 2022-10-07
Add ParticipantProperties to bridge selector

JonathanLennox:


jicofo-selector/src/main/kotlin/org/jitsi/jicofo/bridge/colibri/ColibriV2SessionManager.kt
```diff
@@ -250,11 +251,14 @@ class ColibriV2SessionManager(
...
             // the requests were sent.
-            val bridge = bridgeSelector.selectBridge(getBridges(), participant.region, bridgeVersion)
-                ?: run {
-                    eventEmitter.fireEvent { bridgeSelectionFailed() }
-                    throw BridgeSelectionFailedException()
-                }
+            val bridge = bridgeSelector.selectBridge(
+                getBridges(),
+                ParticipantProperties(participant.region),
+                bridgeVersion
```
bgrozev, line 257: Should `bridgeVersion` go to `ConferenceBridgeProperties` now?

JonathanLennox: It's not a bad idea but ConferenceBridgeProperties was the previous PR. :slightly_smiling_face:

So probably that should be a separate PR if we want to do it. I can look at doing it next week.

bgrozev (approved):

PR #1005, opened 2022-11-16
ref(endpoint-allocation): add conference expiration and re-request allocation upon already existed conference error

mykhailo-ua:
Added conference expiration and then re-request allocation for endpoint when allocation is failed with error of already existed conference which can happen after jicofo restart.

Hi guys, added the code which should handle the case when jicofo after restart tries to re-create endpoints and have error from jvb that such conferences already existed which prevent jicofo to restore the existed before conferences.

### DIFF
--- a/jicofo-selector/src/main/kotlin/org/jitsi/jicofo/bridge/colibri/ColibriV2SessionManager.kt
+++ b/jicofo-selector/src/main/kotlin/org/jitsi/jicofo/bridge/colibri/ColibriV2SessionManager.kt
@@ -311,7 +311,12 @@ class ColibriV2SessionManager(
             try {
                 return handleResponse(response, session, created, participantInfo)
             } catch (e: Exception) {
-                logger.error("Failed to allocate a colibri2 endpoint for ${participantInfo.id}: ${e.message}")
+                if (e is ConferenceAlreadyExistsException) {
+                    logger.warn("Failed to allocate a colibri2 endpoint for ${participantInfo.id}: ${e.message}")
+                } else {
+                    logger.error("Failed to allocate a colibri2 endpoint for ${participantInfo.id}: ${e.message}")
+                }
+
                 if (e is ColibriAllocationFailedException && e.removeBridge) {
                     // Add participantInfo just in case it wasn't there already (the set will take care of dups).
                     val removedParticipants = removeSession(session) + participantInfo
@@ -386,6 +391,10 @@ class ColibriV2SessionManager(
                             "XMPP error: ${response.error?.toXML()}",
                             true
                         )
+                    } else if (reason == Colibri2Error.Reason.CONFERENCE_ALREADY_EXISTS) {
+                        // The conference on the bridge already exists. The state between jicofo and the bridge
+                        // is out of sync.
+                        throw ConferenceAlreadyExistsException("Conference already exists error", true)
                     } else {
                         // An error coming from the bridge. The state between jicofo and the bridge must be out of sync.
                         // It's not clear how to handle this. Ideally we should expire the conference and retry, but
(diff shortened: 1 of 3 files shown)


jicofo-selector/src/main/kotlin/org/jitsi/jicofo/bridge/colibri/ColibriV2SessionManager.kt
```diff
@@ -371,6 +395,10 @@ class ColibriV2SessionManager(
                         // The conference on the bridge has expired. The state between jicofo and the bridge is out of
                         // sync.
                         throw ColibriAllocationFailedException("Conference not found", true)
+                    } else if (reason == Colibri2Error.Reason.CONFERENCE_ALREADY_EXISTS) {
```
bgrozev, line 398: I don't think we'd see `condition = item_not_found` and `reason = CONFERENCE_ALREADY_EXISTS`. When a colibri conference exists the bridge should respond with `condition = conflict` (and `reason = CONFERENCE_ALREADY_EXISTS`)

jicofo-selector/src/main/kotlin/org/jitsi/jicofo/bridge/colibri/ColibriV2SessionManager.kt
```diff
@@ -325,6 +322,33 @@ class ColibriV2SessionManager(
         }
     }
 
+    private fun expireAndAllocate(session: Colibri2Session, participantInfo: ParticipantInfo): ColibriAllocation {
+        return try {
+            session.expire()
+            session.created = false
+
+            val stanzaCollector = session.sendAllocationRequest(participantInfo)
```
bgrozev, line 330: I am wary of having a different codepath for this. Can you call `expire` and then return an error so that the allocation is re-tried?

jicofo-selector/src/main/kotlin/org/jitsi/jicofo/bridge/colibri/Exceptions.kt
```diff
@@ -18,6 +18,8 @@ package org.jitsi.jicofo.bridge.colibri
 /** Bridge selection failed, i.e. there were no bridges available. */
 class BridgeSelectionFailedException : Exception("Bridge selection failed")
 
+class ConferenceAlreadyExistsException : Exception()
```
bgrozev, line 21: Should this extend `ColibriAllocationFailedException`?

bgrozev (changes requested):

jicofo-selector/src/main/kotlin/org/jitsi/jicofo/bridge/colibri/Exceptions.kt
mykhailo-ua: Right, make it as extend  for ColibriAllocationFailedException

jicofo-selector/src/main/kotlin/org/jitsi/jicofo/bridge/colibri/ColibriV2SessionManager.kt
mykhailo-ua: You are right about it, sorry for that. Moved it to the proper place.

jicofo-selector/src/main/kotlin/org/jitsi/jicofo/bridge/colibri/ColibriV2SessionManager.kt
mykhailo-ua: I made more local testing and realised that expiration is already part `removeSession(session)` call, so there is no need to add additional codepath. I reverted the change but changed the log level to WARN for that situation as it eventually can restore the conference.

bgrozev:
jenkins add to whitelist

mykhailo-ua:
At the end the approach was changed because after more local testing I realised that expiration is already part of removeSession(session) call, so there is no need to add additional expiration and alternative codepath. I reverted the original change but changed the log level to WARN for that situation as it eventually can restore the conference.

PR #1006, opened 2022-11-17
Refactor Jingle code

bgrozev:
- ref: Remove LipSyncHack.
- ref: Check for session only once.
- ref: Move JingleSession to kotlin.
- ref: Move JingleRequestHandler to kotlin.
- ref: Move logic to JingleSession.
- ref: Remove MockParticipant.jingleHandler.
- Keep only WeakReference to JingleSessions.
- ref: Remove terminateHandlersSessions.
- ref: Move terminateSession to JingleSession.
- ref: Rename OperationSetJingle to JingleApi, remove the interface.
- ref: Move JingleApi to the jingle package.
- ref: Access Jingle through Participant->ChatMember.
- ref: Use Metrics for jingle stats.
- ref: Move replaceTransport to JingleSession.
- fix: Treat timeouts as errors.
- ref: Move removeSource to JingleSession.
- ref: Move addSource to JingleSession.
- ref: Move encodeSourcesAsJson to JingleSession.
- ref: Cleanup JingleSession creation and initiation.
- ref: Move encodeSourcesAsJson to JingleSession.
- fix: Treat session-initiate timeout as an error.
- ref: Move encodeSources to JingleSession.
- ref: Remove ourJID.
- ref: Rename JingleSession fields.
- ref: Rename JingleApi to JingleIqRequestHandler.
- ref: Move JingleIqRequestHandler to kotlin.
- ref: Remove connection from JingleIqRequestHandler.
- fix: Return non-null.


jicofo-common/src/main/kotlin/org/jitsi/jicofo/xmpp/jingle/JingleRequestHandler.kt
```diff
@@ -0,0 +1,89 @@
...
+import org.jitsi.xmpp.extensions.jingle.ContentPacketExtension
+import org.jitsi.xmpp.extensions.jingle.JingleIQ
+import org.jivesoftware.smack.packet.StanzaError
+
+/**
+ * Listener class notified about Jingle requests received during a session.
+ *
+ * @author Pawel Domas
+ */
+abstract class JingleRequestHandler {
```
JonathanLennox, line 29: Is there a reason you turned this into an abstract class rather than keeping it an interface?

jicofo-common/src/main/java/org/jitsi/protocol/xmpp/AbstractOperationSetJingle.java
```diff
@@ -302,68 +310,6 @@ private List<ContentPacketExtension> encodeSources(
...
-        case REMOVESOURCE:
-        case SOURCEREMOVE:
-            error = requestHandler.onRemoveSource(session, iq.getContentList());
-            break;
-        default:
-            error = StanzaError.getBuilder(StanzaError.Condition.feature_not_implemented).build();
-            logger.warn("unsupported action " + action);
-        }
-
-        // FIXME IQ type is not taken into account
```
JonathanLennox, line 374: Is this fixme now fixed with the logic rearrangement?

bgrozev: I don't understand what this comment refers to. The instance type is checked to be a JingleIQ. The IQ type is always "set" because that's the type of the AbstractIqRequestHandler instance.

jicofo-common/src/main/kotlin/org/jitsi/jicofo/xmpp/jingle/JingleRequestHandler.kt
bgrozev: Java and kotlin quirks. I removed a whole dummy implementation of this in java and wanted to use default methods, but to make kotlin defaults usable in java you need `@JvmDefault` which recquires a JVM param. Now that I think about it I can define a dummy impl in kotlin, so I'll change it back to an interface. I think we're run into this before.

jicofo-common/src/main/kotlin/org/jitsi/jicofo/xmpp/jingle/JingleRequestHandler.kt
```diff
@@ -0,0 +1,93 @@
...
+    ): StanzaError? = null
+
+    /**
+     * A 'transport-reject' IQ was received.
+     */
+    fun onTransportReject(jingleSession: JingleSession, iq: JingleIQ) { }
+}
+
+/** Export a default impl so it can be used in java without -Xjvm-default */
+class NoOpeJingleRequestHandler : JingleRequestHandler
```
JonathanLennox, line 93: Should be "NoOp" presumably

JonathanLennox (approved):

PR #1014, opened 2022-12-02
Initial implementation of visitor mucs.

bgrozev:
- Add config for visitors XMPP connections.
- Initial implementation of visitor mucs.


bgrozev:
Will re-open from a different branch

PR #1021, opened 2022-12-08
fix: Fix JingleSession handling.

bgrozev:
Participant.jingleSession is now set when we initially create the
JingleSession instance and send a session-initiate IQ. Whether the
session is active is kept track of using JingleSession.state.

### DIFF
--- a/jicofo-common/src/main/kotlin/org/jitsi/jicofo/xmpp/jingle/JingleSession.kt
+++ b/jicofo-common/src/main/kotlin/org/jitsi/jicofo/xmpp/jingle/JingleSession.kt
@@ -109,19 +109,25 @@ class JingleSession(
         sendIq: Boolean
     ) {
         logger.info("Terminating session with $remoteJid, reason=$reason, sendIq=$sendIq")
-        if (state != State.ACTIVE) logger.error("Sending session-terminate for session in state $state")
+        val oldState = state
         state = State.ENDED
 
+        if (oldState == State.ENDED) logger.warn("Terminating session which is already in state ENDED")
+
         if (sendIq) {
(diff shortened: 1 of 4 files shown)


jicofo-common/src/main/kotlin/org/jitsi/jicofo/xmpp/jingle/JingleSession.kt
```diff
@@ -88,11 +105,14 @@ class JingleSession(
     fun terminate(
         reason: Reason,
         message: String?,
-        sendTerminate: Boolean
+        /** Whether to send a session-terminate IQ, or only terminate the session locally. */
+        sendIq: Boolean
     ) {
-        logger.info("Terminating session with $remoteJid, reason=$reason, sendTerminate=$sendTerminate")
+        logger.info("Terminating session with $remoteJid, reason=$reason, sendIq=$sendIq")
+        if (state != State.ACTIVE) logger.error("Sending session-terminate for session in state $state")
```
JonathanLennox, line 112: Can this happen?  Do you definitely want to send session-terminate if you hit this, rather than bailing out?

bgrozev: I think it can happen with a session in PENDING (i.e. session-initiate sent, session-accept not received (yet)) and I think in this case it makes sense to send it. I'm not sure what the previous behavior was, it would probably not send anything as it didn't have a reference to the session. 

I cleaned up the log messages and suppressed sending the IQ if the session is already ENDED. WDYT?

JonathanLennox (approved):

PR #1023, opened 2022-12-13
feat: Keep track of participant features better, expose stats.

bgrozev:


jicofo-common/src/main/kotlin/org/jitsi/jicofo/xmpp/XmppCapsStats.kt
```diff
@@ -0,0 +1,59 @@
...
+                } else {
+                    map[nodeVer]?.let {
+                        it.count++
+                    } ?: logger.warn("Too many entries. Ignoring nodeVer=$nodeVer, features=$features")
+                }
+            }
+        }
+    }
+
+    private class E(val features: List<String>) {
```
JonathanLennox, line 52: Can this get a better name?

JonathanLennox:
Is the assumption that every node with a given capsNodeVer will have identical features?

JonathanLennox (approved):

PR #1024, opened 2022-12-27
feat: cname is needed for better A/V synchronization

cinques:
https://datatracker.ietf.org/doc/html/draft-ietf-rtcweb-rtp-usage-00#section-4.9-2

https://datatracker.ietf.org/doc/html/rfc7022#section-4.1

Right now it is not possible to correlate the audio and video tracks of the remote participant from sdp

### DIFF
--- a/jicofo-common/src/main/kotlin/org/jitsi/jicofo/conference/source/Source.kt
+++ b/jicofo-common/src/main/kotlin/org/jitsi/jicofo/conference/source/Source.kt
@@ -33,6 +33,8 @@ data class Source(
     val name: String? = null,
     /** Optional msid */
     val msid: String? = null,
+    /** Optional cname */
+    val cname: String? = null,
     /** Optional video type (camera or desktop) */
     val videoType: VideoType? = null
 ) {
@@ -43,6 +45,8 @@ data class Source(
         sourcePacketExtension.name,
         sourcePacketExtension.getChildExtensionsOfType(ParameterPacketExtension::class.java)
             .filter { it.name == "msid" }.map { it.value }.firstOrNull(),
+        sourcePacketExtension.getChildExtensionsOfType(ParameterPacketExtension::class.java)
+            .filter { it.name == "cname" }.map { it.value }.firstOrNull(),
         if (sourcePacketExtension.videoType == null) null else VideoType.parseString(sourcePacketExtension.videoType)
     )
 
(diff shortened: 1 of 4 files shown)


cinques:
@saghul could you review it, please?

saghul:
Sorry, I'm not familiar enough with this code to be able to review.

Please allow some time for this to be reviewed. People are trying to take a break until the new year.

bgrozev:
We'd like to make it as easy as possible for people to experiment with sync, but we want to keep the feature off by default (we'd have to do some testing, there were issues in the past). Can you add a config flag?

Is this sufficient for a receiver to synchronize the audio/video stream from the same sender? Do jitsi-meet senders include cname in their signaling?

alexn-vowel:
Please note that Chromium WEBRTC uses MSID rather than CNAME for A/V syncing. The relevant part is here https://source.chromium.org/chromium/chromium/src/+/main:third_party/webrtc/media/engine/webrtc_voice_engine.cc;drc=12033b464df8254c8fd243c488bf3da4a642e17e;l=1918

cinques:
> Do jitsi-meet senders include cname in their signaling?

Do you mean client-side lib? Need a few fixes for that. There is no problem

> Is this sufficient for a receiver to synchronize the audio/video stream from the same sender?

Probably no, MSID is different for video and audio tracks from same sender. So, I think, WebRTC can't match ids to sync A/V.
For example:
`a=msid:4b954633-audio-0-1 b113da94-ba07-434a-972b-049c2732afbc-1`
`a=msid:4b954633-video-0-1 7ff0616a-fa91-47ea-b14a-0658f3ccc10d-1`

bgrozev:
> Probably no, MSID is different for video and audio tracks from same remote participant. So, I think, WebRTC can't match ids to sync A/V.
> For example:
> `a=msid:4b954633-audio-0-1 b113da94-ba07-434a-972b-049c2732afbc-1`
> `a=msid:4b954633-video-0-1 7ff0616a-fa91-47ea-b14a-0658f3ccc10d-1`

That's what I thought. We had some code in jicofo to spoof msid to the same value (look for LipSyncHack in git history), but I recently removed it because it was getting in the way and hadn't been used in a long time. The last time we tried to use it, years ag, we ran into problems on the client side because the audio would not play until a video frame was ready to be rendered, and we don't always have a video frame (e.g. if the sender has no video, or video is suspended due to insuffiucient bandwidth). The LipSyncHack functionality shouldn't be too hard to replicate on top of master if you want to play around with it.

I wouldn't want to include `cname` support without a good reason, and as it is it doesn't accomplish anything on its own (I'm not even clear if it's required when the MSIDs match).

PR #1028, opened 2023-01-05
doc: Add a doc for conference-request

bgrozev:


JonathanLennox:
The vnode part could probably use more documentation, but maybe that should be part of a general visitor mode document?  Otherwise it seems good.

bgrozev:
Yes, I think we need a separate document to describe the visitor scheme

JonathanLennox (approved):

PR #1038, opened 2023-01-27
Send extmap-allow-mixed in Jingle and Colibri2, if configured.

JonathanLennox:
Depends on https://github.com/jitsi/jitsi-xmpp-extensions/pull/90.


JonathanLennox:
That's just copying the existing RtpDescription, yes?  Or creating an empty one that doesn't have any codecs or header extensions specified.  I don't think it makes sense to add extmap-allow-mixed in places that don't have header extensions specified.

bgrozev (approved):

bgrozev (approved):

PR #1042, opened 2023-02-06
Define constants for more features.

bgrozev:


JonathanLennox (approved):
No instances of "Unrecognized" in the PR-tester's jicofo.log, so looks good.

PR #1045, opened 2023-02-13
feat: Return ready=false when the MUC is not joined (regardless of whether authentication is enabled).

bgrozev:
- feat: Return ready=false when the MUC is not joined (regardless of whether authentication is enabled).
- ref: Remove isFocusAnonymous (unused now).


damencho (approved):

bgrozev:
The test failure is unrelated and fixed in #1046

Issue #1047, opened 2023-02-15
Whole bridge terminates upon Unknown endpoint jvb error


mykhailo-ua:
*This Issue tracker is only for reporting bugs and tracking code related issues.*

Before posting, please make sure you check community.jitsi.org to see if the same or similar bugs have already been discussed. General questions, installation help, and feature requests can also be posted to community.jitsi.org.

## Description
When a participant has network problems and his endpoint is expired on jvb, when he is back online and send any update (which happened while he was offline) then jicofo receives Unknown endpoint error from jvb and invalidate whole bridge and then re-invite all the participants.
---
## Current behavior
When participant's endpoint is expired on jvb and he is not kicked out from the conference by timeout then with sending any update (which happened while he was offline) jicofo would receive Unknown endpoint error from jvb and invalidate whole bridge and then re-invite all the participants.
---
## Expected Behavior
Only the invalid endpoint is removed from the bridge and participant re-invited.
---
## Possible Solution
To not invalidate the whole bridge but rather endpoint when processing [sendRequest iq](https://github.com/jitsi/jicofo/blob/0e041934124190cdb5a9950df4807d68a23e69d4/jicofo-selector/src/main/kotlin/org/jitsi/jicofo/bridge/colibri/Colibri2Session.kt#L290) for [updateParticipant](https://github.com/jitsi/jicofo/blob/0e041934124190cdb5a9950df4807d68a23e69d4/jicofo-selector/src/main/kotlin/org/jitsi/jicofo/bridge/colibri/Colibri2Session.kt#L142) and jvb triggers [Unknown endpoint error](https://github.com/jitsi/jitsi-videobridge/blob/master/jvb/src/main/kotlin/org/jitsi/videobridge/colibri2/Colibri2ConferenceHandler.kt#L199)
---
## Steps to reproduce
- start the conference from any device and use it to keep the meeting alive. (used mobile phone)
- join the same meeting with another participant from another device, but make sure that both microphone and camera are off when joining. (used laptop)
- Disconnect the wifi on the laptop for 80 seconds
- Before switching back wifi on laptop turn the microphone on.
- Check the JS console on laptop, it should contain a message about “focus disconnect reason=moving”
![Screenshot 2023-02-15 at 17 50 04](https://user-images.githubusercontent.com/104766779/219102663-d6acd514-dcea-4339-9903-8b41a284659e.png)
---
# Environment details
https://meet.jit.si/
---

bgrozev:
[...]

PR #1048, opened 2023-02-23
ref(colibri2 session participant re-invite): re-invite user on unknown endpoint error

mykhailo-ua:
Added functionality to re-invite participant when UNKNOWN_ENDPOINT Colibri2Error happened instead of recreating the whole bridge and re-invite all participants. It can happen when endpoint is expired on jvb but jicofo tries to send update about this endpoint and gets error back.
Added this change as part of https://github.com/jitsi/jicofo/issues/1047

### DIFF
--- a/jicofo-selector/src/main/kotlin/org/jitsi/jicofo/bridge/colibri/Colibri2Session.kt
+++ b/jicofo-selector/src/main/kotlin/org/jitsi/jicofo/bridge/colibri/Colibri2Session.kt
@@ -30,6 +30,7 @@ import org.jitsi.utils.logging2.Logger
 import org.jitsi.utils.logging2.createChildLogger
 import org.jitsi.xmpp.extensions.colibri.WebSocketPacketExtension
 import org.jitsi.xmpp.extensions.colibri2.Colibri2Endpoint
+import org.jitsi.xmpp.extensions.colibri2.Colibri2Error
 import org.jitsi.xmpp.extensions.colibri2.Colibri2Relay
 import org.jitsi.xmpp.extensions.colibri2.ConferenceModifiedIQ
 import org.jitsi.xmpp.extensions.colibri2.ConferenceModifyIQ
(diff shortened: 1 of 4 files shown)


bgrozev:
Hey @mykhailo-ua,

This looks good to me. I reverted some of the changes in JitsiMeetConferenceImpl, because if we use `updateParticipant = false` we will fail to send `source-remove` to the other participants in the conference. With my change, we will log "No ParticipantInfo for $participantId" from `ColibriV2SessionManager`, but I think that's OK.

Let me know what you think, and if you're able to run any additional tests with the current state of the two remaining PRs that would be useful.

bgrozev:
Jenkins add to whitelist

JonathanLennox (approved):

mykhailo-ua:
Hey guys, I tested the latest changes for jicofo and jvb (https://github.com/jitsi/jitsi-videobridge/pull/1999), which is working for us as well, but during re-inviting participant we have  logged the jicofo error `No ParticipantInfo for XXX` which we would like to avoid. 
There is no point sending update message/or logging error if we already know that jvb endpoint is not present [error_creation_link](https://github.com/jitsi/jicofo/blob/eab721e557520bf47c2362c6618831ec499bb33d/jicofo-selector/src/main/kotlin/org/jitsi/jicofo/bridge/colibri/ColibriV2SessionManager.kt#L531)
Tis error is generated as part of removing participant source [link](https://github.com/jitsi/jicofo/blob/master/jicofo/src/main/java/org/jitsi/jicofo/conference/JitsiMeetConferenceImpl.java#L1196) from re-inviting participants method [link](https://github.com/jitsi/jicofo/blob/master/jicofo/src/main/java/org/jitsi/jicofo/conference/JitsiMeetConferenceImpl.java#L1670) 
My original commit contains this change to eliminate logging the error (pic below), but probably there would be more elegant way how to do it. 
Screenshot 2023-03-23 at 17 39 36

bgrozev:
See my comment above: the original commit introduces a problem because `sendSourceRemove` is not called for the participant being re-invited. I thought the extra code and complexity just to avoid a log was too much, so I removed it, but feel free to re-introduce it in a way that avoids the other problem.

mykhailo-ua:
> 

Aha, sorry missed that comment regarding `sendSourceRemove`, but I think this is just the question of moving if statement to the right place (pic below). For us avoiding extra error logs means less error notifications which requires time for investigation, so I would add that part if you don't mind. Let me do it tomorrow and test if this is working fine. 
Screenshot 2023-03-23 at 18 25 43

bgrozev (approved):

bgrozev:
Thanks!

PR #1051, opened 2023-03-13
Add config to tell JVBs to use SCTP datachannels for relay-to-relay bridge channels.

JonathanLennox:


jicofo-common/src/main/kotlin/org/jitsi/jicofo/OctoConfig.kt
```diff
@@ -29,6 +29,10 @@ class OctoConfig {
         "jicofo.octo.allow-mixed-versions".from(newConfig)
     }
 
+    val sctpChannels: Boolean by config {
+        "jicofo.octo.sctp-channels".from(newConfig)
```
bgrozev, line 33: This doesn't match `reference.conf`  (it uses "sctp-datachannels")

bgrozev (approved):

PR #1055, opened 2023-03-17
Refactor ChatRoom

bgrozev:
- ref: Move avmoderation logic to a class, simplify.
- ref: Add @NotNull annotation.
- ref: Move ChatRoomImpl to kotlin.
- ref: Move ChatRoom to kotlin.
- Move ChatRoom[Impl] to org.jitsi.jicofo.xmpp.muc.
- Move kt file to the correct source tree.
- ref: Clean up the presence extensions API.
- ref: Cleanup the audio/video sender count code.
- ref: Cleanup startMuted code.
- ref: Cleanup ChatRoomImpl some more.
- ref: Merge av-moderation logic back into ChatRoomImpl.
- Add av_moderation to debug state (again).


jicofo-common/src/main/kotlin/org/jitsi/jicofo/xmpp/muc/ChatRoomAvModeration.kt
```diff
@@ -0,0 +1,81 @@
...
+        } ?: logger.warn("Invalid media type $mediaType")
+    }
+
+    fun setWhitelist(mediaType: MediaType, whitelist: List<String>) {
+        byMediaType[mediaType]?.let {
+            it.whitelist = whitelist
+        } ?: logger.warn("Invalid media type $mediaType")
+    }
+
+    private val byMediaType = mapOf(
```
JonathanLennox: Style is usually to put val's before fun's in Kotlin code, especially when the fun's use the val's.

jicofo/src/main/kotlin/org/jitsi/jicofo/xmpp/AvModerationHandler.kt
```diff
@@ -80,6 +80,8 @@ class AvModerationHandler(
 
                     val enabled = incomingJson["enabled"] as Boolean?
 
+                    // The message either updates the "enabled" state (if "enabled" is present), or the whitelists (if
+                    // "enabled" is not present).
```
JonathanLennox, line 84: It's not possible for a message to carry both enabled state and a whitelist?

jicofo-common/src/main/java/org/jitsi/impl/protocol/xmpp/ChatRoomImpl.kt
```diff
@@ -0,0 +1,769 @@
...
+    private val logger = createLogger().apply {
+        addContext("room", roomJid.toString())
+    }
+
+    private val memberListener: MemberListener = MemberListener()
+
+    private val userListener = LocalUserStatusListener()
+
+    /** Listener for presence that smack sends on our behalf. */
+    private var presenceInterceptor: org.jivesoftware.smack.util.Consumer<PresenceBuilder>? = null
```
JonathanLennox: Does this still need to be fully-qualified, since `leaveCallback` doesn't need to be declared as the other type of `Consumer`?

jicofo-common/src/main/java/org/jitsi/impl/protocol/xmpp/ChatRoomImpl.kt
```diff
@@ -0,0 +1,769 @@
...
+    }
+
+    /**
+     * Prepare this [ChatRoomImpl] for a call to [.joinAs], which send initial presence to
+     * the MUC. Resets any state that might have been set the previous time the MUC was joined.
+     */
+    private fun resetState() {
+        synchronized(members) {
+            if (members.isNotEmpty()) {
+                logger.warn("Removing " + members.size + " stale members.")
```
JonathanLennox: Should use string interpolation

jicofo-common/src/main/java/org/jitsi/impl/protocol/xmpp/ChatRoomImpl.kt
```diff
@@ -0,0 +1,769 @@
...
+            lastPresenceSent = null
+            meetingId = null
+            logger.addContext("meeting_id", "")
+            isBreakoutRoom = false
+            mainRoom = null
+            avModeration.reset()
+        }
+    }
+
+    @Throws(SmackException::class, XMPPException::class, InterruptedException::class)
```
JonathanLennox: Do you still need the `@Throws` declaration in Kotlin?

jicofo-common/src/main/java/org/jitsi/impl/protocol/xmpp/ChatRoomImpl.kt
```diff
@@ -0,0 +1,769 @@
...
+    private fun joinAs(nickname: Resourcepart) {
+        myOccupantJid = JidCreate.entityFullFrom(roomJid, nickname)
+        presenceInterceptor = org.jivesoftware.smack.util.Consumer { presenceBuilder: PresenceBuilder ->
```
JonathanLennox: You shouldn't need the argument type explicitly declared.

jicofo-common/src/main/java/org/jitsi/impl/protocol/xmpp/ChatRoomImpl.kt
```diff
@@ -0,0 +1,769 @@
...
+            muc.leave()
+        }
+        muc.addPresenceInterceptor(presenceInterceptor)
+        muc.createOrJoin(nickname)
+        val config = muc.configurationForm
+
+        // Read breakout rooms options
+        val isBreakoutRoomField = config.getField(MucConfigFields.IS_BREAKOUT_ROOM)
+        if (isBreakoutRoomField != null) {
+            isBreakoutRoom = java.lang.Boolean.parseBoolean(isBreakoutRoomField.firstValue)
```
JonathanLennox: You shouldn't need the `java.lang` here.  Also it seems like this could be expressed in a more Kotlin-y way?

jicofo-common/src/main/java/org/jitsi/impl/protocol/xmpp/ChatRoomImpl.kt
```diff
@@ -0,0 +1,769 @@
...
+
+    private fun leave(reason: String, jid: EntityBareJid) {
+        logger.info("Leave, reason: $reason alt-jid: $jid")
+        leave()
+    }
+
+    override fun leave() {
+        if (presenceInterceptor != null) {
+            muc.removePresenceInterceptor(presenceInterceptor)
+        }
```
JonathanLennox: Should be expressable with a .?let I think (which might make it possible to let removePresenceInterceptor's argument non-nullable?)

jicofo-common/src/main/java/org/jitsi/impl/protocol/xmpp/ChatRoomImpl.kt
```diff
@@ -0,0 +1,769 @@
...
+            if (o == null) {
+                return null
+            } else {
+                role = fromSmack(o.role, o.affiliation)
+            }
+        }
+        return role
+    }
+
+    /** Resets cached role instance so that it will be refreshed when [ ][.getUserRole] is called. */
```
JonathanLennox: Something a bit wacky with the auto-translated KDoc syntax here

jicofo-common/src/main/java/org/jitsi/impl/protocol/xmpp/ChatRoomImpl.kt
```diff
@@ -0,0 +1,769 @@
...
+     * Resets cached role instance for given occupantJid.
+     * @param occupantJid full mucJID of the occupant for whom we want to
+     * reset the cached role instance.
+     */
+    private fun resetRoleForOccupant(occupantJid: EntityFullJid) {
+        if (occupantJid.resourcepart == myOccupantJid!!.resourcepart) {
+            resetCachedUserRole()
+        } else {
+            val member = members[occupantJid]
+            if (member != null) {
```
JonathanLennox: Should be expressable more Kotliny

jicofo-common/src/main/java/org/jitsi/impl/protocol/xmpp/ChatRoomImpl.kt
```diff
@@ -0,0 +1,769 @@
...
+    }
+
+
+    /**
+     * {@inheritDoc}
+     */
+    @Synchronized
+    override fun containsPresenceExtension(elementName: String, namespace: String): Boolean {
+        return (lastPresenceSent != null
+                && lastPresenceSent!!.getExtension(QName(namespace, elementName))
```
JonathanLennox: Should be Kotlinizable

jicofo-common/src/main/java/org/jitsi/impl/protocol/xmpp/ChatRoomImpl.kt
```diff
@@ -0,0 +1,769 @@
...
+        admin.type = IQ.Type.set
+        admin.to = roomJid
+        val item = MUCItem(MUCAffiliation.owner, member.jid!!.asBareJid())
+        admin.addItem(item)
+        val connection = xmppProvider.xmppConnection
+        try {
+            val reply = connection.sendIqAndGetResponse(admin)
+            if (reply == null || reply.type != IQ.Type.result) {
+                // XXX FIXME throw a declared exception.
+                throw RuntimeException("Failed to grant owner: " + if (reply == null) "" else reply.toXML())
```
JonathanLennox: Should be kotlinizable and convertable to string interpolation

jicofo-common/src/main/java/org/jitsi/impl/protocol/xmpp/ChatRoomImpl.kt
```diff
@@ -0,0 +1,769 @@
...
+            if (lastPresenceSent == null) {
+                logger.error("No presence packet obtained yet")
+                return
+            }
+            var presenceUpdated = false
+
+            // Remove old
+            val old = lastPresenceSent!!.getExtension(extension.qName)
+            if (old != null) {
+                lastPresenceSent!!.removeExtension(old)
```
JonathanLennox: Is there a way to rewrite this so Kotlin's non-null inference works for lastPresenceSent?

jicofo-common/src/main/java/org/jitsi/impl/protocol/xmpp/ChatRoomImpl.kt
```diff
@@ -0,0 +1,769 @@
...
+            sendPresence(presenceToSend!!)
+        }
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Synchronized
+    override fun getPresenceExtensions(): Collection<ExtensionElement> {
+        return if (lastPresenceSent != null) ArrayList(lastPresenceSent!!.extensions) else emptyList()
```
JonathanLennox: Should be rewritable to avoid the !!

jicofo-common/src/main/java/org/jitsi/impl/protocol/xmpp/ChatRoomImpl.kt
```diff
@@ -0,0 +1,769 @@
...
+    ) {
+        var presenceToSend: Presence
+        synchronized(this) {
+            if (lastPresenceSent == null) {
+                logger.error("No presence packet obtained yet")
+                return
+            }
+
+            // Remove old
+            toRemove.forEach(java.util.function.Consumer { extension: ExtensionElement? ->
```
JonathanLennox: This should be Kotlinizable

jicofo-common/src/main/java/org/jitsi/impl/protocol/xmpp/ChatRoomImpl.kt
```diff
@@ -0,0 +1,769 @@
...
+     */
+    private fun processOwnPresence(presence: Presence) {
+        val mucUser = getMUCUserExtension(presence)
+        if (mucUser != null) {
+            val affiliation = mucUser.item.affiliation
+            val role = mucUser.item.role
+
+            // this is the presence for our member initial role and
+            // affiliation, as smack do not fire any initial
+            // events lets check it and fire events
```
JonathanLennox: Grammar of this comment (I know it's old) - "does not fire", "let's"

jicofo-common/src/main/java/org/jitsi/impl/protocol/xmpp/ChatRoomImpl.kt
```diff
@@ -0,0 +1,769 @@
...
+     */
+    private fun processOtherPresence(presence: Presence) {
+        val jid = presence.from.asEntityFullJidIfPossible() ?: run {
+            logger.warn("Presence without a valid jid: " + presence.from)
+            return
+        }
+
+        var memberJoined = false
+        var memberLeft = false
+        val member: ChatRoomMemberImpl? = synchronized(members) {
```
JonathanLennox: Does this need the explicit type declaration?

jicofo-common/src/main/java/org/jitsi/impl/protocol/xmpp/ChatRoomImpl.kt
```diff
@@ -0,0 +1,769 @@
...
+    }
+
+    /**
+     * Processes an incoming presence packet.
+     *
+     * @param presence the incoming presence.
+     */
+    override fun processPresence(presence: Presence?) {
+        if (presence == null || presence.error != null) {
+            logger.warn("Unable to handle packet: " + if (presence == null) "null" else presence.toXML())
```
JonathanLennox: This should work with string interpolation and presence?.toXML()

jicofo-common/src/main/java/org/jitsi/impl/protocol/xmpp/ChatRoomImpl.kt
```diff
@@ -165,10 +164,10 @@ class ChatRoomImpl(
      * the MUC. Resets any state that might have been set the previous time the MUC was joined.
      */
     private fun resetState() {
-        synchronized(members) {
-            if (members.isNotEmpty()) {
-                logger.warn("Removing " + members.size + " stale members.")
-                members.clear()
+        synchronized(membersMap) {
+            if (membersMap.isNotEmpty()) {
+                logger.warn("Removing " + membersMap.size + " stale members.")
```
JonathanLennox: String interpolation

jicofo-common/src/main/java/org/jitsi/impl/protocol/xmpp/ChatRoomImpl.kt
```diff
@@ -127,17 +130,13 @@ class ChatRoomImpl(
...
-    override fun getMeetingId(): String? = meetingId
     override fun addListener(listener: ChatRoomListener) = eventEmitter.addHandler(listener)
     override fun removeListener(listener: ChatRoomListener) = eventEmitter.removeHandler(listener)
-    override fun isJoined() = muc.isJoined
-    override fun getMembers(): List<ChatRoomMember> = synchronized(members) { return members.values.toList() }
-    override fun getChatMember(occupantJid: EntityFullJid) = members[occupantJid]
-    override fun getMembersCount() = members.size
+    override val isJoined
+        get() = muc.isJoined
+    override val members: List<ChatRoomMember>
```
JonathanLennox: Removing `membersSize` from the interface means that calls to `members.size` construct the full `members` list before returning its size.  How often do we call that?  Is this a performance consideration?

jicofo-common/src/main/kotlin/org/jitsi/jicofo/xmpp/muc/ChatRoomImpl.kt
```diff
@@ -362,78 +350,47 @@ class ChatRoomImpl(
...
-     * {@inheritDoc}
-     */
     override fun modifyPresence(
         toRemove: Collection<ExtensionElement>,
         toAdd: Collection<ExtensionElement>
     ) {
-        var presenceToSend: Presence
-        synchronized(this) {
-            if (lastPresenceSent == null) {
+        val presenceToSend: Presence? = synchronized(this) {
```
JonathanLennox, line 352: Does this need an explicit type declaration?

jicofo/src/main/kotlin/org/jitsi/jicofo/health/JicofoHealthChecker.kt
```diff
@@ -160,8 +162,8 @@ class JicofoHealthChecker(
         val p = Ping(JidCreate.bareFrom(xmppDomain))
         val listener = StanzaListener { packet: Stanza? -> pingResponseWait.countDown() }
         try {
-            xmppProvider.xmppConnection.addSyncStanzaListener(listener) {
-                stanza: Stanza -> stanza.stanzaId != null && stanza.stanzaId == p.stanzaId
+            xmppProvider.xmppConnection.addSyncStanzaListener(listener) { stanza: Stanza ->
+                stanza.stanzaId != null && stanza.stanzaId == p.stanzaId
```
JonathanLennox, line 166: I think Kotlin == operator means the != null is redundant?

jicofo/src/main/java/org/jitsi/jicofo/conference/JitsiMeetConferenceImpl.java
```diff
@@ -2055,11 +2055,11 @@ public void bridgeSelectionFailed()
         {
             ChatRoom chatRoom = getChatRoom();
             if (chatRoom != null
-                    && !chatRoom.containsPresenceExtension(
-                    BridgeNotAvailablePacketExt.ELEMENT,
-                    BridgeNotAvailablePacketExt.NAMESPACE))
+                    && chatRoom.getPresenceExtensions().stream().noneMatch(e ->
```
JonathanLennox, line 2058: Why did you remove containsPresenceExtension?

JonathanLennox (commented):
I went through this commit-by-commit, it looks like some of my comments were obviated by later commits, feel free to just resolve them if so.

jicofo-common/src/main/java/org/jitsi/impl/protocol/xmpp/ChatRoomImpl.kt
bgrozev: I don't think it's strictly needed, but it's nice to have IMO

jicofo-common/src/main/java/org/jitsi/impl/protocol/xmpp/ChatRoomImpl.kt
bgrozev: I'd have to specify the type of `Consumer` without it

jicofo-common/src/main/java/org/jitsi/impl/protocol/xmpp/ChatRoomImpl.kt
bgrozev: Oh, good point. I think that's why we introduced `getMembersCount()` in the first place 🤦

jicofo/src/main/java/org/jitsi/jicofo/conference/JitsiMeetConferenceImpl.java
bgrozev: To simplify the API.

jicofo/src/main/kotlin/org/jitsi/jicofo/xmpp/AvModerationHandler.kt
bgrozev: I asked the same question when Damian wrote the original code. I changed it to support both at the same time :)

jicofo-common/src/main/kotlin/org/jitsi/jicofo/xmpp/muc/ChatRoomImpl.kt
```diff
@@ -0,0 +1,565 @@
...
+
+    @Throws(SmackException::class, XMPPException::class, InterruptedException::class)
+    override fun join() {
+        // TODO: clean-up the way we figure out what nickname to use.
+        resetState()
+        joinAs(xmppProvider.config.username)
+    }
+
+    /**
+     * Prepare this [ChatRoomImpl] for a call to [.joinAs], which send initial presence to
```
JonathanLennox, line 171: Nit: "which sends"

jicofo-common/src/main/kotlin/org/jitsi/jicofo/xmpp/muc/ChatRoomMemberImpl.kt
```diff
@@ -75,25 +74,13 @@ class ChatRoomMemberImpl(
...
-                if (o == null) {
-                    return MemberRole.VISITOR // ??
-                } else {
-                    field = fromSmack(o.role, o.affiliation)
-                }
-            }
-            return field
+            return chatRoom.getOccupant(this)?.let {
+                return fromSmack(it.role, it.affiliation)
+            } ?: MemberRole.VISITOR
```
JonathanLennox, line 81: The double return here is a bit weird?

jicofo/src/main/java/org/jitsi/jicofo/conference/JitsiMeetConferenceImpl.java
bgrozev: Well, I prefer the complexity to be either in the underlying implementation or the code that uses the interface, not the interface itself.

I changed it again. Now there's some extra complexity in ChatRoomImpl, but I think everything else is simpler.

jicofo-common/src/main/kotlin/org/jitsi/jicofo/xmpp/muc/ChatRoomImpl.kt
bgrozev: Sorry, there were two instances and I only fixed one.

jicofo-common/src/main/kotlin/org/jitsi/jicofo/xmpp/muc/ChatRoomImpl.kt
```diff
@@ -294,35 +294,58 @@ class ChatRoomImpl(
...
-                null
             }
         }
+        presenceToSend?.let { xmppProvider.xmppConnection.tryToSendStanza(it) }
+    }
+
+    override fun removePresenceExtensions(pred: (ExtensionElement) -> Boolean) {
+        val presenceToSend = synchronized(this) {
+            lastPresenceSent?.extensions?.filter { pred(it) }?.let {
+                modifyPresenceExtensions(toRemove = it)
```
JonathanLennox, line 337: This doesn't trigger a ConcurrentModificationException?

bgrozev: Not unless I'm misunderstanding something. I accesses `lastPresenceSent?.extensions`, filters the result to a list, and then calls `modifyPresenceExtensions`.

JonathanLennox: Oh, ok, because filter constructs a whole new list object, rather than being another iterator - right.

JonathanLennox (approved):

PR #1058, opened 2023-03-22
Update metrics periodically, port more stats to metrics.

bgrozev:
- feat: Update metrics periodically instead of when /stats is accessed.
- Move the threads stat to a metric.
- Add a "healthy" metrics.
- Convert BridgeSelector stats to metrics.
- Bring back bridges_removed, now as a metric.
- Port jibri counts to metrics.

### DIFF
--- a/jicofo-common/src/main/kotlin/org/jitsi/jicofo/metrics/JicofoMetricsContainer.kt
+++ b/jicofo-common/src/main/kotlin/org/jitsi/jicofo/metrics/JicofoMetricsContainer.kt
@@ -17,11 +17,67 @@
  */
 package org.jitsi.jicofo.metrics
 
+import org.jitsi.config.JitsiConfig
+import org.jitsi.jicofo.TaskPools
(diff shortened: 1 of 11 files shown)


jicofo-common/src/main/kotlin/org/jitsi/jicofo/metrics/JicofoMetricsContainer.kt
```diff
@@ -17,11 +17,67 @@
...
+                it.invoke()
+            } catch (e: Exception) {
+                logger.warn("Exception while running subtask", e)
+            }
+        }
+    }
+
+    fun stop() = synchronized(this) {
+        updateTask?.cancel(false)
+        updateTask = null
```
JonathanLennox, line 72: Might want to clear subtasks to make sure we don't hold on to handles to anything - though if this only happens on shutdown it's probably not a big deal

JonathanLennox (approved):

PR #1090, opened 2023-05-05
fix: Fix audio/video limits since muted stated was moved to SourceInfo.

bgrozev:


jicofo-common/src/main/kotlin/org/jitsi/jicofo/xmpp/muc/ChatRoomMemberImpl.kt
```diff
@@ -21,17 +21,16 @@ import org.jitsi.jicofo.xmpp.Features
 import org.jitsi.jicofo.xmpp.XmppCapsStats
 import org.jitsi.jicofo.xmpp.XmppConfig
 import org.jitsi.jicofo.xmpp.muc.MemberRole.Companion.fromSmack
+import org.jitsi.utils.MediaType
 import org.jitsi.utils.OrderedJsonObject
 import org.jitsi.utils.logging2.Logger
 import org.jitsi.utils.logging2.createChildLogger
-import org.jitsi.xmpp.extensions.jitsimeet.AudioMutedExtension
 import org.jitsi.xmpp.extensions.jitsimeet.FeaturesExtension
```
damencho, line 28: Does it matter what jigasi is sending? As It still uses those extensions?

jicofo-common/src/main/kotlin/org/jitsi/jicofo/xmpp/muc/ChatRoomMemberImpl.kt
```diff
@@ -175,22 +188,6 @@ class ChatRoomMemberImpl(
         presence.getExtension(StatsId::class.java)?.let {
             statsId = it.statsId
         }
-
```
jallamsetty1, line 178: Should we not leave this in if the endpoint doesn't support source-names ?

bgrozev: Do endpoints with source name support always include `SourceInfo` (even when they don't have any tracks)?

jallamsetty1: Nope, SourceInfo gets added only if there are local tracks.

jallamsetty1: Sorry, just realized that we send an empty SourceInfo on join if the endpoint doesn't have any local tracks. 
"<SourceInfo>{}</SourceInfo>"

jicofo-common/src/main/kotlin/org/jitsi/jicofo/xmpp/muc/SourceInfo.kt
```diff
@@ -16,15 +16,31 @@
...
+private val logger = LoggerImpl(SourceInfo::javaClass.name)
+
 /** The information about a source contained in a jitsi-meet SourceInfo extension. */
 data class SourceInfo(
     val name: String,
     val muted: Boolean,
     val videoType: VideoType?
-)
+) {
+    val mediaType: MediaType? = name.indexOf('-').let { indexOfDash ->
```
JonathanLennox, line 32: I'd rather we signal this explicitly rather than implicitly as here, though that can potentially be a separate PR (and would require client changes).

JonathanLennox (approved):

JonathanLennox (approved):

PR #1098, opened 2023-06-08
Count visitors added to a ChatRoom in the last 500 ms toward the visitorCount.

JonathanLennox:


JonathanLennox:
Now needs https://github.com/jitsi/jitsi-utils/pull/130

damencho (approved):

PR #1101, opened 2023-06-15
Avoid counting participants

JonathanLennox:
Iterating over participant lists causes quadratic behavior when adding large numbers of visitors to a conference.

Instead, keep counts of the relevant values in parallel with the maps.

### DIFF
--- a/jicofo/src/main/java/org/jitsi/jicofo/conference/JitsiMeetConferenceImpl.java
+++ b/jicofo/src/main/java/org/jitsi/jicofo/conference/JitsiMeetConferenceImpl.java
@@ -956,10 +956,6 @@ public class JitsiMeetConferenceImpl
             {
                 expireBridgeSessions();
             }
-
-            visitorCount.setValue(() -> (int) participants.values().stream()
-                .filter(p -> p.getChatMember().getRole() == MemberRole.VISITOR)
-                .count());
         }
 
         maybeStop();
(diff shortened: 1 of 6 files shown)


jicofo/src/main/java/org/jitsi/jicofo/conference/JitsiMeetConferenceImpl.java
```diff
@@ -810,7 +805,16 @@ private void inviteChatMember(ChatRoomMember chatRoomMember, boolean justJoined)
                 ConferenceMetrics.participantsNoSourceName.inc();
             }
 
-            participants.put(chatRoomMember.getOccupantJid(), participant);
+            boolean added = (participants.put(chatRoomMember.getOccupantJid(), participant) == null);
+            if (added && participant.isUserParticipant())
+            {
+                userParticipantAdded();
+            }
+            if (added && participant.getChatMember().getRole() == MemberRole.VISITOR)
```
damencho, line 813: Minor: maybe add and else here?

jicofo/src/main/java/org/jitsi/jicofo/conference/JitsiMeetConferenceImpl.java
```diff
@@ -1021,6 +1023,14 @@ private void terminateParticipant(
             Participant removed = participants.remove(participant.getChatMember().getOccupantJid());
             logger.info(
                     "Removed participant " + participant.getChatMember().getName() + " removed=" + (removed != null));
+            if (!willReinvite && removed != null && removed.isUserParticipant())
+            {
+                userParticipantRemoved();
+            }
+            if (!willReinvite && removed != null && removed.getChatMember().getRole() == MemberRole.VISITOR)
```
damencho, line 1030: Maybe an else? And just one check for `!willReinvite && removed != null`?

damencho (approved):

damencho (approved):

PR #1116, opened 2023-10-02
Update for ktlint 1.0.0 (ktlint-maven-plugin 3.0.0).

JonathanLennox:


jicofo-common/src/main/kotlin/org/jitsi/jicofo/xmpp/Util.kt
```diff
@@ -58,14 +58,13 @@ fun parseJidFromClientProxyJid(
...
-        sendStanza(stanza)
-    } catch (e: SmackException.NotConnectedException) {
-        logger.error("No connection - unable to send packet: " + stanza.toXML(), e)
-    } catch (e: InterruptedException) {
-        logger.error("Failed to send packet: " + stanza.toXML().toString(), e)
-    }
+fun XMPPConnection.tryToSendStanza(stanza: Stanza) = try {
+    sendStanza(stanza)
+} catch (e: SmackException.NotConnectedException) {
+    logger.error("No connection - unable to send packet: " + stanza.toXML(), e)
```
bgrozev, line 64: Looks like it doesn't enforce string templates?

JonathanLennox: Guess not.  I'll suggest it to them.

PR #1118, opened 2023-10-26
avoid xmlstring builder toString

bgrozev:
- chore: Update jitsi-xmpp-extensions.
- log: Remove System.err logs.
- fix: Avoid using XmlStringBuilder.toString.

### DIFF
--- a/jicofo-common/src/main/java/org/jitsi/impl/protocol/xmpp/log/PacketDebugger.java
+++ b/jicofo-common/src/main/java/org/jitsi/impl/protocol/xmpp/log/PacketDebugger.java
@@ -17,6 +17,7 @@ package org.jitsi.impl.protocol.xmpp.log;
 
 import edu.umd.cs.findbugs.annotations.*;
 import org.jitsi.utils.logging2.*;
+import org.jitsi.xmpp.util.*;
 import org.jivesoftware.smack.*;
 import org.jivesoftware.smack.debugger.*;
 import org.jivesoftware.smack.packet.*;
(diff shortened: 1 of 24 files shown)


jicofo-common/src/main/java/org/jitsi/impl/protocol/xmpp/log/PacketDebugger.java
```diff
@@ -66,13 +67,17 @@ public PacketDebugger(XMPPConnection connection, @NonNull String id)
...
+                () -> "RCV PKT (" + id + "): " + XmlStringBuilderUtil.Companion.toStringOpt(streamElement.toXML())
+        );
     }
 
     @Override
     public void onOutgoingStreamElement(TopLevelStreamElement streamElement)
     {
-        logger.debug(() -> "SENT PKT (" + id + "): " + streamElement.toXML());
+        logger.debug(
+                () -> "SENT PKT (" + id + "): " + XmlStringBuilderUtil.Companion.toStringOpt(streamElement.toXML())
```
JonathanLennox, line 79: If your method in jitsi-xmpp-extensions were on `Element` rather than `ExtensionElement` this could just use the non-companion flavor of `toStringOpt` -- is it worth it?  (I think it clearly would be if this were Kotlin, but it's less obvious for Java.)

bgrozev: I noticed this was awkward, but I didn't think it's worth another jitsi-xmpp-extensions update

JonathanLennox (approved):

PR #1119, opened 2023-10-27
fix: Do not leave the conference soon after a conference-request.

bgrozev:
Fixes a race condition in which the conference was stopped while a
client was in the process of joining.


damencho (approved):

bgrozev:
Jicofo needs to leave the room when the last occupant leaves, otherwise "end conference" can not be enforced. We'll address the issue with a client reload instead

PR #1135, opened 2024-02-08
feat: Enable transcription based on MUC config form

bgrozev:
Read room_metadata from the MUC config form, and enable transcription if
it is requested.

### DIFF
--- a/jicofo/src/main/java/org/jitsi/jicofo/jigasi/TranscriberManager.java
+++ b/jicofo/src/main/java/org/jitsi/jicofo/jigasi/TranscriberManager.java
@@ -77,7 +77,8 @@ public class TranscriberManager
 
     /**
      * A single-threaded {@link ExecutorService} to offload inviting the
-     * Transcriber from the smack thread updating presence.
+     * Transcriber from the smack thread updating presence. It's important that requests are handled sequentially to
+     * prevent multiple jigasis being invited.
      */
     private final ExecutorService executorService = Executors.newSingleThreadExecutor();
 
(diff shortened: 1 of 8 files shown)


jicofo/src/main/java/org/jitsi/jicofo/jigasi/TranscriberManager.java
```diff
@@ -125,20 +126,37 @@ private void memberPresenceChanged(@NotNull ChatRoomMember member)
...
-            if (jigasiDetector == null)
+            tryToStart();
+        }
+    }
+
+    private void tryToStart()
+    {
+        if (jigasiDetector == null)
+        {
+            logger.warn("Transcription requested, but jigasiDetector is not configured.");
```
saghul, line 142: Note for the future: it would be useful to somehow propagate this all the way to the client.

jicofo/src/main/java/org/jitsi/jicofo/jigasi/TranscriberManager.java
```diff
@@ -275,5 +277,15 @@ public void memberPresenceChanged(@NotNull ChatRoomMember member)
         {
             TranscriberManager.this.memberPresenceChanged(member);
         }
+
+        @Override
+        public void transcriptionRequestedChanged(boolean transcriptionRequested)
+        {
+            if (transcriptionRequested)
```
saghul, line 284: Shouldn't this check for the active flag too? Note that right now there will be 2 ways to trigger this and both will happen roughly at once:

- presence
- room metadata

jicofo/src/main/java/org/jitsi/jicofo/jigasi/TranscriberManager.java
bgrozev: Agreed

jicofo/src/main/java/org/jitsi/jicofo/jigasi/TranscriberManager.java
bgrozev: That's why the tasks are put on a single threaded executor. I can check the flag here, but I'll have to re-check it in the other thread too.

saghul: 👍  I'm not familiar with the code so I thought I'd ask. SGTM!

saghul (approved):

jicofo/src/main/java/org/jitsi/jicofo/jigasi/TranscriberManager.java
bgrozev: It's a good question

PR #1138, opened 2024-02-24
Update README.md

hasandiwan:
Fix URL to Debian quickstart


damencho:
You can use https://jitsi.org/qi

PR #1156, opened 2024-06-21
fix: Fix scheduling jvb health checks.

bgrozev:
Scheduling at a fixed rate leads to IO threads building up when a bridge
becomes unresponsibe and the duration of health checks becomes longer
than the health check interval.


JonathanLennox:
I think your solution is the same as `ScheduledExecutorService.scheduleWithFixedDelay`, isn't it?

bgrozev:
I actually used `scheduleWithFixedDelay` initially but it didn't go well :)

It `Submits a periodic action that becomes enabled first after the given initial delay, and subsequently with the given delay`

bgrozev:
Wow, I see what you mean 🤦

JonathanLennox (approved):

PR #1166, opened 2024-09-17
feat: Verify the rayo JvbRoomName header if it exists.

bgrozev:
Note it's just a precaution, prosody already verifies it.

### DIFF
--- a/jicofo/src/main/kotlin/org/jitsi/jicofo/xmpp/JigasiIqHandler.kt
+++ b/jicofo/src/main/kotlin/org/jitsi/jicofo/xmpp/JigasiIqHandler.kt
@@ -35,6 +35,7 @@ import org.jivesoftware.smack.packet.IQ
 import org.jivesoftware.smack.packet.StanzaError
 import org.jivesoftware.smack.packet.id.StandardStanzaIdSource
 import org.jxmpp.jid.Jid
+import org.jxmpp.jid.impl.JidCreate
 import java.util.concurrent.atomic.AtomicInteger
 
 class JigasiIqHandler(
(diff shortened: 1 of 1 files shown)


jicofo/src/main/kotlin/org/jitsi/jicofo/xmpp/JigasiIqHandler.kt
```diff
@@ -82,6 +78,17 @@ class JigasiIqHandler(
             }
         }
 
+        val roomNameHeader = request.iq.getHeader("JvbRoomName")
```
damencho, line 81: Does this search the whole structure of the stanza? Cause the dial-iq is like:
```
<iq to = 'focus@auth.meet.jitsi/focus' type = 'set' xml:lang = 'en'
    id = ''
    from = 'damencho@conference.tenant.v1.meet.jitsi/36ff70d6'>
    <dial to = 'jitsi_meet_transcribe' xmlns = 'urn:xmpp:rayo:1' from = 'fromnumber'>
        <header value = 'damencho@conference.tenant.v1.meet.jitsi' name = 'JvbRoomName' />
    </dial>
</iq>
```

bgrozev: I think so, this is the `getHeader` from `RayoIq`
https://github.com/jitsi/jitsi-xmpp-extensions/blob/3816e5a154160056290c2a2fbff1ab63cbac95e0/src/main/java/org/jitsi/xmpp/extensions/rayo/RayoIq.java#L55

damencho: But that gets all direct child extensions and the only child of the IQ is a dial extension. The header is the child of the dial extension.

damencho: Oh this method gets the DialIq `request: IqRequest<DialIq>`. Sorry for the noise.

damencho (dismissed):

jicofo/src/main/kotlin/org/jitsi/jicofo/xmpp/JigasiIqHandler.kt
bgrozev: I think this test verifies it works as expected?
https://github.com/jitsi/jitsi-xmpp-extensions/blob/3816e5a154160056290c2a2fbff1ab63cbac95e0/src/test/java/org/jitsi/xmpp/extensions/rayo/DialIqProviderTest.java#L138

damencho (approved):

PR #1168, opened 2024-09-18
fix: Wait for disconnect response before leaving visitor room.

bgrozev:


damencho:
Will no need this. This destroy is used to handle the case where the visitor room is destroyed due to no visitors.
Which is not the problem we were chasing. The problem is main room destroyed then visitor room destroyed and it happen jicofo to leave first before we are being able to tell destroy the room with a message so the visitors to see it.

PR #1190, opened 2024-11-25
Reconnect logic

damencho:


jicofo-common/src/main/kotlin/org/jitsi/jicofo/xmpp/XmppProvider.kt
```diff
@@ -97,6 +97,14 @@ class XmppProvider(val config: XmppConnectionConfig, parentLogger: Logger) {
         override fun authenticated(connection: XMPPConnection?, resumed: Boolean) {
             registered = true
             logger.info("Registered.")
+            if (connection is XMPPTCPConnection) {
+                logger.info(
```
bgrozev, line 101: Can we have a single log message and include the `resumed` flag just in case?

bgrozev (approved):

PR #1207, opened 2025-01-24
fix: Do not stop health checks after an error response.

bgrozev:


jicofo/src/main/java/org/jitsi/jicofo/conference/JitsiMeetConferenceImpl.java
```diff
@@ -2325,18 +2325,25 @@ public void bridgeIsShuttingDown(@NotNull Bridge bridge)
         }
 
         @Override
-        public void bridgeRemoved(@NotNull Bridge bridge)
+        public void bridgeFailedHealthCheck(@NotNull Bridge bridge)
```
JonathanLennox, line 2328: I suggest having a common helper function that both of these call, and passing a reason phrase, would be better than having `bridgeRemoved` call `bridgeFailedHealthCheck` and leaving it ambiguous in the log message.

JonathanLennox (approved):

PR #1217, opened 2025-04-09
feat: Read startMuted configuration from room metadata (only).

damencho:


damencho:
jenkins test this please.

PR #1235, opened 2025-07-01
feat: Handles mute desktop IQ.

damencho:


jicofo/src/main/java/org/jitsi/jicofo/conference/JitsiMeetConferenceImpl.java
```diff
@@ -1593,7 +1594,12 @@ else if (!participant.hasModeratorRights()
         logger.info("Will " + (doMute ? "mute" : "unmute") + " " + toBeMutedJid + " on behalf of " + muterJid
             + " for " + mediaType);
 
-        getColibriSessionManager().mute(participant.getEndpointId(), doMute, mediaType);
+        // we ignore desktop as for that we use some signaling restrictions in prosody
+        if (mediaType == MediaType.AUDIO || mediaType == MediaType.VIDEO)
+        {
+            getColibriSessionManager().mute(participant.getEndpointId(), doMute, mediaType);
```
saghul, line 1600: I guess here is where we need to check if the user if not on both whitelists before doing this?

PR #1250, opened 2025-08-28
fix: Handle XMPP-related events in a per-ChatRoom queue.

bgrozev:


jicofo-common/src/main/kotlin/org/jitsi/jicofo/xmpp/muc/ChatRoomImpl.kt
```diff
@@ -103,13 +103,20 @@ class ChatRoomImpl(
...
+    private val xmppTaskQueue = PacketQueue<Runnable>(
         Integer.MAX_VALUE,
         false,
         "ChatRoomImpl presence queue",
-        { doProcessPresence(it) },
+        {
+            try {
+                it.run()
+            } catch (e: Exception) {
+                logger.warn("Error processing ", e)
```
JonathanLennox, line 116: Incomplete message?  Probably should say "Error processing XMPP task".

jicofo-common/src/main/kotlin/org/jitsi/jicofo/xmpp/muc/ChatRoomImpl.kt
```diff
@@ -400,7 +410,7 @@ class ChatRoomImpl(
         // Unblock any threads waiting on the latches
         roomMetadataLatch.countDown()
         roomJoinedLatch.countDown()
-        presenceQueue.close()
+        xmppTaskQueue.close()
```
JonathanLennox, line 413: If you have tasks on the queue they could run at this point (between `roomJoinedLatch.countDown()` and `xmppTaskQueue.close()`.). Is that a problem?

PR #1266, opened 2025-12-02
feat: Add configurable HTTP headers to colibri export for transcription.

bgrozev:


jicofo-selector/src/main/resources/reference.conf
```diff
@@ -439,6 +439,14 @@ jicofo {
     # will be replaced with the unique meeting ID and the region of the bridge used for transcription. The region is
     # optional. Note no whitespace is allowed in the templates, they have to match exactly.
     // url-template = "wss://{{REGION}}.example.com/recorder/{{MEETING_ID}}"
+
+    # HTTP headers to include when connecting to the transcription service. These can include authentication tokens,
+    # API keys, content type specifications, etc.
+    http-headers {
+      // "Authorization" = "Bearer your-api-key"
+      // "Content-Type" = "application/json"
+      // "User-Agent" = "Jicofo/1.0"
```
JonathanLennox: This example is misleading because the websocket will be coming from the JVB.

PR #1284, opened 2026-06-01
Replace json-simple with Jackson

bgrozev:
Replace the `org.json.simple` dependency with Jackson (`jackson-databind`, `jackson-module-kotlin`).

- Debug state methods now return `JsonNode`
- Update `jitsi-utils` to `1.0-SNAPSHOT`


JonathanLennox (approved):
Looks good, pending update of dependencies

PR #1286, opened 2026-06-17
feat: Live audio translation support

bgrozev:
Adds the jicofo side of live audio translation.

- Maintains synthetic audio sources (one per requested target language, named `<base>.<lang>`) in the conference source map, driven by the aggregated request map from RoomMetadata.
- Drives the translator `<connect>` to the bridge(s), signaled as **id-keyed create/expire deltas** (matching jitsi/jitsi-xmpp-extensions#143).
- Two placement modes via `jicofo.translation.mode`:
  - **per-source** (default) — each sender's audio is translated by connect(s) on its own local bridge, split into multiple connects when it has more than `max-languages-per-connect` target languages;
  - **single-bridge** — all translation aggregated onto one selected bridge (previous behaviour). `max-languages-per-connect` applies only to per-source.
- Configurable translator URL template and HTTP headers.

**Depends on jitsi/jitsi-xmpp-extensions#143** for the connect id/create/expire attributes. Will not compile against a released xmpp-ext until #143 merges and is released, after which the pom dependency gets bumped.

### DIFF
--- a/jicofo-selector/src/main/kotlin/org/jitsi/jicofo/bridge/colibri/ColibriV2SessionManager.kt
+++ b/jicofo-selector/src/main/kotlin/org/jitsi/jicofo/bridge/colibri/ColibriV2SessionManager.kt
@@ -138,6 +138,9 @@ class ColibriV2SessionManager(
     /** The current per-sender translation requests. */
     private var translatorRequests: List<TranslationRequest> = emptyList()
 
+    /** Per-room translator connect headers (merged config + room metadata), or null to use config headers. */
+    private var translatorCustomHeaders: Map<String, String>? = null
+
     /**
      * The colibri2 sessions that are currently active, mapped by the relayId of the [Bridge] that they use.
      */
@@ -391,7 +394,7 @@ class ColibriV2SessionManager(
         type = Connect.Types.TRANSLATOR,
         exports = translatorRequests.map { it.sourceName },
         requests = translatorRequests.flatMap { it.syntheticSourceNames },
-        httpHeaders = TranslationConfig.config.httpHeaders,
+        httpHeaders = translatorCustomHeaders ?: TranslationConfig.config.httpHeaders,
         ping = translatorPing()
     )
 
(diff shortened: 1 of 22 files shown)


jicofo-selector/src/main/kotlin/org/jitsi/jicofo/bridge/colibri/ColibriV2SessionManager.kt
```diff
@@ -64,6 +64,27 @@ private const val TRANSCRIBER_CONNECT_ID = "transcriber"
...
+ * synthetic outputs. Connect ids are "translator-<senderEndpointId>-<chunkIndex>" so they are stable across updates.
+ */
+internal fun perSourceTranslatorSpecs(
+    request: TranslationRequest,
+    url: URI,
+    maxLanguagesPerConnect: Int,
+    httpHeaders: Map<String, String>
+): List<ConnectSpec> = request.syntheticSourceNames.chunked(maxLanguagesPerConnect).mapIndexed { index, chunk ->
+    ConnectSpec(
+        id = "$TRANSLATOR_CONNECT_ID-${request.senderEndpointId}-$index",
```
JonathanLennox: If an ep has more than one audio source, and we request translations for both of them, won't this collide?

jicofo-selector/src/main/kotlin/org/jitsi/jicofo/bridge/colibri/ColibriV2SessionManager.kt
```diff
@@ -64,6 +64,27 @@ private const val TRANSCRIBER_CONNECT_ID = "transcriber"
...
+ * The per-source translator connect specs for a single [request]: one connect per chunk of at most
+ * [maxLanguagesPerConnect] requested languages, each exporting the sender's source and requesting its chunk of
+ * synthetic outputs. Connect ids are "translator-<senderEndpointId>-<chunkIndex>" so they are stable across updates.
+ */
+internal fun perSourceTranslatorSpecs(
+    request: TranslationRequest,
+    url: URI,
+    maxLanguagesPerConnect: Int,
+    httpHeaders: Map<String, String>
+): List<ConnectSpec> = request.syntheticSourceNames.chunked(maxLanguagesPerConnect).mapIndexed { index, chunk ->
```
JonathanLennox, line 84: When we're overflowing into multiple chunks and a language in the first chunk is removed, the first language in the second chunk will move translators.  This might cause glitches?  I don't know if keeping track of the chunks explicitly is worth it though, since someone being translated into this many languages should be rare.

JonathanLennox (approved):

PR #1288, opened 2026-06-22
feat(jibri): wait and retry selection when all Jibris are busy

aaronkvanmeerten:
## Problem

When a recording/streaming start request arrives and **all connected Jibri instances are busy**, jicofo fails immediately with a `resource_constraint` error and the recording never starts.

The existing `jicofo.jibri.num-retries` does **not** help this case: it only moves on to a *different* instance after one fails. The moment `JibriDetector.selectJibri()` returns no available instance, `startInternal()` throws `AllBusy` and gives up — there is no way to make jicofo **wait** for the pool to free up.

In production we see this as a hard recording failure during brief capacity shortfalls:
- a demand spike at the top of the hour while the autoscaler is still spinning up instances, or
- two start requests racing for the last free instance (the second arrives a few hundred ms too late).

## Change

Add an optional **asynchronous** busy-retry on the "all busy" path, controlled by two new config properties:

| Property | Default | Meaning |
|---|---|---|
| `jicofo.jibri.busy-retries` | `0` (disabled) | How many extra times to re-select when all instances are busy |
| `jicofo.jibri.busy-retry-delay` | `5 seconds` | How long to wait between busy-retries |

When all connected instances are busy, jicofo schedules a retry (up to `busy-retries` times) and **returns immediately**. The client gets a `result` and learns the eventual outcome via presence: the recording turns **ON** if an instance frees up, or is reported **failed** (`OFF`/`ERROR`) once the retries are exhausted. The default of `0` preserves current behavior exactly (immediate `resource_constraint` error).

The "no instances connected at all" case (`NotAvailable`) still fails fast.

## Design notes (async, non-blocking)

The retry is **not** a blocking wait. An earlier revision of this PR used `Thread.sleep` on the synchronized start path, but that runs on the per-room XMPP task-queue thread (`TaskPools.ioPool`) and would (a) block that room's request queue for the whole wait, (b) hold the `JibriSession` monitor (delaying a concurrent `stop()`), and (c) multiply the total wait by `num-retries` since the num-retries path re-enters the wait. The current implementation instead:

- schedules the delay on the **scheduled pool** (3 threads, timing only) and runs the actual start-IQ exchange on the **IO pool** — neither the calling thread nor the session monitor is held during the wait;
- tracks `busyRetries` as an instance field, so the total is bounded by `busy-retries` across the whole session regardless of `num-retries`;
- cancels a pending retry on `stop()` (with a guard for the already-firing race) and clears presence if the session was still waiting for capacity.

## Tests

`JibriSessionTest` adds two async cases (verified with `verify(timeout=…)`):
- busy-retries enabled: selection returns nothing twice then frees up → the session starts after retrying;
- busy-retries exhausted: pool never frees up → the owner is notified of failure via presence.

All existing tests pass; build and checkstyle are clean (Java 17/21 CI green).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/jicofo-selector/src/main/resources/reference.conf
+++ b/jicofo-selector/src/main/resources/reference.conf
@@ -303,6 +303,19 @@ jicofo {
     // How many times to retry a given Jibri request before giving up. Set to -1 to allow infinite retries.
     num-retries = 5
 
+    // When a start request can not be served because all connected Jibri instances are busy, retry the selection
+    // this many additional times (waiting `busy-retry-delay` between attempts) before failing with a
+    // "resource_constraint" error. This bridges brief capacity shortfalls, e.g. a demand spike at the top of the
+    // hour while the autoscaler spins up more instances, or two requests racing for the last free instance. Unlike
+    // `num-retries` (which moves on to a *different* instance after one fails), this waits for the pool itself to
+    // free up. Set to 0 (the default) to fail immediately, preserving the previous behavior. Note that this delays
+    // the response to the client and briefly blocks the conference's request queue, so keep the total wait
+    // (`busy-retries` * `busy-retry-delay`) modest.
+    busy-retries = 0
+
+    // How long to wait between busy retries (see `busy-retries`). Only used when `busy-retries` > 0.
+    busy-retry-delay = 5 seconds
+
     // How long to wait for Jibri to start recording from the time it accepts a START request.
     pending-timeout = 90 seconds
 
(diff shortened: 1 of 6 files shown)


aaronkvanmeerten:
Updated: reworked the busy-retry to be **asynchronous** (commit 52e5e03e). The first cut used a blocking `Thread.sleep` on the synchronized start path; that blocked the per-room XMPP task queue, held the session monitor (delaying `stop()`), and multiplied the wait by `num-retries`. It now schedules the delay on the scheduled pool and runs the attempt on the IO pool, returns immediately, and reports the eventual outcome via presence. Bounded by `busy-retries` across the session. PR description updated accordingly.

aaronkvanmeerten:
Folded in a review fix (commit 1427dd65): a `busy-retry-delay` of 0 no longer silently disables the feature. Since retries are now async, 0 is honored as "retry immediately" (no thread spin); the feature is disabled only via `busy-retries = 0`. Added a test.

PR #1290, opened 2026-07-03
feat: ensure to leave chatRoom on failure

savely-krasovsky:
When Jicofo starts before Prosody is ready, might try to join a chat room and faill. I know it shouln't happen, but due to some bug in our charts probably, it's sometimes not.

1. `BaseBrewery.start()` calls `xmppProvider.createRoom(jvbbrewery)`.
2. `createRoom()` puts the `ChatRoomImpl` into the local map before `join()`.
3. `join()` fails with some error.
4. `BaseBrewery` only removed the listener and set `chatRoom = null`, but did not remove the room from `XmppProvider.Muc.rooms`.
5. The next attempt to start the brewery fails with `RoomExistsException`, and Jicofo remains Ready, but without Joined the room / bridges.


savely-krasovsky:
I already signed the CLA.

bgrozev (approved):

bgrozev:
Thank you!

PR #1303, opened 2026-08-18
feat: Check that endpoints advertise required capabilities.

bgrozev:
Jicofo now checks that endpoints advertise the capabilities that the deployment requires, and tells an endpoint which does not.

- The set of requirements is maintained in `ClientRequirementsConfig.defaultRequirements`. The set is empty for now. Configuration can change the level, the text and the URL of an entry, disable an entry, or add a new one.
- An endpoint which does not advertise a capability gets a `client-requirements` IQ. An endpoint which does not support the IQ gets one chat message for each capability, which any client renders.
- An endpoint which does not advertise a capability with a level of `hard` is not invited to the conference. It stays in the MUC and can still use the features which do not need a media session, for example chat. Jicofo does not kick it.
- Enforcement is off by default (`enforce = false`). In that mode jicofo only logs, updates metrics, and sends a warning.
- Jicofo never enforces a requirement when it could not discover the features of an endpoint. This keeps a failed disco#info from locking endpoints out.
- Jibri, jigasi and the transcriber are not checked.
- Each endpoint is checked once. A later re-invite uses the first result and does not notify the endpoint again.
- New metrics: `participants_missing_features` (with feature and level labels), `participants_rejected_missing_features`, `participants_features_not_discovered` and `client_requirements_notifications`.
- The "Member joined" log line now includes the client version, which the client advertises in presence.

Needs the IQ from the jitsi-xmpp-extensions branch with the same name.

Set of PRs, all on a branch named `client-capability-requirements`:

- jitsi-xmpp-extensions: https://github.com/jitsi/jitsi-xmpp-extensions/pull/149
- jicofo: https://github.com/jitsi/jicofo/pull/1303
- lib-jitsi-meet: https://github.com/jitsi/lib-jitsi-meet/pull/3091
- jitsi-meet: https://github.com/jitsi/jitsi-meet/pull/17730

### DIFF
--- a/jicofo/src/main/resources/reference.conf
+++ b/jicofo/src/main/resources/reference.conf
@@ -85,6 +85,32 @@ jicofo {
         max-requests = 5
         interval = 1 minute
     }
+
+    // The capabilities that endpoints are required to advertise. An endpoint which does not advertise a capability is
+    // notified, and (for a "hard" requirement) is not invited to the conference. It stays in the MUC and can still use
+    // the features which do not need a media session, e.g. chat.
+    client-requirements {
+      // Whether to check the capabilities of endpoints.
+      enabled = true
+
+      // Whether to actually skip inviting an endpoint which does not advertise a capability with a "hard" level. When
+      // this is false jicofo only logs and updates metrics, and notifies the endpoint as if the level were "soft".
+      enforce = false
+
+      // The requirements, keyed by the name of a feature known to jicofo (see Features.kt). The set of requirements is
+      // maintained in jicofo (see ClientRequirementsConfig.defaultRequirements), and the entries here change it.
+      //   level:   "hard" (the endpoint is not invited), "soft" (the endpoint is invited and warned), or "off" (the
+      //            requirement is disabled).
+      //   details: text (in English) sent to the endpoint, which describes how to add support for the capability.
+      //   url:     a URL with more information, sent to the endpoint.
+      requirements {
+        #SSRC_REWRITING_V1 {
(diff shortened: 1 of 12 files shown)


jicofo/src/main/resources/reference.conf
```diff
@@ -93,6 +93,32 @@ jicofo {
...
+      // Whether to check the capabilities of endpoints.
+      enabled = true
+
+      // Whether to actually skip inviting an endpoint which does not advertise a capability with a "hard" level. When
+      // this is false jicofo only logs and updates metrics, and notifies the endpoint as if the level were "soft".
+      enforce = false
+
+      // The requirements, keyed by the name of a feature known to jicofo (see Features.kt). The set of requirements is
+      // maintained in jicofo (see ClientRequirementsConfig.defaultRequirements), and the entries here change it.
+      //   level:   "hard" (the endpoint is not invited), "soft" (the endpoint is invited and warned), or "off" (the
```
saghul, line 110: I feel like we might want to add an extra "critical" or something here, where we even refuse letting them into the MUC.

We might not have the use case today, but maybe it could be a safety net for the future?

saghul (approved):
LGTM with a suggestion.

jicofo/src/main/resources/reference.conf
bgrozev: This would have to go to prosody somehow. We could have jicofo kick them, but I decided to stay on the safe side. The capabilities mechanism is voluntary, so we couldn't use it to enforce policy.

PR #1317, opened 2026-09-11
Update to Kotlin 2.4.20, JUnit 5.14.4, kotest 5.9.1, mockk 1.14.11, Jackson 2.21.5

JonathanLennox:
Bumps Kotlin to 2.4.20 and the test tooling to current versions. Jackson 2.21.5 matches what jitsi-utils now uses.

Declares kotlin-stdlib directly instead of the empty kotlin-stdlib-jdk8 shell so the runtime stdlib always matches the compiler. Manages junit-platform-engine alongside the launcher: kotest 5.9 drags in platform 1.8, which otherwise wins over the version the Jupiter 5.14 engine needs and crashes the forked test JVM. Disables kotest classpath autoscan; the set of tests run is unchanged.

This is one of a set of same-named kotlin-2.4 branches across jitsi-utils, jitsi-metaconfig, jicoco, jitsi-xmpp-extensions, ice4j, jicofo, jitsi-videobridge and jibri. Each PR only bumps the compiler and test/doc tooling; dependency versions between the jitsi repos are unchanged, so the PRs are independently mergeable. A library built with Kotlin 2.4 can only be consumed by Kotlin 2.3 or newer, so the compiler bumps should land everywhere before any repo picks up a 2.4-built release of another.

### DIFF
--- a/jicofo-common/pom.xml
+++ b/jicofo-common/pom.xml
@@ -213,6 +213,7 @@
         <configuration>
           <systemPropertyVariables>
             <java.util.logging.config.file>${project.parent.basedir}/lib/logging.properties</java.util.logging.config.file>
+            <kotest.framework.classpath.scanning.autoscan.disable>true</kotest.framework.classpath.scanning.autoscan.disable>
           </systemPropertyVariables>
         </configuration>
       </plugin>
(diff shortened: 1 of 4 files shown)


damencho (approved):

JonathanLennox:
retest this please

